PR #14, opened 2016-06-03
Fix template filepath

channemann:
Path for `RemoteSettings-template.plist` was specifying a directory in which the user was already located


loudnate:
LGTM thanks!

PR #19, opened 2016-06-06
Added example icon

channemann:


loudnate:
This is great! Ideally we can link this to either your repository or somewhere for folks to download the design files.

channemann:
Good call; I'll link to the design files in my repo when I switch over.

Issue #35, opened 2016-06-19
Move RemoteSettings.plist config into the app


loudnate:
And properly store sensitive strings in the Keychain.

PR #57, opened 2016-07-04
Allow NIghtscout credentials to be updated within the app settings

loudnate:
Towards #35

### DIFF
--- a/Loop/Models/NightscoutService.swift
+++ b/Loop/Models/NightscoutService.swift
@@ -1,0 +1,80 @@
+//
+//  NightscoutService.swift
+//  Loop
+//
+//  Created by Nate Racklyeft on 7/3/16.
+//  Copyright © 2016 Nathan Racklyeft. All rights reserved.
+//
+
+import Foundation
+import NightscoutUploadKit
+
+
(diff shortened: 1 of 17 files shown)


Loop/Models/NightscoutService.swift
```diff
@@ -0,0 +1,80 @@
...
+
+    private(set) var isAuthorized: Bool = false
+
+    mutating func verify(completion: (success: Bool, error: ErrorType?) -> Void) {
+        guard let siteURL = siteURL, APISecret = APISecret else {
+            completion(success: false, error: nil)
+            return
+        }
+
+        self.uploader = NightscoutUploader(siteURL: siteURL.absoluteString, APISecret: APISecret)
```
loudnate: Filed https://github.com/ps2/rileylink_ios/issues/147 to see if we can add a simple no-op check of these credentials.

amazaheri:
awesome I added something similar for Push Notification azure setting. I can remove that and extend this...

loudnate:
@amazaheri yes, by conforming to the `ServiceAuthentication` protocol, you can get the UI nearly for free.

PR #68, opened 2016-07-16
722 reservoir vol

jlucasvt:
Added 722 and 522 HUD Reservoir Status display.  I know you want to add the Model volumes to PumpModel Enum in MinimedKit (in rileylink_ios) but those mods are in source outside my current capacity at the moment (I think).


loudnate:
@jlucasvt this PR has your app icons and other project changes in it. I can't merge it unless you revert those changes and squash; it's probably easier to create a clean branch from loop/dev. I will do some code review later this weekend to help you along with future PRs.

jlucasvt:
I see after I submitted the PR that it was messed up.    

I assume I can pull from loudnate/Loop Dev to a Local Branch
Merge the StatusTableViewController.swift commit from my forked/branch
Commit changes, push and Submit a new clean PR

Loop/View Controllers/StatusTableViewController.swift
```diff
@@ -195,7 +195,16 @@ class StatusTableViewController: UITableViewController, UIGestureRecognizerDeleg
             }
 
             reservoirVolume = dataManager.latestReservoirValue?.unitVolume
-
+			
+			// Remaining Volume HUD for x22 pumps
+			if reservoirVolume > 0 {
+				switch NSUserDefaults.standardUserDefaults().pumpModelNumber! {
```
loudnate: There's no guarantee this value exists, so force-unwrapping it could cause users to crash (especially first-time users or those who recently switched pump IDs).
It also should be made available to users with other pumps who might not have sentry packets configured for any reason, so it would be better in an else block on line 202/211:
```swift
} else if let capacity = dataManager.pumpState?.pumpModel?.reservoirCapacity {
    reservoirLevel = min(1, max(0, Double(reservoirVolume / capacity) / 100)
}
```

I know you expressed hesitation about adding reservoirCapacity to the rileylink_ios repo, but that doesn't prevent you from temporarily extending the `PumpModel` enum in this code base.

`Extensions/PumpModel.swift`
```swift
extension PumpModel {
    private var size: Int {
        return Int(rawValue)! / 100
    }

    var reservoirCapacity: Int {
        switch size {
        case 5:
            return 176
        case 7:
            return 300
        default:
            fatalError("Unknown reservoir capacity for PumpModel.\(self)")
        }
    }
}
```

jlucasvt:
Thank you for the Swift lesson. I hope I found all the spaces.  I Added the PumpModel extension and Modified the codebase to use extension and handle the unrwapped val.

Loop/Extensions/PumpModel.swift
```diff
@@ -0,0 +1,27 @@
+//
+//  PumpModel.swift
```
loudnate: There's still tabs in this file,

Loop/View Controllers/StatusTableViewController.swift
```diff
@@ -196,13 +196,10 @@ class StatusTableViewController: UITableViewController, UIGestureRecognizerDeleg
...
             reservoirVolume = dataManager.latestReservoirValue?.unitVolume
 			
-			// Remaining Volume HUD for x22 pumps
-			if reservoirVolume > 0 {
-				switch NSUserDefaults.standardUserDefaults().pumpModelNumber! {
-				case "722": reservoirLevel = Double(reservoirVolume! / 300) / 100  // 722 Reservior Vol = 300u
-				case "522": reservoirLevel = Double(reservoirVolume! / 176) / 100  // 522 Reservoir Vol= 176u
-				default: reservoirLevel = 0
-				}
+			if let capacity = dataManager.pumpState?.pumpModel?.reservoirCapacity {
```
loudnate: You can chain these two lines together:
```swift
if let capacity = dataManager.pumpState?.pumpModel?.reservoirCapacity,
    reservoirVolume = reservoirVolume
{
    reservoirLevel = min(1, max(0, Double(reservoirVolume / capacity) / 100))
}
```

loudnate:
New changes look good (with a few comments) but this still needs to be squashed and rebased against dev.

jlucasvt:
Closing this request and re-submitting under dev branch squash of only changes.

PR #71, opened 2016-07-17
Update for x22 Pumps to Display Reservoir Status HUD

jlucasvt:
Update to allow x22 pumps to display the Reservoir Status in the HUD.  Added PumpModel.swift Extension, Updated Logic and Tested Values and Display.


jlucasvt:
I also added the x22 BIoB value from the Chart Decay

loudnate:
That's not bolus IOB, that's net IOB.

loudnate:
The first commit looks good, but the second one is displaying incorrect data. Please resubmit with just the reservoir change.

Loop/View Controllers/StatusTableViewController.swift
```diff
@@ -501,7 +506,14 @@ class StatusTableViewController: UITableViewController, UIGestureRecognizerDeleg
                     let numberValue = NSNumber(double: iob).descriptionWithLocale(locale)
                     cell.detailTextLabel?.text = "\(numberValue) Units"
                 } else {
-                    cell.detailTextLabel?.text = emptyValueString
+                    //x22 BIoB Value from IOB Chart Decay Data
+                    dataManager.doseStore.getInsulinOnBoardValues(startDate: NSDate(), endDate: NSDate().dateByAddingTimeInterval(300)) { (values, error) -> Void in
```
loudnate: A tip here: you don't usually want to make asynchronous calls in table data source methods, because you don't own the underlying cell views or control their mapping and reuse. By the time this completion closure executes, the system may have reused the cell for a completely different index path – and your call would be setting label text that's completely out-of-place.

Another issue is that the documentation for `getInsulinOnBoardValues` states that it executes your completion closure on an arbitrary background queue. You cannot use the vast majority of UIKit APIs (like setting label text) on any queue except the main one, so you'll need to create a dispatch call to the main queue. Using UIKit on a background queue can result in unexpected behavior. At best, that means the label text isn't drawn at all until someone else invokes a display call. At worst, the app can crash.

Finally, 
```swift
NSDate().dateByAddingTimeInterval(300)
```
can be more expressively written as
```swift
NSDate(timeIntervalSinceNow: NSTimeInterval(minutes: 5))
```

Loop/View Controllers/StatusTableViewController.swift
```diff
@@ -501,7 +506,14 @@ class StatusTableViewController: UITableViewController, UIGestureRecognizerDeleg
...
+                        if let error = error {
+                            self.dataManager.logger.addError(error, fromSource: "DoseStore")
+                        }else{
+                            cell.detailTextLabel?.text = "\(NSString(format: "%.2f", values[0].value)) Units"
```
loudnate: There's a few things to note here. First, formatting numbers to strings should be done by customizing an instance of `NSNumberFormatter`. This will ensure the user's locale is respected; for example, when choosing a decimal separator: `7.3` vs `7,3`.

Second, the pattern `"%@ Units"` should allow for internationalization, by wrapping it in the `NSLocalizedString` function. If someone does want to translate Loop, they'll be able to do that easily if all english labels are properly setup ahead of time.

You might notice there's lots of cells in this section that aren't using `NSLocalizedString`... and that's because I've planned to continue phasing this entire text-based section of the table view out of the UI.

jlucasvt:
I pulled the Net IOB Code out.   Especially if the plan is to move away from text cells in future.  Thanks again for the knowledge share.

loudnate:
Awesome, thanks! I can squash the the multiple commits into one, single commit. Just be aware that you'll need to re-merge from Loop/dev into your own fork because the commit hash will change.

Issue #37, opened 2016-06-20
"Last sensor" is confusing for new users


danamlewis:
"Last sensor" is confusing - suggesting re-naming to "Last Sensor Reading" or something more descriptive.

PR #75, opened 2016-07-18
Change title for clarity

sarahspins:
As mentioned in #37 by @danamlewis but there wasn't room to fit her suggestion.


loudnate:
Thanks! These text rows are slowly going away, but I'm happy to take this change now.

sarahspins:
But they're so helpful! :P  Ideally I'd rather have a time since last CGM rather than the timestamp where I have to do the mental math.... not quite sure how to accomplish that though.

PR #80, opened 2016-07-21
Update to RL 0.7.1 frameworks

ps2:


.gitignore
```diff
@@ -65,4 +65,8 @@ fastlane/report.xml
 fastlane/screenshots
 
 # Sensitive details about remote backup
-RemoteSettings.plist
\ No newline at end of file
+RemoteSettings.plist
+
+# OS
+.DS_Store
```
loudnate: If I'm begin pedantic, these should be in your global `.gitignore`.

PR #81, opened 2016-07-21
Highlight project selection on bundle identifier screenshot

ps2:
I know I've missed this, and I think others have too, so thought it might be worth a skitch highlight.


loudnate:
Awesome, thanks!

PR #88, opened 2016-07-25
Add support for G4 Share over BLE

mddub:
This has been working well for me for the last several days.

* To build, you'll need to run `carthage bootstrap`. When I did that to bring in the new dependency, it resulted in diffs for every included framework. I don't really understand why. Should I commit what Carthage produced for me, or let you do that?

* Would also appreciate any feedback you have on [xDripG4Share][1].

[1]: https://github.com/mddub/xDripG4Share


mddub:
Sorry, meant to open this against dev.

PR #89, opened 2016-07-25
Add support for G4 Share over BLE

mddub:
This has been working well for me for the last several days.

* To build, you'll need to run `carthage bootstrap`. When I did that to bring in the new dependency, it resulted in diffs for every included framework. I don't really understand why. Should I commit what Carthage produced for me, or let you do that?

* Would also appreciate any feedback you have on [xDripG4Share][1].

[1]: https://github.com/mddub/xDripG4Share


mddub:
Looks like I need to commit `xDripG4Share.framework` after all. Otherwise, CI can't run.

loudnate:
Excellent surprise! So thrilled you built this.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -533,6 +560,59 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate {
...
+        guard let latest = glucoseHistory.sort({ (a, b) in a.sequence < b.sequence }).last where latest != latestGlucoseG4 else {
+            return
+        }
+        latestGlucoseG4 = latest
+
+        guard let glucoseStore = glucoseStore else {
+            return
+        }
+
+        let validGlucose = glucoseHistory.map({
```
loudnate: flatMap will handle the subsequent nil filter for you

mddub:
This is ready for review. I've renamed the framework to `G4ShareSpy` to avoid confusion, and changed the switch label to "G4 Share Receiver (beta)".

All the interesting Bluetooth handling in the framework is in [`BluetoothManager`][1]; the rest of the code is message parsing.

[1]: https://github.com/mddub/G4ShareSpy/blob/master/G4ShareSpy/BluetoothManager.swift

loudnate:
👍🏻👏

Loop/Managers/DeviceDataManager.swift
```diff
@@ -533,6 +560,59 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate {
...
+    // MARK: - Share Receiver
+
+    // MARK: ReceiverDelegate
+
+    var latestGlucoseG4: GlucoseG4?
+
+    func receiver(receiver: Receiver, didReadGlucoseHistory glucoseHistory: [GlucoseG4]) {
+        assertCurrentPumpData()
+
+        guard let latest = glucoseHistory.sort({ (a, b) in a.sequence < b.sequence }).last where latest != latestGlucoseG4 else {
```
loudnate: Tip: you can write `{ (a, b) in a.sequence < b.sequence }` as `{ $0.sequence < $1.sequence }`. You can also implement the `Comparable` protocol on `GlucoseG4` to use the `sequence` attribute, and then you can simply write `glucoseHistory.sort(<)`

Loop/Managers/DeviceDataManager.swift
```diff
@@ -859,6 +936,7 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate {
 
         defer {
             transmitterID = NSUserDefaults.standardUserDefaults().transmitterID
```
loudnate: This is unfortunately on my cleanup list... tricking the `didSet` pattern isn't great.

Loop/Models/GlucoseG4.swift
```diff
@@ -0,0 +1,31 @@
...
+//
+//  Created by Mark Wilson on 7/21/16.
+//  Copyright © 2016 Nathan Racklyeft. All rights reserved.
+//
+
+import Foundation
+import G4ShareSpy
+
+
+let TREND_TO_DESCRIPTION: [UInt8: String] = [
```
loudnate: `private let trendDescriptionMap`?

Loop/Extensions/NSUserDefaults.swift
```diff
@@ -192,6 +193,15 @@ extension NSUserDefaults {
         }
     }
 
+    var receiverEnabled: Bool {
+        get {
+            return boolForKey(Key.ReceiverEnabled.rawValue)
+        }
+        set {
+            setObject(newValue, forKey: Key.ReceiverEnabled.rawValue)
```
loudnate: `setBool(_, forKey:)`

loudnate:
Does the G4 Share app keep the receiver's time up-to-date with the phone? If not, this will need to consider time zone issues much like what's done for the pump.

mddub:
> Does the G4 Share app keep the receiver's time up-to-date with the phone? If not, this will need to consider time zone issues much like what's done for the pump.

Display time on the receiver doesn't matter, only system time (from the RTC). Every record has a system time and a display time. Every time the Share app requests glucose history, it also requests the system time, which I use to [store/update the system clock offset][1], so every glucose record gets an absolute timestamp (which matches what ends up on the Share server).

[1]: https://github.com/mddub/G4ShareSpy/blob/0a009f/G4ShareSpy/Receiver.swift#L93-L105

loudnate:
Awesome, OK - that's exactly how the G5 works too (since the system time epoch is when first activated).

Nice work!

PR #92, opened 2016-07-30
Version 0.8.0

loudnate:
Adds `G4ShareSpy.framework` for capturing Bluetooth communication between a G4 Receiver and the Share2 app, enabling offline access to CGM data.

This is a major improvement in connectivity and reliability for G4 users, and it's all thanks to the hard work of @mddub.


amazaheri:
nice! and well done

PR #93, opened 2016-07-31
Confirmed Loop working with AU - MM554 and Enlite CGM

johnmales:


loudnate:
Thanks! FYI that I'm going to squash this down, which will adjust the commit hash.

PR #96, opened 2016-08-01
Uploading DoseStore history using new NightscoutUploadKit APIs

loudnate:


Loop/Managers/DeviceDataManager.swift
```diff
@@ -871,6 +851,36 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverDelegat
 
     let doseStore: DoseStore
 
+    // MARK: DoseStoreDelegate
+
+    func doseStore(doseStore: DoseStore, hasEventsNeedingUpload pumpEvents: [(objectID: NSManagedObjectID, date: NSDate, dose: DoseEntry?, raw: NSData?)], withCompletion completionHandler: (uploadedObjects: [NSManagedObjectID]) -> Void) {
+
+        guard let uploader = remoteDataManager.nightscoutUploader, device = rileyLinkManager.firstConnectedDevice, pumpModel = pumpState?.pumpModel else {
```
ps2: Device used to indicate the RL that actually fetched the events. It's unused beyond just debugging, afaik, but I'm wondering if there's an easy way to keep that association.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -225,7 +223,9 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverDelegat
             NotificationManager.sendPumpBatteryLowNotification()
         }
 
-        remoteDataManager.nightscoutUploader?.handlePumpStatus(status, device: device.deviceURI)
+        if let nsUploader = remoteDataManager.nightscoutUploader, let pumpStatus = try? nsUploader.getPumpStatusFromMySentryPumpStatus(status) {
+            nsUploader.uploadDeviceStatus(DeviceStatus(device: device.deviceURI, timestamp: pumpDate, pumpStatus: pumpStatus, uploaderStatus: nil, loopStatus: nil))
+        }
```
ps2: Thanks for doing this for me. :)

Loop/Managers/DeviceDataManager.swift
loudnate: Yeah, I noticed that. Same goes for PumpState, which seem to fly around left-and-right (and why I want to change that to a struct). If it's important, I can update the schema to store the device identifier. I'm not sure the specific device-name-based string matters in any real-world scenarios. Is the pump ID sent to Nightscout?

ps2:
LGTM.  It'd be nice to find a way to keep the RL association with events, but I understand that we're crossing a couple abstraction layers and it might require some gymnastics to do so.

Loop/Managers/DeviceDataManager.swift
ps2: The only use I've ever had for it was when using multiple RLs; and seeing when one took over.  PumpID is not uploaded, but I could certainly see that being useful if you were switching pumps.   Switching pumps is a pretty complicated scenario; would be difficult to offer full support for it, I think, but things like tracking which pump an event came from would certainly help.

loudnate: Do you regularly use multiple RileyLinks at once? I'm trying to understand what questions you could answer by knowing which one a particular treatment came from. 

On the other hand, I can easily think of questions that would require knowing which pump a treatment came from. Even without the support for easily switching pumps in one app. Shortly, when insulin data is retrieved from history, switching pumps will be totally feasible.

loudnate: What I'm getting at is, can we construct the identifier from Pump ID and use a generic name for all RileyLinks here.

loudnate: com.rileylink://123456

And what about the app bundle ID?

loudnate: Haha so I'm hearing we should include everything? What's the preferred syntax?

```
iCorder-6s:com.loudnate.Loop:Upstairs-RileyLink:123456//
```

For all those identifiers above, only the RileyLink name isn't directly known at the time of the delegate call, but I can change that by adding a device ID string to the DoseStore PumpEvent schema (Pump ID is already there).

I filed https://github.com/loudnate/LoopKit/issues/48 to handle the storage and #98 to handle the identifier format.

loudnate:
Going to merge this, with adjustment promises tracked as https://github.com/loudnate/LoopKit/issues/48 and #98.

PR #99, opened 2016-08-02
Reporting pump status from x22 devices

ps2:
Just creating this for discussion/review.  Not ready for merge. I'll fix the MAIN_APP_BUNDLE_IDENTIFIER and the Cartfile will eventually reference v0.8.0, which I'm making now.


amazaheri:
w00t! 👍

Cartfile.resolved
```diff
@@ -1,8 +1,8 @@
-github "amplitude/Amplitude-iOS" "a3d6720626e72a678489bce83ed9dacadb505dd7"
+github "amplitude/Amplitude-iOS" "32d36977147d9f9a6330002afdc85fe31cfbcfaa"
```
loudnate: Can you revert this guy until I can test it? `carthage update rileylink_ios` will isolate the fetch

Loop/Managers/DeviceDataManager.swift
```diff
@@ -318,20 +389,45 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverDelegat
...
         device.assertIdleListening()
 
-        // How long should we wait before we poll for new reservoir data?
-        let reservoirTolerance = rileyLinkManager.idleListeningEnabled ? NSTimeInterval(minutes: 11) : NSTimeInterval(minutes: 4)
+        // How long should we wait before we poll for new pump data?
+        let pumpStatusAgeTolerance = rileyLinkManager.idleListeningEnabled ? NSTimeInterval(minutes: 11) : NSTimeInterval(minutes: 4)
 
-        // If we don't yet have reservoir data, or it's old, poll for it.
-        if latestReservoirValue == nil || latestReservoirValue!.startDate.timeIntervalSinceNow <= -reservoirTolerance {
-            readReservoirVolume { (result) in
```
loudnate: Can you delete this method now?

ps2: It is also fetched before bolusing if the value is older than 5 minutes.  I wasn't sure if that should be removed.

Cartfile.resolved
ps2: Reverted.  It's failing travis because the binaries aren't checked in.  Would you like me to commit those?  I'm not sure how much this kind of activity grows the repo.

Loop/Managers/DeviceDataManager.swift
loudnate: hmm, that's true, but since these values are always being updated in bulk anyway, if one is old then they're all old...

Loop/Managers/DeviceDataManager.swift
```diff
@@ -151,7 +151,27 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverDelegat
...
+        didSet {
+            if let update = latestPumpStatusFromMySentry, let timeZone = pumpState?.timeZone {
+                let pumpClock = update.pumpDateComponents
+                pumpClock.timeZone = timeZone
+                latestPumpStatusDate = pumpClock.date
+            }
+        }
+    }
+
+    var latestPolledPumpStatus: RileyLinkKit.PumpStatus? {
```
loudnate: `private`

Loop/Managers/DeviceDataManager.swift
```diff
@@ -151,7 +151,27 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverDelegat
 
     // MARK: Pump data
 
-    var latestPumpStatus: MySentryPumpStatusMessageBody?
+    var latestPumpStatusDate: NSDate?
```
loudnate: `private`

Loop/Managers/DeviceDataManager.swift
```diff
@@ -181,16 +201,33 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverDelegat
...
 
-        latestPumpStatus = status
+        latestPumpStatusFromMySentry = status
+
+
+        // Gather PumpStatus from MySentry packet
+        let pumpStatus: NightscoutUploadKit.PumpStatus?
+        if let pumpDate = status.pumpDateComponents.date, let pumpID = pumpID {
+
+            let batteryStatus = BatteryStatus(percent: status.batteryRemainingPercent, status: "normal")
```
loudnate: Should `"normal"` be a default value in the `BatteryStatus` constructor instead?

Loop/Managers/DeviceDataManager.swift
```diff
@@ -181,16 +201,33 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverDelegat
...
+            let iobStatus = IOBStatus(iob: status.iob, basaliob: 0, timestamp: pumpDate)
```
loudnate: (same question for `basaliob`)

Loop/Managers/DeviceDataManager.swift
```diff
@@ -181,16 +201,33 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverDelegat
...
+
+            pumpStatus = NightscoutUploadKit.PumpStatus(clock: pumpDate, pumpID: pumpID, iob: iobStatus, battery: batteryStatus, reservoir: status.reservoirRemainingUnits)
+        } else {
+            pumpStatus = nil
+            print("Could not interpret pump clock: \(status.pumpDateComponents)")
```
loudnate: Remove, or log to the diagnostic logger?

Loop/Managers/DeviceDataManager.swift
```diff
@@ -318,20 +389,45 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverDelegat
...
-        if latestReservoirValue == nil || latestReservoirValue!.startDate.timeIntervalSinceNow <= -reservoirTolerance {
-            readReservoirVolume { (result) in
+        // If we don't yet have pump status, or it's old, poll for it.
+        if latestPumpStatusDate == nil || latestPumpStatusDate!.timeIntervalSinceNow <= -pumpStatusAgeTolerance {
+            guard let device = rileyLinkManager.firstConnectedDevice else {
+                return
+            }
+
+            guard let ops = device.ops else {
+                self.troubleshootPumpCommsWithDevice(device)
```
loudnate: Why do this here?

Loop/Managers/DeviceDataManager.swift
```diff
@@ -318,20 +389,45 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverDelegat
...
-        let reservoirTolerance = rileyLinkManager.idleListeningEnabled ? NSTimeInterval(minutes: 11) : NSTimeInterval(minutes: 4)
+        // How long should we wait before we poll for new pump data?
+        let pumpStatusAgeTolerance = rileyLinkManager.idleListeningEnabled ? NSTimeInterval(minutes: 11) : NSTimeInterval(minutes: 4)
 
-        // If we don't yet have reservoir data, or it's old, poll for it.
-        if latestReservoirValue == nil || latestReservoirValue!.startDate.timeIntervalSinceNow <= -reservoirTolerance {
-            readReservoirVolume { (result) in
+        // If we don't yet have pump status, or it's old, poll for it.
+        if latestPumpStatusDate == nil || latestPumpStatusDate!.timeIntervalSinceNow <= -pumpStatusAgeTolerance {
+            guard let device = rileyLinkManager.firstConnectedDevice else {
```
loudnate: repeat of line 386

Loop/Managers/DeviceDataManager.swift
```diff
@@ -318,20 +389,45 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverDelegat
...
+            }
+
+            ops.readPumpStatus({ (result) in
                 switch result {
-                case .Success(let (units, date)):
-                    self.updateReservoirVolume(units, atDate: date, withTimeLeft: nil)
+                case .Success(let status):
+                    self.latestPolledPumpStatus = status
+
+                    let battery = BatteryStatus(voltage: status.batteryVolts, status: String(status.batteryStatus).lowercaseString)
```
loudnate: > String(status.batteryStatus).lowercaseString

Is this intended to be i18n friendly? or is it a Nightscout enumeration of some kind? If the former, then it should be adjusted to pull from the strings bundle. If the latter, then best to keep it internal to `NightscoutUploadKit` and surface a native `enum`.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -318,20 +389,45 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverDelegat
...
+                    status.clock.timeZone = ops.pumpState.timeZone
+                    guard let date = status.clock.date else {
+                        print("Could not interpret clock")
```
loudnate: diagnostic logger or remove?

Loop/View Controllers/StatusTableViewController.swift
```diff
@@ -211,9 +211,11 @@ class StatusTableViewController: UITableViewController, UIGestureRecognizerDeleg
                 resVol = reservoirVolume {
                 reservoirLevel = min(1, max(0, Double(resVol / Double(capacity))))
             }
-            
-            if let status = dataManager.latestPumpStatus {
-                reservoirLevel = Double(status.reservoirRemainingPercent) / 100
+
+            reservoirLevel = dataManager.latestReservoirValue?.unitVolume
```
loudnate: This is undoing line 212

loudnate: And it's mixing whats intended to be a percent with a unit volume. (I can't wait for the iOS 10 `Measurement` type)

Loop/View Controllers/StatusTableViewController.swift
```diff
@@ -494,15 +496,15 @@ class StatusTableViewController: UITableViewController, UIGestureRecognizerDeleg
             case .Date:
                 cell.textLabel?.text = NSLocalizedString("Last Sentry", comment: "The title of the cell containing the last updated date")
 
-                if let date = dataManager.latestPumpStatus?.pumpDateComponents.date {
+                if let date = dataManager.latestPumpStatusDate {
```
loudnate: If this reflects the last date we read pump data from any source now, then the label on 497 should change.

Cartfile.resolved
loudnate: Yes – go ahead and check the framework bundles in. I'll squash when I merge, so there's no concern in committing multiple updates if necessary if you want to do so before you cut the release tag (but please run `carthage update rileylink_ios` one last time after the tag is created)

Loop/Managers/DeviceDataManager.swift
ps2: The readReservoirVolume does not have any effect on the bolusing action; i.e. even if the readReservoirVolume indicates a lot more insulin on board than previous, the bolus will still be for the same number of units.
Based on that, I think I should probably just remove this call.   If I'm missing something, let me know.

loudnate: right, it's only called before bolusing if the reservoir data is stale to try and preserve the integrity of the IOB calculation. I'm ok with replacing it with the read status call.

Loop/View Controllers/StatusTableViewController.swift
ps2: yikes!  Sorry about that.

Loop/View Controllers/StatusTableViewController.swift
ps2: I think it's more useful as a mysentry diagnostic line.. Reverting.

loudnate: I'm OK with it either way; I'm planning to kill off all of these text lines in favor of something more glance-able, and I would probably use `latestPumpStatusDate`

Loop/Managers/DeviceDataManager.swift
```diff
@@ -224,10 +264,41 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverDelegat
...
+            return
+        }
+
+
+        // Gather UploaderStatus
+        let uploaderDevice = UIDevice.currentDevice()
 
-        remoteDataManager.nightscoutUploader?.handlePumpStatus(status, device: device.deviceURI)
+        let battery: Int?
+        if uploaderDevice.batteryMonitoringEnabled {
```
loudnate: Do we need to enable device battery monitoring when NightscoutUploader becomes authenticated?

ps2: To report phone battery to NS, yes, we'll need to enable that.  Good catch.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -151,7 +151,27 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverDelegat
...
+        didSet {
+            if let update = latestPumpStatusFromMySentry, let timeZone = pumpState?.timeZone {
+                let pumpClock = update.pumpDateComponents
+                pumpClock.timeZone = timeZone
+                latestPumpStatusDate = pumpClock.date
+            }
+        }
+    }
+
+    private var latestPolledPumpStatus: RileyLinkKit.PumpStatus? {
```
loudnate: Isn't this `NightscoutUploadKit.PumpStatus`?

Loop/Managers/DeviceDataManager.swift
```diff
@@ -697,8 +789,9 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverDelegat
             if let sentrySupported = pumpState?.pumpModel?.larger where !sentrySupported {
                 rileyLinkManager.idleListeningEnabled = false
             }
-
             NSUserDefaults.standardUserDefaults().pumpModelNumber = pumpState?.pumpModel?.rawValue
+            self.enableRileyLinkHeartbeatIfNeeded()
```
loudnate: This call shouldn't be necessary, as it's only dependent on sensor hardware configuration.

ps2: If they are using share and switch from a 523 pump to a 522 pump, then don't we need to enable the heartbeat?

loudnate: No, a 523 pump still needs the heartbeat to assert idle is running.

loudnate: The G5 transmitter and G4 receiver are the only other reliable heartbeats, and they're preferred due to their longer (5 min) interval

ps2: I was thinking idle listening for mysentry was enough of a heartbeat; but you had issues with idle being disabled somehow?

loudnate: Idle listening will fail to restart ~once every 24 hours since it doesn't restart independently of the app waking.

This was an observation I made early on, and it was relatively recently re-confirmed when someone tried looping using just G4 download before I surfaced the timer tick.

Loop/View Controllers/StatusTableViewController.swift
```diff
@@ -205,16 +205,14 @@ class StatusTableViewController: UITableViewController, UIGestureRecognizerDeleg
                 }
             }
 
-            reservoirVolume = dataManager.latestReservoirValue?.unitVolume
```
loudnate: This (class property) needs to be set here before its referenced below on line 209

loudnate:
LGTM! Whenever you're ready, cut the RL release and commit the framework binaries.

ps2:
Ok, I have everything in here I want to get in. Thanks for all the reviewing!

loudnate:
Woo! I squashed, so you'll want to work from `dev` rather than the branch commits.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -417,27 +448,19 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverDelegat
...
-                            self.logger.addError(error, fromSource: "Bolus")
-                            completion(error: LoopError.CommunicationError)
-                        } else {
-                            self.latestReservoirValue = newValue
-                            setBolus()
-                        }
-                    }
-                case .Failure(let error):
-                    completion(error: error)
+            readAndUpdatePumpData({ (error) in
```
loudnate: This code path, unfortunately, triggers a whole bunch of things like pump history fetching and the basal adjustment calculation via `updateReservoirVolume`, which is why the previous block wasn't calling it. I'm working on a few refactors to streamline this file.. sorry for the confusion.

PR #101, opened 2016-08-05
Adjustments to pump status uploading:

loudnate:
* Removing some unnecessary state
* Preventing a pre-bolus status read from triggering the loop and history dump

### DIFF
--- a/Loop/Managers/DeviceDataManager.swift
+++ b/Loop/Managers/DeviceDataManager.swift
@@ -151,27 +151,7 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverDelegat
 
     // MARK: Pump data
 
-    private var latestPumpStatusDate: NSDate?
-
-    var latestPumpStatusFromMySentry: MySentryPumpStatusMessageBody? {
-        didSet {
-            if let update = latestPumpStatusFromMySentry, let timeZone = pumpState?.timeZone {
-                let pumpClock = update.pumpDateComponents
-                pumpClock.timeZone = timeZone
-                latestPumpStatusDate = pumpClock.date
-            }
-        }
-    }
-
-    private var latestPolledPumpStatus: RileyLinkKit.PumpStatus? {
-        didSet {
-            if let update = latestPolledPumpStatus, let timeZone = pumpState?.timeZone {
-                let pumpClock = update.clock
-                pumpClock.timeZone = timeZone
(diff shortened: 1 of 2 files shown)


Loop/Managers/DeviceDataManager.swift
```diff
@@ -151,27 +151,7 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverDelegat
 
     // MARK: Pump data
 
-    private var latestPumpStatusDate: NSDate?
```
loudnate: This was exactly the same as (and in-sync with) the existing `latestReservoirValue.startDate`, so I just reverted the uses of this one.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -151,27 +151,7 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverDelegat
...
-        didSet {
-            if let update = latestPumpStatusFromMySentry, let timeZone = pumpState?.timeZone {
-                let pumpClock = update.pumpDateComponents
-                pumpClock.timeZone = timeZone
-                latestPumpStatusDate = pumpClock.date
-            }
-        }
-    }
-
-    private var latestPolledPumpStatus: RileyLinkKit.PumpStatus? {
```
loudnate: This wasn't being read, though its easy to re-add it when we need it.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -307,35 +285,39 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverDelegat
...
-
                 status.clock.timeZone = ops.pumpState.timeZone
                 guard let date = status.clock.date else {
                     self.logger.addError("Could not interpret pump clock: \(status.clock)", fromSource: "RileyLink")
-                    completion(error: LoopError.ConfigurationError)
+                    completion(.Failure(LoopError.ConfigurationError))
                     return
                 }
 
-                self.updateReservoirVolume(status.reservoir, atDate: date, withTimeLeft: nil)
```
loudnate: The read + update cases needed to be decoupled originally for the bolus case, since `updateReservoirVolume` triggers history polling and looping.

short-term, it seemed best to revert the behavior until history-based insulin dosage is ready, for which the device-trigger web will need to be pliable yet again.

PR #110, opened 2016-08-08
Loop status upload

ps2:
Uploads Loop status to Nightscout, with BG forecasts.


Loop/Managers/LoopDataManager.swift
```diff
@@ -94,11 +101,80 @@ class LoopDataManager {
...
+        let failureReason: String?
+
+
+        if let lastLoopError = lastLoopError {
+            failureReason = String(lastLoopError)
+        } else {
+            failureReason = nil
+        }
+
+        let loopName = NSBundle.mainBundle().bundleDisplayName
```
loudnate: Can we move lines like this out of the LoopDataManager and into the receiver's method?

loudnate: I think ideally all the NightscoutUploadKit models would be created somewhere else, like DiagnosticLogger, or an extension of NightscoutUploader?

Loop/Managers/LoopDataManager.swift
```diff
@@ -94,11 +101,80 @@ class LoopDataManager {
...
+       let eventualBG = self.predictedGlucose?.last?.quantity.doubleValueForUnit(HKUnit.milligramsPerDeciliterUnit())
+
+        let loopSuggested: LoopSuggested?
+
+        let glucoseVal = glucose?.quantity.doubleValueForUnit(HKUnit.milligramsPerDeciliterUnit())
+
+        let predBGs: [Double]?
+        if let predicted = predictedGlucose {
+            predBGs = predicted[0...10].map({ (value) -> Double in
+                value.quantity.doubleValueForUnit(HKUnit.milligramsPerDeciliterUnit())
```
loudnate: Pull the unit class creation out of the map here

Loop/Managers/LoopDataManager.swift
```diff
@@ -94,11 +101,80 @@ class LoopDataManager {
...
+
+    func uploadLoopStatus(recommendation: TempBasalRecommendation?) {
+
+        let statusTime = NSDate()
+
+        let glucose = deviceDataManager.glucoseStore?.latestGlucose
+
+        let iob: IOBStatus?
+
+        if let iobValues = insulinOnBoardValues, iobValue = iobValues.first {
```
loudnate: Can collapse to `insulinOnBoardValues?.first`

Loop/Managers/LoopDataManager.swift
```diff
@@ -94,11 +101,80 @@ class LoopDataManager {
...
+
+       let eventualBG = self.predictedGlucose?.last?.quantity.doubleValueForUnit(HKUnit.milligramsPerDeciliterUnit())
+
+        let loopSuggested: LoopSuggested?
+
+        let glucoseVal = glucose?.quantity.doubleValueForUnit(HKUnit.milligramsPerDeciliterUnit())
+
+        let predBGs: [Double]?
+        if let predicted = predictedGlucose {
+            predBGs = predicted[0...10].map({ (value) -> Double in
```
loudnate: Length check first?

Loop/Managers/LoopDataManager.swift
```diff
@@ -147,11 +223,37 @@ class LoopDataManager {
             }
         }
 
+        // Not used directly in loop decision making, but used for reporting loop state
```
loudnate: Hmm. `WatchDataManager` and `StatusTabkeViewController` are both clients of Loop data but neither requires embedding in the LDM itself. Can I suggest that approach?

Loop/Managers/LoopDataManager.swift
```diff
@@ -11,6 +11,8 @@ import CarbKit
 import InsulinKit
 import LoopKit
 import MinimedKit
+import NightscoutUploadKit
```
loudnate: Per comment below, this relationship should really be reversed.

Loop/Managers/LoopDataManager.swift
```diff
@@ -278,12 +386,40 @@ class LoopDataManager {
         }
     }
 
+    private func updateIOB(completionHandler: (values: [InsulinValue]?, error: ErrorType?) -> Void) {
+        let glucose = deviceDataManager.glucoseStore?.latestGlucose
+
+        deviceDataManager.doseStore.getInsulinOnBoardValues(startDate: glucose?.startDate) { (values, error) -> Void in
```
loudnate: Since you only need one value here, use `deviceDataManager.doseStore.insulinOnBoardAtDate(startDate)`

Loop/Managers/LoopDataManager.swift
```diff
@@ -147,11 +223,37 @@ class LoopDataManager {
...
+                    self.insulinOnBoardValues = values
+                } else {
+                    self.insulinOnBoardValues = nil
+                }
+                dispatch_group_leave(updateGroup)
+            }
+        }
+
+        // Not used directly in loop decision making, but used for reporting loop state
+        if carbsOnBoardValues == nil {
```
loudnate: I don't see this used anywhere?

Loop/Managers/LoopDataManager.swift
```diff
@@ -94,11 +101,80 @@ class LoopDataManager {
...
+        let loopEnacted: LoopEnacted?
+        if let tempBasal = lastTempBasal where tempBasal.unit == .unitsPerHour &&
+            lastTempBasalUploaded?.startDate != tempBasal.startDate {
+            let duration = tempBasal.endDate.timeIntervalSinceDate(tempBasal.startDate)
+            loopEnacted = LoopEnacted(rate: tempBasal.value, duration: duration, timestamp: tempBasal.startDate, received:
+                true)
+            lastTempBasalUploaded = tempBasal
+        } else if let recommendation = recommendation {
+            // notEnacted
+            loopEnacted = LoopEnacted(rate: recommendation.rate, duration: recommendation.duration, timestamp: recommendation.recommendedDate, received: false)
```
loudnate: `self.lastTempBasal`?

mddub:
:tada::+1::clap::joy::heart_eyes: So glad you did this, @ps2.

This will make integrating with Urchin so much easier - I've been running a much simpler/hackier version of these changes to show the prediction graph and eventual BG:

img_3200

Loop/Managers/LoopDataManager.swift
mddub: Why not include all of them?

ps2: Yep, going to remove that, and also upload the carb and insulin only lines as well

Loop/Managers/LoopDataManager.swift
```diff
@@ -147,11 +221,23 @@ class LoopDataManager {
             }
         }
 
+        if insulinOnBoard == nil {
+            dispatch_group_enter(updateGroup)
+            deviceDataManager.doseStore.insulinOnBoardAtDate(NSDate()) { (value, error) in
```
ps2: Is this the right timestamp to use?

Loop/Managers/LoopDataManager.swift
```diff
@@ -357,6 +447,9 @@ class LoopDataManager {
         } else {
             recommendedTempBasal = nil
         }
+
+        recommendedBolus = nil
+        recommendedBolus = try self.recommendBolus()
```
loudnate: You should do this in `NightscoutDataManager` now.

Loop/Managers/LoopDataManager.swift
loudnate: from a dosing perspective, yeah, you'd want it to be as close-as-possible to `startDate` on line 386, which it will be here, since these data retrievals should be pretty fast. If it needs to be exact, you could set `startDate` on line 186 and pass it in as an argument to `self.updatePredictedGlucoseAndRecommendedBasal`

Loop/Managers/LoopDataManager.swift
ps2: I see WatchDataManager is also asking for current recommended bolus; that wouldn't make sense as a part of LoopStatus, would it?

loudnate: You tell me – why is it being called here then?

ps2: I'm calling it to put it in the upload.  I want to be able to remotely see if I should maybe give her a bolus (if temp basal isn't enough). It's not currently displayed, but I am planning on making a change in NS to display it when hovering on the NS pill.

loudnate: > I see WatchDataManager is also asking for current recommended bolus; that wouldn't make sense as a part of LoopStatus, would it?
Sounds like the same use case then?

Loop/Managers/NightscoutDataManager.swift
```diff
@@ -0,0 +1,113 @@
...
+        } else {
+            loopEnacted = nil
+        }
+        
+        let loopName = NSBundle.mainBundle().bundleDisplayName
+        let loopVersion = NSBundle.mainBundle().shortVersionString
+        
+        let loopStatus = LoopStatus(name: loopName, version: loopVersion, timestamp: statusTime, iob: iob, suggested: loopSuggested, enacted: loopEnacted, failureReason: lastLoopError)
+        
+        deviceDataManager.remoteDataManager.uploadDeviceStatus(nil, loopStatus: loopStatus)
```
loudnate: Maybe this implementation can move into here, now, to keep all the NS models together?

Also, would be good to punt at the top of this method if there's no nightscout uploader

ps2: Sure, sounds good.

ps2:
@mddub I've been using urchin, and love it. Forecasted bg would be amazing!  I'm just about done with this, I think.

ps2:
Was trying to get COB and IOB NS attributes populated to wrap this up, but it looks like they might require more work, so I'll tackle that as a separate PR.  I'll cut a RL release, and then point this to it, and the PR should be ready for review again.

Loop/Managers/LoopDataManager.swift
```diff
@@ -184,12 +199,15 @@ class LoopDataManager {
 
      - parameter resultsHandler: A closure called once the values have been retrieved. The closure takes the following arguments:
         - predictedGlucose:     The calculated timeline of predicted glucose values
+        - insulinEffect:        The predicted effect of insulin
+        - carbEffect:           The predicted effect of carbohydrates
```
loudnate: still need to add these?

Loop/Managers/LoopDataManager.swift
```diff
@@ -229,8 +248,11 @@ class LoopDataManager {
     private var predictedGlucose: [GlucoseValue]? {
         didSet {
             recommendedTempBasal = nil
+            recommendedBolus = nil
         }
     }
+
+    private var recommendedBolus: Double?
```
loudnate: Remove this now?

Loop/Managers/LoopDataManager.swift
```diff
@@ -296,7 +318,7 @@ class LoopDataManager {
      
      *This method should only be called from the `dataAccessQueue`*
      */
-    private func updatePredictedGlucoseAndRecommendedBasal() throws {
+    private func updatePredictedGlucoseAndRecommendations() throws {
```
loudnate: Revert name change?

Loop/Managers/NightscoutDataManager.swift
```diff
@@ -0,0 +1,167 @@
...
+
+    private func uploadLoopStatus(insulinOnBoard: InsulinValue?, predictedGlucose: [GlucoseValue]?, recommendedTempBasal: LoopDataManager.TempBasalRecommendation?, recommendedBolus: Double?, lastTempBasal: DoseEntry?, lastLoopError: ErrorType?) {
+
+        guard deviceDataManager.remoteDataManager.nightscoutUploader != nil else {
+            return
+        }
+
+        let statusTime = NSDate()
+        
+        // Is this just predictedGlucose[0]?
```
loudnate: yep

Loop/Managers/LoopDataManager.swift
ps2: I'm going to soon; I just need to figure out how to turn them into BG lines.  carbEffect -> predicted BG if only carbs were considered, and insulinEffect -> predicted BG if only insulin were considered.

But I can remove them for this commit, if you'd like.

Loop/Managers/NightscoutDataManager.swift
ps2: Are there situations where `deviceDataManager.glucoseStore?.latestGlucose` will be set, but predictedGlucose?[0] is not?

Looking at LoopDataManager, it fetches `self.deviceDataManager.glucoseStore?.latestGlucose` despite having access to predictedGlucose.

loudnate: Yes. If one of the preconditions for prediction fail (e.g. insulin data old)

ps2: Actually, that's probably what I want; there is no field to track date of the glucose value, and from how it's displayed, it seems to convey the value that the prediction was based on, so it should probably be nil if we have no predicted glucose.

Loop/Managers/LoopDataManager.swift
loudnate: Not sure I follow with the "only [x] were considered"?

I'm good with merging this PR if you are, which would mean reconciling this docstring and reverting the `import` line at the top. It sounds like there's still some munging of the carb/insulin effect lines for Nightscout, though ideally you can sort that processing out in the cloud. Just let me know what you'd like to do.

Loop/Managers/NightscoutDataManager.swift
```diff
@@ -0,0 +1,164 @@
...
+        let eventualBG = predictedGlucose?.last?.quantity
+        
+        let loopSuggested: LoopSuggested?
+
+        let glucoseVal = predictedGlucose?.first?.quantity
+        
+        let predBGs: PredictedBG?
+        
+        if let predicted = predictedGlucose {
+            let values = predicted.map({ (value) -> HKQuantity in
```
loudnate: You can condense this to `predicted.map { $0.quantity }` if you like.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -193,7 +195,7 @@ class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverDelegat
         }
 
         // Trigger device status upload, even if something is wrong with pumpStatus
-        remoteDataManager.uploadDeviceStatus(pumpStatus)
+        nightscoutDataManager.uploadDeviceStatus(pumpStatus)
```
loudnate: Would be nice to follow up later and move lines 184-195 into `uploadDeviceStatus`

Loop/Managers/LoopDataManager.swift
```diff
@@ -184,12 +199,13 @@ class LoopDataManager {
 
      - parameter resultsHandler: A closure called once the values have been retrieved. The closure takes the following arguments:
         - predictedGlucose:     The calculated timeline of predicted glucose values
+        - insulinOnBoard        Current insulin on board
```
loudnate: The order here doesn't match the arguments below (which is OK for Swift 2, but not for Swift 3)

Loop/Managers/LoopDataManager.swift
loudnate: We can remove this import now, yes?

loudnate:
LGTM – cut a final build of RileyLink frameworks?

loudnate:
Huzzah! Thanks for all the work you put into this.

PR #117, opened 2016-08-11
Capture loop errors and upload to NS

ps2:
Also, prevent error when no new data available from share.

### DIFF
--- a/Loop/Managers/DeviceDataManager.swift
+++ b/Loop/Managers/DeviceDataManager.swift
@@ -548,18 +548,13 @@ final class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverD
                 return (quantity: $0.quantity, date: $0.startDate, isDisplayOnly: false)
             }
 
-            if (newGlucose.count > 0) {
-
-                glucoseStore.addGlucoseValues(newGlucose, device: nil) { (_, _, error) -> Void in
-                    if let error = error {
-                        self.logger.addError(error, fromSource: "GlucoseStore")
-                    }
-
-                    NSNotificationCenter.defaultCenter().postNotificationName(self.dynamicType.GlucoseUpdatedNotification, object: self)
-
-                    completion?()
+            glucoseStore.addGlucoseValues(newGlucose, device: nil) { (_, _, error) -> Void in
+                if let error = error {
+                    self.logger.addError(error, fromSource: "GlucoseStore")
                 }
-            } else {
+
+                NSNotificationCenter.defaultCenter().postNotificationName(self.dynamicType.GlucoseUpdatedNotification, object: self)
+
                 completion?()
             }
         }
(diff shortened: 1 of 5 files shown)


Loop/Managers/LoopDataManager.swift
```diff
@@ -201,10 +201,11 @@ final class LoopDataManager {
         - recommendedTempBasal: The recommended temp basal based on predicted glucose
         - lastTempBasal:        The last set temp basal
         - lastLoopCompleted:    The last date at which a loop completed, from prediction to dose (if dosing is enabled)
+        - lastLoopError         The error, if any, from the last loop attempt
         - insulinOnBoard        Current insulin on board
         - error:                An error object explaining why the retrieval failed
      */
-    func getLoopStatus(resultsHandler: (predictedGlucose: [GlucoseValue]?, recommendedTempBasal: TempBasalRecommendation?, lastTempBasal: DoseEntry?, lastLoopCompleted: NSDate?, insulinOnBoard: InsulinValue?, error: ErrorType?) -> Void) {
+    func getLoopStatus(resultsHandler: (predictedGlucose: [GlucoseValue]?, recommendedTempBasal: TempBasalRecommendation?, lastTempBasal: DoseEntry?, lastLoopCompleted: NSDate?, lastLoopError: ErrorType?, insulinOnBoard: InsulinValue?, error: ErrorType?) -> Void) {
```
loudnate: `lastLoopError` and `error` aren't both necessary here.

ps2: Well, if enact fails, lastLoopError will be set, and update() will not throw the same error.  I suppose if update() succeeds, and lastLoopError is set, I could return that, and keep a single error argument.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -548,13 +548,18 @@ final class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverD
...
             }
 
-            glucoseStore.addGlucoseValues(newGlucose, device: nil) { (_, _, error) -> Void in
-                if let error = error {
-                    self.logger.addError(error, fromSource: "GlucoseStore")
-                }
+            if (newGlucose.count > 0) {
 
-                NSNotificationCenter.defaultCenter().postNotificationName(self.dynamicType.GlucoseUpdatedNotification, object: self)
+                glucoseStore.addGlucoseValues(newGlucose, device: nil) { (_, _, error) -> Void in
```
loudnate: oh geez I had no idea this was returning an error. I'll put a guard statement in LoopKit.

Loop/Managers/LoopDataManager.swift
loudnate: so after looking more at this, it's pretty clear I haven't designed this API well. `getLoopStatus` actually triggers an `update` cycle, which exists because I wanted it to be possible for someone who opens the UI mid-loop to see the latest prediction data while avoiding doing any mid-loop calculations while the app isn't open.

I think your suggestion is best – `error ?? lastLoopError`. Because you're checking status immediately after looping, `error` should never be `nil` unless there's a data problem, in which case it's exactly the same as `lastLoopError`.

loudnate:
Merged as part of #118

PR #126, opened 2016-08-19
Remove "beta" label from G4 Share Receiver

mddub:
No issues with this in the past two weeks for me, nor in the past week for @channemann.


loudnate:
Who hoo!

PR #128, opened 2016-08-20
Saving pump history into LoopKit and delegating upload to Nightscout

loudnate:
cc @ps2


loudnate:
I've been running this for 3 weeks, but I still don't use Nightscout. Time to head to dev and work out any issues as they arise there.

PR #132, opened 2016-08-22
Upload to loop plugin

ps2:
This populates the new structures for the Nightscout Loop plugin.


Loop/Managers/DeviceDataManager.swift
```diff
@@ -543,7 +559,7 @@ final class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverD
                 if let error = error {
                     self.logger.addError(error, fromSource: "ShareClient")
                 }
-
+                completion?()
```
loudnate: thanks for catching this

Loop/Managers/DeviceDataManager.swift
```diff
@@ -331,14 +329,32 @@ final class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverD
         // If we don't yet have pump status, or it's old, poll for it.
         if latestReservoirValue == nil || latestReservoirValue!.startDate.timeIntervalSinceNow <= -pumpStatusAgeTolerance {
             readPumpData { (result) in
+                let nsPumpStatus: NightscoutUploadKit.PumpStatus?
                 switch result {
                 case .Success(let (status, date)):
                     self.updateReservoirVolume(status.reservoir, atDate: date, withTimeLeft: nil)
+                    let battery = BatteryStatus(voltage: status.batteryVolts, status: BatteryIndicator(batteryStatus: status.batteryStatus))
+                    nsPumpStatus = NightscoutUploadKit.PumpStatus(clock: date, pumpID: status.pumpID, iob: nil, battery: battery, suspended: status.suspended, bolusing: status.bolusing, reservoir: status.reservoir)
```
loudnate: What about throwing `status` over the fence to `NightscoutUploadManager` and letting it make the decision to upload? (and store the new state on line 49)

Loop/Managers/DeviceDataManager.swift
```diff
@@ -331,14 +329,32 @@ final class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverD
...
+                    nsPumpStatus = nil
+                }
+                
+                if nsPumpStatus != nil {
+                    // Always upload new pump data
+                    self.nightscoutDataManager.uploadDeviceStatus(nsPumpStatus)
+                    self.lastDeviceStatusUpload = NSDate()
+                } else {
+                    // Only upload if it's been a while
+                    if self.lastDeviceStatusUpload == nil || -(self.lastDeviceStatusUpload!.timeIntervalSinceNow) >= NSTimeInterval(minutes: 4) {
```
loudnate: Minor, but would be good to keep the comparison consistent with the pattern on 330:

```swift
latestReservoirValue == nil || latestReservoirValue!.startDate.timeIntervalSinceNow <= -pumpStatusAgeTolerance
```

Loop/Managers/DeviceDataManager.swift
```diff
@@ -331,14 +329,32 @@ final class DeviceDataManager: CarbStoreDelegate, TransmitterDelegate, ReceiverD
...
+                        self.nightscoutDataManager.uploadDeviceStatus(nsPumpStatus)
```
loudnate: What does this method do if `nsPumpStatus` is `nil`?

loudnate: It looks like it just uploads the phone's battery level, so it acts like a heartbeat for the remote viewer?

ps2: Yes, it shows uploader battery and shows that the Loop app is running and able to upload to NS, even if loop() is progressing for some reason (more on this later).

loudnate: cool, makes total sense.

ps2:
Currently, if fetching pump (and reservoir) status fails, we do not run loop().  There are two reasons we maybe should: 1) loop() currently can still make a new prediction and recommendation if BG has been updated, and pump status is not older than 15 minutes. 2) Even if loop() cannot progress, we would like to capture the error reason and send a new loop status to NS.

So I am considering generating a PumpStatusUpdatedNotification even when fetching pump status fails. If there is a better way to do this, I'm open to it.

mhaeberli:
@ps2 - I like your idea above ```PumpStatusUpdatedNotification even when fetching pump status fails``` but have nothing better to offer at the moment ...

loudnate:
I'm not a fan of this idea if the goal is to set a temp basal. A lot of insulin can be delivered in a 5 minute window. 

If the goal is just to trigger a NS loop status upload, invoke NightscoutUploadManager to do that directly. PumpStatusUpdatedNotificafion means there are new insulin data and it should be recalculated. That isn't true in this case and I don't want to break any assumptions.

However, if we have new BGs, the prediction is already flagged for recalculation and you can trigger it from NSUploadManager by calling getLoopStatus.

ps2:
RE: triggering getLoopStatus directly, that sounds good. I'll submit another diff to cover this situation.

I agree that even 5 minutes is a long window to go without new insulin data. This line: https://github.com/loudnate/Loop/blob/master/Loop/Managers/LoopDataManager.swift#L330
seems to be indicating that it's ok to run if the data is < 15 minutes old.  In practice, since we only trigger loop() when insulin data is updated, it probably never gets used.  Just pointing it out in case it should be updated.

loudnate:
Yeah, the 15 could probably be tightened, but keep in mind it's also based
on the pump's clock so we want a little bit of error for intra-hour drift.

ps2:
Ready for review again

ps2:
Ah, maybe one more change; it appears that we're trying to fetch pump status once a minute (RL timer tick) if it's old.  Seems a bit aggressive. Thinking I should put a limit on it, like try at most once every 3 minutes.

mhaeberli:
@ps2 once every 3 minutes sounds good.  I once imagined a too-fancy approach with longer and longer times but constant is simple.

loudnate:
> it appears that we're trying to fetch pump status once a minute (RL timer tick) if it's old

I'm not sure I see the issue here with retrying every minute, especially because the solution introduces yet additional state? The RL timer tick is disabled for any user using the G4 and G5 BLE support anyhow, which is the vast majority

ps2:
Well, it ends up waking the pump, taking more execution time on the phone, and uploading more data. You're right about it being a small set of users (I'm one, but that's just because I don't wear a sensor). I can back that out, since it's not usually a normal situation for me either.

loudnate:
> Well, it ends up waking the pump, taking more execution time on the phone, and uploading more data. 

I hadn't thought about the data sent on every failure, but I did think it was unlikely that it would be able to wake the pump every minute but still fail to retrieve pump status.
Simply for the data-sending case, you could create a separate method for specifically handling the heartbeat despite comms failures, and rate-limit that from within the `NightscoutUploadManager`

loudnate:
LGTM – are you ready to merge?

ps2:
Yes, let's get this in!

PR #150, opened 2016-09-03
Added script to instruct users to run

beached:
This will allow users to easily update the MAIN_APP_BUNDLE_IDENTIFIER without confusion.  Run it, give the new handle and it will open the project
source update_id.sh

### DIFF
--- a/update_id.sh
+++ b/update_id.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+
+export PROJ_FILE="./Loop.xcodeproj/project.pbxproj"
+
+echo -n "Enter a new handle.  A single word, usually lower-case, no special characters: "
+read NEW_HANDLE
+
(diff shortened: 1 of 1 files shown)


update_id.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+
+export PROJ_FILE="./Loop.xcodeproj/project.pbxproj"
+
+echo -n "Enter a new handle.  A single word, usually lower-case, no special characters: "
+read NEW_HANDLE
+
+sed -i '' "s/MAIN_APP_BUNDLE_IDENTIFIER = com\..*\.Loop/MAIN_APP_BUNDLE_IDENTIFIER = com.${NEW_HANDLE}.Loop/g" "${PROJ_FILE}"
```
loudnate: From previous experience, it's not a good idea to `sed` the pbxproj file. It can be a bit of a wild west. However, I can move this variable out into a `xcconfig` file (which I should have done a while ago) to make this replace simpler.

loudnate:
Thanks again for this PR. In this case, I decided to go a different route to improving the setup experience.

PR #162, opened 2016-09-05
Update to RileyLink frameworks v0.11.2

### LINKED ISSUE #161: Nightscout Integration regression
I cannot see why, but since 0.11.1 the temp basal data sent to nightscout does not show up in the basal display or basal pill.  I think it is related to history upload but cannot see the relevant change in the patches.

I have re-tested with 0.11.0 and 0.10.1 and it works there.

ps2:
Fixes #161


loudnate:
Awesome, thanks.

PR #163, opened 2016-09-06
Fix for display of glucose

beached:
This should work for mg/dl and mmol/L as there is only ever 3 significant digits at most, sometimes 2 when glucose is <=99mg/dl or <=5.5mmol/L

Tested and fixes https://github.com/loudnate/Loop/issues/159


loudnate:
Thanks, though I've already got fixes in for these:
https://github.com/loudnate/Loop/commit/e5170b905eea53ea06f4576620f43f8f0920b7c2
https://github.com/loudnate/Loop/commit/5871203b5592cee156df92463af347d495c908a1

loudnate:
👍

PR #218, opened 2016-10-06
Fix Gitter Button

channemann:
Button was linked to LoopKit/LoopKit instead of the correct LoopKit/Loop

Thanks to @ldunn12 for the heads-up!


scottleibrand:
Oh, looky!  Something in the Loop repo I am actually qualified to 👍 !  ;-)

PR #228, opened 2016-10-20
Upload carbs-on-board to Nightscout

mddub:
For https://github.com/mddub/urchin-cgm/pull/50.

I would appreciate it if someone else (@ps2?) could test this before I merge. You should see something like this in `/api/v1/devicestatus.json?find[loop][$exists]=1`:

```
{
  "loop": {
    "cob": {
      "timestamp": "2016-10-20T23:40:00Z",
      "cob": 18.2929316872428
    }, ...
  }, ...
}

### DIFF
--- a/Loop/Managers/LoopDataManager.swift
+++ b/Loop/Managers/LoopDataManager.swift
@@ -76,6 +76,7 @@ final class LoopDataManager {
             center.addObserver(forName: .CarbEntriesDidUpdate, object: nil, queue: nil) { (note) -> Void in
                 self.dataAccessQueue.async {
                     self.carbEffect = nil
+                    self.carbsOnBoardSeries = nil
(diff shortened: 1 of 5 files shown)


ps2:
I see cob uploading to NS, and I see it showing in urchin!  Very nice!  It's not showing in the COB plugin in NS, but I believe that is just that NS needs to be updated to look there for Loop cob.

ps2:
Just waiting to see it update.  I reviewed the code and couldn't see anything to improve upon.  Will approve once I see the update.

ps2:
I do see some occasional issue in NS about 'cob' being undefined, but that's probably more NS issues to work through.

mddub:
Awesome! Thanks for reviewing and helping test!

Issue #141, opened 2016-08-30
Battery Status Empty


ashrind:
Any thoughts as to why my battery status might not be filling in? Running on a 722, changed the battery when I started with Loop last night
![img_7367](https://cloud.githubusercontent.com/assets/21300821/18098995/7fb8d52a-6e99-11e6-81c9-11afdf2a8c62.PNG)

loudnate:
This is a know issue with the x22 pumps. @jlucasvt has started documenting the time-voltage curve for a lithium battery. I'm hoping someone can do the same for an alkaline as well.

ashrind:
@loudnate Thanks for the answer!

jlucasvt:
I'm finally committed to gathering Alkaline data.  Should have it mapped out in about a week.

tomasboudr:
Awesome that this is getting some love! Always wondering why it was blank since you could see the voltage in pump status.

jlucasvt:
Alkaline and Lithium Battery data.

https://docs.google.com/spreadsheets/d/19FpO5YO5ewQrdph_4iIXuear6xRGzl-AYTiFeHwPcVk/edit?usp=sharing

loudnate:
Very awesome!

loudnate:
@jlucasvt my personal recommendation would be to expose a voltage-to-percentage remaining function on the PumpModel enum in MinimedKit, in the rileylink_ios repo.

jlucasvt:
Yes if we use the linear eq's it should be pretty simple.  I  I did do a quick run with excel to generate a 4th Ord Polynomial Eq for both but they may be a bit overboard..  

![screen shot 2016-09-25 at 10 45 21 am](https://cloud.githubusercontent.com/assets/19819388/18815997/a7a2f4d4-830d-11e6-9962-292219f63077.png)

loudnate:
Ideally there is no user setting for choosing their battery type.

tomasboudr:
I agree with you nate, ideally - but if its tricky to detect in software, you have to realize its a switch that will be used very rarely. I would rather time be spent on other things and there be a switch.

jlucasvt:
Yeah , My one touch ping made you choose on battery swap (ALK or LITH).  It will be ideal to not worry about the battery type.  Something that can be improved if we want to track battery swaps and actual decay.  

In the past I wired this up in the DeviceDataManager.assertCurrentPumpData just after batterystatus is set and before NSUpload.
  
```
[...]

PR #241, opened 2016-10-30
Add x22 Battery Status

jlucasvt:
Adds Battery Status using Linear Alkaline and Lithium Decay #141 . Let me know if there is a more structured approach to adding this support. I may need help finding my way adding support to frameworks.

### DIFF
--- a/Loop/Extensions/NSUserDefaults.swift
+++ b/Loop/Extensions/NSUserDefaults.swift
@@ -31,6 +31,7 @@ extension UserDefaults {
         case PumpRegion = "com.loopkit.Loop.PumpRegion"
         case PumpTimeZone = "com.loudnate.Naterade.PumpTimeZone"
         case RetrospectiveCorrectionEnabled = "com.loudnate.Loop.RetrospectiveCorrectionEnabled"
+        case BatteryIsLithium = "com.jlucasvt.Loop.BatteryIsLithium"
     }
 
     var basalRateSchedule: BasalRateSchedule? {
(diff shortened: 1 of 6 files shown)


Loop/Extensions/NSUserDefaults.swift
```diff
@@ -31,6 +31,7 @@ extension UserDefaults {
         case PumpRegion = "com.loopkit.Loop.PumpRegion"
         case PumpTimeZone = "com.loudnate.Naterade.PumpTimeZone"
         case RetrospectiveCorrectionEnabled = "com.loudnate.Loop.RetrospectiveCorrectionEnabled"
+        case BatteryIsLithium = "com.jlucasvt.Loop.BatteryIsLithium"
```
loudnate: Please use `com.loopkit.Loop.` for new keys going forward.

ps2:
I think the UI would be more clear with a choice between Alkaline and Lithium.  There is already a RadioSelectionTableViewController that is used for "Preferred Insulin Source", and that would also give some room to explain why the user may need to make a choice.

Loop/Managers/LoopDataManager.swift
```diff
@@ -39,6 +39,14 @@ final class LoopDataManager {
             notify(forChange: .preferences)
         }
     }
+    
+    var batteryIsLithium: Bool {
```
ps2: Since this variable isn't related to Loop calculations, it probably makes more sense to have it on DeviceDataManager.

jlucasvt: Yes with the above UI recommendation to select a battery chemistry type, I think that batteryIsLithium will be unused and can be removed.

jlucasvt:
Closing and Updating with new code selection code this week.

PR #259, opened 2016-11-15
Dev 1.1.1 x22 battery status

jlucasvt:
New PR for #141 Updated to now allow for discrete selection of Alkaline and LIthium battery chemistry.  View Controller and Enum added to allow for named selection following pattern of preferredInsulinDataSource.  

img_6206

### DIFF
--- a/Loop/Managers/DeviceDataManager.swift
+++ b/Loop/Managers/DeviceDataManager.swift
@@ -698,6 +698,13 @@ final class DeviceDataManager: CarbStoreDelegate, CarbStoreSyncDelegate, DoseSto
             UserDefaults.standard.preferredInsulinDataSource = preferredInsulinDataSource
         }
     }
+    
+    /// The Default battery chemistry is Alkline
+    var batteryChemistry = UserDefaults.standard.batteryChemistry ?? .alkline {
+        didSet {
+            UserDefaults.standard.batteryChemistry = batteryChemistry
+        }
+    }
 
     // MARK: G5 Transmitter
 
--- a/Loop/Managers/DeviceDataManager.swift
+++ b/Loop/Managers/DeviceDataManager.swift
@@ -353,6 +353,12 @@ final class DeviceDataManager: CarbStoreDelegate, CarbStoreSyncDelegate, DoseSto
                 case .success(let (status, date)):
                     self.updateReservoirVolume(status.reservoir, at: date, withTimeLeft: nil)
                     let battery = BatteryStatus(voltage: status.batteryVolts, status: BatteryIndicator(batteryStatus: status.batteryStatus))
+                    
+                    //Non MySentry Battery Status #141
+                    if let sentrySupported = self.pumpState?.pumpModel?.hasMySentry , !sentrySupported {
+                        self.setBatteryStatusforNonMySentryPumps(currVoltage: status.batteryVolts)
+                    }
+                    
                     nsPumpStatus = NightscoutUploadKit.PumpStatus(clock: date, pumpID: status.pumpID, iob: nil, battery: battery, suspended: status.suspended, bolusing: status.bolusing, reservoir: status.reservoir)
                 case .failure(let error):
                     self.troubleshootPumpComms(using: device)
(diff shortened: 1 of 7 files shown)


Loop/Managers/DeviceDataManager.swift
```diff
@@ -364,6 +370,38 @@ final class DeviceDataManager: CarbStoreDelegate, CarbStoreSyncDelegate, DoseSto
...
+    private func setBatteryStatusforNonMySentryPumps(currVoltage : Double){
+        var minVoltage : Double
+        var maxVoltage : Double
+        var batteryNotification : Double
+     
+        // if Lithium set min and max linear voltages
+        if (self.batteryChemistry == .lithium){
+            minVoltage = 1.32
+            maxVoltage = 1.58
+            batteryNotification = 0.12
```
ps2: The point at which we display 0% battery should be equal the point at which a notification should be made, no?

Loop/Managers/DeviceDataManager.swift
```diff
@@ -364,6 +370,38 @@ final class DeviceDataManager: CarbStoreDelegate, CarbStoreSyncDelegate, DoseSto
         }
     }
 
+    /// NonMySentry Battery Calculation for Alkaline and Lithuim  #141
+    ///
+    /// - parameter currVoltage: Current Voltage Reading from Pump
+    ///
+    public var x22BatteryPercentRemaining : Double = -1
```
ps2: Since we already have a battery % for x23 devices, let's track battery % as a single ivar.

Updating that ivar can trigger both battery change event (`AnalyticsManager.sharedManager.pumpBatteryWasReplaced()`) and notification (`NotificationManager.sendPumpBatteryLowNotification()`), and then the StatusTableViewController can check one place for battery percent, instead of checking both `dataManager.latestPumpStatusFromMySentry.batteryRemainingPercent` and `dataManager.x22BatteryPercentRemaining`

Loop/View Controllers/BatteryTypeSelectionTableViewController.swift
```diff
@@ -0,0 +1,86 @@
+//
+//  BatteryTypeSelectionTableViewController.swift
+//  Loop
```
ps2: Was it not possible to re-use RadioSelectionTableViewController?

Loop/Managers/DeviceDataManager.swift
loudnate: Don't use magic number sentinels; use an optional. 
Since percentage is derived from chemistry, and chemistry can be changed at any time, this shouldn't be stored at all.

Loop/Managers/DeviceDataManager.swift
jlucasvt: The intent of this Notification was to "warn" a user of an impending battery failure or intermittent communications failures due to low battery.  This is a 8 to 12 hour warning prior to battery = 0.

https://docs.google.com/spreadsheets/d/19FpO5YO5ewQrdph_4iIXuear6xRGzl-AYTiFeHwPcVk/edit#gid=246762532.

Loop/View Controllers/BatteryTypeSelectionTableViewController.swift
jlucasvt: I am sure it may be. I am just not swift enough to fully comprehend how.  So I leveraged the existing pattern Nate had provided.

Loop/Managers/DeviceDataManager.swift
jlucasvt: As I understand it you are saying I should write the Battery % Remaining to dataManager.latestPumpStatusFromMySentry.batteryRemainingPercent.  I totally agree reuse of the already existing structure would be great, however I am unsure how to write to the batteryRemainingPercent var.

`latestPumpStatusFromMySentry?.batteryRemainingPercent = ((currVoltage - minVoltage)/(maxVoltage - minVoltage))`

I receive an error

`Cannot assign to property: 'batteryRemainingPercent' is a 'let' constant`

jlucasvt:
Closed for #275

Issue #276, opened 2016-11-22
Create a lock screen widget to show basic info


bharat:
I've done some basic research on this and it looks like it should be relatively straightforward to create a "Today Extension" that shows basic glucose, IOB, COB info on the iOS lock screen. Creating this issue as a way to track and collaborate any work on this project. I plan to work on this in the near term (but we'll see how that goes).

elnjensen:
I think this is a great idea; I had thought of it before but didn't follow through.  I think something similar to the HUD (top row of icons) in Loop could work well.  Maybe even some of that code could be reused?

Actually maybe the first three icons (loop symbol, BG + trend, current temp rate) and then replace next two with IOB and COB?  Reservoir and battery status aren't as important for quick look. 

What were you thinking of @bharat?

bharat:
For me, the current BG and the graph are the most useful - but I'm open to ideas here.

elnjensen:
I'm ałready using the lock screen widget from the Dexcom G5 app, and those are the two things it shows already, graph and current BG: https://www.dexcom.com/g5-app-update   I guess G4 users wouldn't have those, so this could replace that. But coming from G5, it would add more value if it also included some basic Loop info.

bharat:
Ok - I've got a basic version of this up and running. Right now it only shows the current blood glucose, but it should be relatively easy to add the rest now that I've got the plumbing in place.

Code is here:
https://github.com/bharat/Loop/tree/lock-screen-widget

Currently looks like this:
![Uploading Screen Shot 2016-11-25 at 7.12.16 PM.png…]()

More to come...

elnjensen:
Excellent!  Let me know if you want me to test something at any point.  Thanks for your work on this.

bharat:
Ok - I have all the plumbing in now for the HUD. I'm going to test it out with different units (thus far only been playing with mg/DL) then submit a PR and let somebody pick through it and tell me what to change :-)

Feel free to play with https://github.com/bharat/Loop/tree/lock-screen-widget if you'd like to try it on your iPhone..

![screen shot 2016-11-26 at 5 13 13 pm](https://cloud.githubusercontent.com/assets/54570/20644854/b4726cc0-b3fb-11e6-832c-e128c3d19015.png)

loudnate:
This is wonderful, thank you!

I would recommend at some point that you refactor the HUD (and maybe chart) views into an iOS framework, rather than duplicate them between targets. 

[...]

PR #281, opened 2016-11-27
Create a lock screen widget (aka Today Extension)

bharat:
Show the standard heads up display, in a widget, available on the lock screen.

The majority of the work is in a new "Today Extension" target. It's relatively straightforward, but some notes to assist in reading/reviewing:

1. Loop and the extension are now in the same app group so that they can share context
2. There's a new TodayExtensionContext class that stores a relatively minimal set of data to power the HUD
3. There's a new TodayExtensionDataManager that watches for changes and populates the context class
4. The HUD views are shared between the main app and the extension
5. Some small areas of Loop required refactoring (eg, I moved the net basal calculations out of the StatusTableViewController
6. It's so simple that I didn't bother making subdirectories 

ref: #276


Loop TodayExtension/TodayExtensionContext.swift
```diff
@@ -0,0 +1,154 @@
...
+    var trendType: GlucoseTrend?
+    var isLocal: Bool
+}
+
+struct GlucoseContext {
+    var latest: GlucoseValue
+    var sensor: SensorDisplayable?
+}
+
+class TodayExtensionContext {
```
loudnate: This class is trying to do more than one thing. On one hand you've got convenience methods on NSUserDefaults which are better in an `extension UserDefaults`. On the other hand you've got raw data serialization which belongs in struct that implements `RawRepresentable`.

Loop TodayExtension/TodayExtensionContext.swift
```diff
@@ -0,0 +1,154 @@
...
+    var data: [String:Any] = [:]
+    let storage = UserDefaults(suiteName: "group.com.loudnate.Loop")
+    
+    var glucose: GlucoseContext? {
+        get {
+            if data["gcgv"] == nil { return nil }
```
loudnate: Use a `guard` statement.

Loop TodayExtension/TodayExtensionContext.swift
```diff
@@ -0,0 +1,154 @@
...
+            
+            var sensor: SensorDisplayableContext? = nil
+            if data["gcsv"] != nil {
```
loudnate: Optional unwrapping is a much clearer approach. There should hardly ever be a case for using `as!`

Loop TodayExtension/TodayExtensionContext.swift
```diff
@@ -0,0 +1,154 @@
...
+            data["nbp"] = b?.percentage
+            data["nbsd"] = b?.startDate
+        }
+    }
+    
+    var eventualGlucose: String? {
+        get {
+            return data["eg"] as? String
+        }
+        set(eg) {
```
loudnate: Don't use cryptic variable names. You can drop "(eg)" and reference `newValue` in the implementation.

Loop TodayExtension/TodayExtensionContext.swift
loudnate: Also, it's unlikely this community will ever agree on a single interface, so this should be named more specifically to allow more Widgets in the future.

Loop TodayExtension/Info.plist
```diff
@@ -0,0 +1,31 @@
...
+	<key>CFBundleIdentifier</key>
+	<string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
+	<key>CFBundleInfoDictionaryVersion</key>
+	<string>6.0</string>
+	<key>CFBundleName</key>
+	<string>$(PRODUCT_NAME)</string>
+	<key>CFBundlePackageType</key>
+	<string>XPC!</string>
+	<key>CFBundleShortVersionString</key>
+	<string>1.0</string>
```
loudnate: This should match the container version. Version strings are set in concert using `agvtool`

Loop.xcodeproj/project.pbxproj
```diff
@@ -235,6 +265,17 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
+		4F70C1EC1DE8DCA8006380B7 /* Embed App Extensions */ = {
+			isa = PBXCopyFilesBuildPhase;
+			buildActionMask = 2147483647;
+			dstPath = "";
+			dstSubfolderSpec = 13;
+			files = (
+				4F70C1E81DE8DCA7006380B7 /* Loop TodayExtension.appex in Embed App Extensions */,
```
loudnate: Similar to the comment above, this should get a more specific name.

Loop TodayExtension/TodayExtensionContext.swift
```diff
@@ -0,0 +1,154 @@
...
+        set(b) {
+            data["nbr"] = b?.rate
+            data["nbp"] = b?.percentage
+            data["nbsd"] = b?.startDate
+        }
+    }
+    
+    var eventualGlucose: String? {
+        get {
+            return data["eg"] as? String
```
loudnate: Don't use cryptic dictionary keys here. Match the property names exactly. I chose to use these on the watch dictionaries because they're being shuttled over a low-bandwidth Bluetooth connection.

Loop TodayExtension/TodayViewController.swift
```diff
@@ -0,0 +1,81 @@
...
+class TodayViewController: UIViewController, NCWidgetProviding {
+    
+    @IBOutlet weak var loopCompletionHUD: LoopCompletionHUDView!
+    @IBOutlet weak var glucoseHUD: GlucoseHUDView!
+    @IBOutlet weak var basalRateHUD: BasalRateHUDView!
+    @IBOutlet weak var reservoirVolumeHUD: ReservoirVolumeHUDView!
+    @IBOutlet weak var batteryHUD: BatteryLevelHUDView!
+    @IBOutlet weak var subtitleLabel: UILabel!
+    
+    override func viewDidLoad() {
```
loudnate: Remove template boilerplate here and below.

Loop TodayExtension/TodayViewController.swift
```diff
@@ -0,0 +1,81 @@
...
+            if let eventualGlucose = context.eventualGlucose {
+                subtitleLabel.text = eventualGlucose
+                subtitleLabel.textColor = UIColor.secondaryLabelColor
+                subtitleLabel.alpha = 1
+            } else {
+                subtitleLabel.alpha = 0
+            }
+            
+            
+            completionHandler(NCUpdateResult.newData)
```
loudnate: You need to call the completion handler in all paths.

Loop TodayExtension/Base.lproj/MainInterface.storyboard
```diff
@@ -0,0 +1,334 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
loudnate: Per my comment on the issue, ideally these views aren't duplicated. You can create a new framework which contains a nib with these heirarchies and the view classes and share it between targets. This will allow use in future extension targets like notification content.

loudnate:
Just because it's not documented in this thread, let me just say Thank You again for this awesome contribution!

bharat:
Thanks for the feedback, Nate! I'll go through and make point fixes, but a couple of general broader comments:

1. I'd like to refactor the views into their own framework in a separate PR, partly because I don't know how to do that yet (happy to learn, but am still new to iOS development), and partly because that's a self-contained change that's unrelated to this extension. I can either do it first, or I can do it afterwards - would prefer to do it afterwards but happy to put this aside and do that first if you'd like.

2. Re: naming the extension - what do you suggest naming it? "WatchApp" and "WatchApp Extension" seemed pretty generic so I followed that pattern. It doesn't seem like XCode is going to make it easy for me to rename it, so I want to do it once :-) Let me know what you'd like it named and I'll take care of it. I imagine that "Lock Screen Widget" is too generic, but maybe "HUD Extension" or "HUD Widget" might be more descriptive? Or "Dashboard Widget". Sigh. Naming is one of the http://martinfowler.com/bliki/TwoHardThings.html

Also, FWIW - there's a bug in the way that it's saving context right now causing the app to crash.. so I wouldn't deploy it yet... :-)

loudnate:
Re: 1. I think that's totally fine. Thanks for the courage to submit the PR and the humility to take the feedback from folks much more immersed in the ecosystem. Definitely helps everybody learn!
Re: 2. The difference with [WatchApp + WatchApp Extension] is that it's a 1:1 relationship. I suggested the naming change because it's possible to have multiple "Today" widgets, and the future framework code (views + view models) might not be specific to just the widgets. Something along the lines of "Status"/"HUD" feels consistent with the existing naming, I guess. I personally think there's nothing wrong with as much `git mv` as you like, so manage the naming when it feels natural, and it's totally fine to change your mind later. Renaming targets is possible in Xcode, but you might need to do some grep-ing to make sure everything is properly transitioned.

Re: crash, hopefully my feedback helps to sort that out, and feel free to share if you're stuck :) I'm an arms-length away from deployment, but in my excitement I can't help but giving feedback, especially since I've had real-world with refactoring extension targets in the past.

Loop TodayExtension/TodayExtensionContext.swift
```diff
@@ -42,113 +42,115 @@ struct GlucoseContext {
     var sensor: SensorDisplayable?
 }
 
-class TodayExtensionContext {
-    var data: [String:Any] = [:]
-    let storage = UserDefaults(suiteName: "group.com.loudnate.Loop")
+final class TodayExtensionContext: NSObject, RawRepresentable {
```
loudnate: No need to extend `NSObject`. And this is probably more canonically a struct.

Loop TodayExtension/TodayExtensionContext.swift
```diff
@@ -42,113 +42,115 @@ struct GlucoseContext {
...
-        }
-        set(lgv) {
-            data["gcgv"] = lgv?.latest.quantity.doubleValue(for: HKUnit.milligramsPerDeciliterUnit())
-            data["gcgd"] = lgv?.latest.startDate
-            data["gcsv"] = lgv?.sensor?.isStateValid
-            data["gcsd"] = lgv?.sensor?.stateDescription
-            data["gcst"] = lgv?.sensor?.trendType
-            data["gcsl"] = lgv?.sensor?.isLocal
-        }
+    var latestGlucose: GlucoseContext?
```
loudnate: Use `let` for all set-once properties throughout this file

Loop TodayExtension/Base.lproj/MainInterface.storyboard
bharat: Great - per our discussion I'm deferring this work and have created #284 to track progress.

Loop TodayExtension/TodayViewController.swift
bharat: fixed.

Loop.xcodeproj/project.pbxproj
bharat: fixed.

Loop TodayExtension/TodayExtensionContext.swift
bharat: I've completely overhauled this class to separate out the functionality.

bharat:
Ok - I believe I've addressed all the comments in your last code review. Thanks for the helpful dialog - I'm new to iOS development so I'm happy to take whatever guidance I can get!

The main change (aside from renaming everything) was to overhaul the context class and separate out its functions while preserving a relatively clean API. FWIW, it seems like both context classes could usefully be replaced by protocol buffers or something similar.

Also, what should I do about unit tests here? I'm not changing any core functionality but since the context class is hand rolled, I could see a unit test that verifies its functionality.

I'm happy to make any further changes, regardless of how nitpicky. Let me know what you find!

Loop TodayExtension/TodayExtensionContext.swift
bharat: I tried that and get the following errors:

1. `StatusExtensionContext.swift: Property 'self.latestGlucose' not initialized at super.init call`
2. `StatusExtensionDataManager.swift: Cannot assign to property: 'latestGlucose' is a 'let' constant`

ideas?

bharat:
oh - and I'm also getting this warning for all framework extensions currently:

`ld: warning: linking against a dylib which is not safe for use in application extensions: Loop/Carthage/Build/iOS/LoopKit.framework/LoopKit`

From some quick searching, it looks like we'd need to flip a bit on the frameworks, but I'm no expert on this. Happy to file an issue for this to track it if it's not something that we can change inside the extension itself.

Loop/Managers/StatusExtensionDataManager.swift
```diff
@@ -0,0 +1,95 @@
...
+                    capacity: capacity)
+            }
+            
+            if let batteryPercentage = dataManager.latestPumpStatusFromMySentry?.batteryRemainingPercent {
+                context.batteryPercentage = Double(batteryPercentage) / 100.0
+            }
+        
+            if let lastPoint = predictedGlucose?.last {
+                context.eventualGlucose = String(
+                    format: NSLocalizedString("Eventually %@", comment: "The subtitle format describing eventual glucose. (1: localized glucose value description)"),
```
loudnate: This should be done in either the view or view controller.

Loop/Managers/StatusExtensionDataManager.swift
```diff
@@ -0,0 +1,95 @@
...
+//  Created by Bharat Mediratta on 11/25/16.
+//  Copyright © 2016 Nathan Racklyeft. All rights reserved.
+//
+
+import HealthKit
+import UIKit
+import CarbKit
+import LoopKit
+
+final class StatusExtensionDataManager: NSObject {
```
loudnate: No need to subclass `NSObject` here.

Loop/Managers/StatusExtensionDataManager.swift
```diff
@@ -0,0 +1,95 @@
...
+    unowned let dataManager: DeviceDataManager
+    private var lastContext: StatusExtensionContext?
```
bharat: yeah - vestigial. looking forward I should be keeping the last context for comparison purposes but I'll get rid of it for now.

Loop Status Extension/StatusExtensionContext.swift
```diff
@@ -0,0 +1,156 @@
...
+    }
+    
+    required init?(rawValue: RawValue) {
+        super.init()
+        let raw = rawValue
+        
+        if let unitString = raw["latestGlucose_unit"] as? String,
+           let latestValue = raw["latestGlucose_value"] as? Double,
+           let startDate = raw["latestGlucose_startDate"] as? Date {
+            latestGlucose = GlucoseContext(
```
loudnate: You could instead let each of these structs implement `RawRepresentable`, and replace the underscore namespacing with nested dictionaries.

It can sometimes be cleaner to read to use extensions to conform to protocols, rather than one long type definition, e.g.:

`extension GlucoseContext: RawRepresentable {`

Loop/Managers/StatusExtensionDataManager.swift
bharat: I'm struggling with this. My original idea was to pass the raw value through and let the extension turn it into a string, but that resulted in a lot of more complex objects (eg. HKQuantity) getting pushed through the context. My thinking was that the status dashboard can deal with a String blob so I'd turn it into a string here. Can't do that in a View/ViewController on the Loop side since I don't want the StatusExtensionDataManager to depend on any of the views. Hence, I'm doing it this way.

Open to ideas for how to proceed. For what it's worth, the current approach is flawed - I haven't figured out how to remove the double's precision from the resulting value. So right now I'm getting context strings like "Eventually 92.1283 mg/dL" on my actual device :-/

I've tried extending HKQuantity to give it a better description, but !@$@# HKQuantity doesn't let you have access to its underlying unit so you can't actually improve the description. Oy. Hoping you have ideas!

Loop/Managers/StatusExtensionDataManager.swift
bharat: fixed.

Loop Status Extension/StatusExtensionContext.swift
bharat: yeah, I toyed with that because I would very much like to use nested dictionaries. But then I wound up with less namespacing issues in the strings, but more boilerplate code to implement RawRepresentable in each class. I figured I'd go for less code, but happy to do it with nested RawRepresentable dicts if you think it's the right way to go.

Loop/Managers/StatusExtensionDataManager.swift
loudnate: I think maybe the hangup is the use of `LoopKit`'s protocols in these context objects, e.g. `ReservoirValue`. Unless parts of LoopKit are refactored to support the limited API allowed in extensions, you'll need to remove all the custom framework dependencies from the widget extension.

bharat: sorry, I didn't follow your last comment - not sure if it was a gihub glitch, or did you cross comment streams? I think your comment might have been in regard to the linker warning we're seeing, but this comment stream is about the eventual glucose string.

re: eventual glucose string, one approach is to pull apart the HKQuantity into a double value and a unit string. Problem with that is that I can't seem to find any way to extract the unit type from the HKQuantity. I'll go ahead and do that and hardcode it to mg/dL for now and we'll do the formatting on the far side in the ViewController.

ps2:
Could you rebase against dev to fix the conflicts?  Looks like it's caused by the translation changes.

bharat:
Not sure how to do that, can you advise? I've done two things:

1. `git merge dev` to pull changes between master and dev into my branch.
2. in this PR, I changed the target to be Loopkit:dev

happy to take whatever steps you'd like. I'm reasonably experienced with git, but I haven't done a lot of rebasing between branches. So pointers to documentation, or command lines to run, etc. are all fine.

Also happy to squash down my many commits into just one and force push back to this branch, but I was holding off on doing that because I have no idea what github will do with all of the various context that's gone into this code review so far.

bharat:
ok.. sorted it out and rebased against dev. Let me know what else you'd like me to do!

bharat:
hey, fwiw - Travis shows that the checks passed, but when I look at the Travis logs I see a bunch of errors:
https://travis-ci.org/LoopKit/Loop/builds/180501249

what gives?

loudnate:
The logs are complaining about Swift version mismatch. travis.yml needs to have its osx_image updated. Pity the build script still exits 0 though.

bharat:
Filed #300 and #301 to track the travis issue.

@ps2 - this is ready for another look.

ps2:
Looks like I have to figure out how to add an app group to my app ids. And all the other users will need to do the same.  Correct me if I'm wrong.  I'll take notes as I go, unless you have steps ready.

Do you know if this is something xcode will provision properly if it's the first time the app-ids and provisioning profiles are created?

ps2:
I've been trying this out.  Looks great!  It took a little while to update first time, but that may have been lag from using the wrong suiteName in the userdefaults extension.  Would love to see if there is a better way for managing the app group name (ideally setting it from xcconfig), but if you can't find anything either, I'll just drop it, and merge, and we'll update the docs with instructions on how to set it up.

Loop.xcodeproj/project.pbxproj
```diff
@@ -1363,13 +1522,14 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CODE_SIGN_ENTITLEMENTS = Loop/Loop.entitlements;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
-				DEVELOPMENT_TEAM = "";
+				DEVELOPMENT_TEAM = R6DR6B5C8B;
```
ps2: This needs to be set back to "", and there is one more below.

bharat: fixed. sorry about that - hard to keep my branches straight when I'm deploying :-)

bharat:
@ps2 - regarding the suiteName - I'm not sure how to actually let that be changeable in the code. I've experimented with using a variable in the entitlement based off of the $(MAIN_APP_BUNDLE_IDENTIFIER) value in Loop.xcconfig, but I also need to set the value in StatusExtensionContext.swift, and I don't know how to get that String value to draw from Loop.xcconfig. I've tried using preprocessor macros, but they don't seem to carry a value, only their existence. Open to any thoughts there. The good news is that the suiteName is arbitrary as far as I can tell so using group.com.loudnate.xxx works fine.

On the issue of lag, currently I update the context in the StatusExtensionDataManager only when a .LoopDataUpdated notification, similar to what WatchDataManager does. So initially it takes a little while for LoopDataManager.update() to finish and send out the notification. Can you suggest a better approach?

bharat:
great - @loudnate can you weigh in a +1 as well? I believe I've resolved all of the issues you've raised.. thanks.

Loop/Views/GlucoseHUDView.swift
```diff
@@ -65,19 +64,19 @@ final class GlucoseHUDView: HUDView {
         }
     }
 
-    func set(_ glucoseValue: GlucoseValue, for unit: HKUnit, from sensor: SensorDisplayable?) {
+    func set(glucoseQuantity: Double, at glucoseStartDate: Date, for unitDisplayString: String, from sensor: SensorDisplayable?) {
```
loudnate: Style nit from Swift's API design guidelines: `for unitDisplayString` should be just `unitDisplayString`, because the `String` type doesn't hold semantic meaning.

Loop Status Extension/Loop Status Extension.entitlements
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>com.apple.security.application-groups</key>
+	<array>
+		<string>group.com.loudnate.Loop</string>
```
loudnate: `group.$(MAIN_APP_BUNDLE_IDENTIFIER)`

loudnate:
> I can't find anything for setting this programmatically. It looks like users will have to update three new places when they build: Loop.entitlements, Loop Status Extension.entitlements, and NSUserDefaults+StatusExtension.swift

Entitlements files support build-setting substitutions, and accessing Info.plist can be done via the `NSBundle` API, which should make all the hard-coding avoidable and add no new setup steps for users.

Loop/Loop.entitlements
```diff
@@ -4,5 +4,9 @@
 <dict>
 	<key>com.apple.developer.healthkit</key>
 	<true/>
+	<key>com.apple.security.application-groups</key>
+	<array>
+		<string>group.com.loudnate.Loop</string>
```
loudnate: Same here.

Loop/Extensions/NSUserDefaults+StatusExtension.swift
```diff
@@ -0,0 +1,33 @@
...
+//  Created by Bharat Mediratta on 11/27/16.
+//  Copyright © 2016 LoopKit Authors. All rights reserved.
+//
+
+import Foundation
+
+extension UserDefaults {
+    
+    private enum Key: String {
+        case StatusExtensionContext = "com.loudnate.Loop.StatusExtensionContext"
```
loudnate: Please use `com.loopkit.Loop` for new keys.

Loop/Extensions/NSUserDefaults+StatusExtension.swift
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    static func shared() -> UserDefaults? {
```
loudnate: This should be a class var like `standard` is rather than a static function. Also, `shared` isn't descriptive enough IMO. `group` or `appGroup` would make more sense.

Loop/Views/GlucoseHUDView.swift
bharat: fixed

Loop/Models/NetBasal.swift
```diff
@@ -0,0 +1,41 @@
...
+//
+//  Created by Bharat Mediratta on 12/7/16.
+//  Copyright © 2016 LoopKit Authors. All rights reserved.
+//
+
+import Foundation
+import InsulinKit
+import LoopKit
+
+class NetBasal {
```
loudnate: Struct would be preferred.

Loop/Models/NetBasal.swift
```diff
@@ -0,0 +1,41 @@
...
+    var rate: Double
```
loudnate: Use `let`

Loop/Extensions/NSUserDefaults+StatusExtension.swift
```diff
@@ -0,0 +1,35 @@
...
+
+extension UserDefaults {
+    
+    private enum Key: String {
+        case StatusExtensionContext = "com.loopkit.Loop.StatusExtensionContext"
+    }
+
+    class var appGroup: UserDefaults? {
+        get {
+            return UserDefaults(suiteName: "group.com.loudnate.Loop")
```
loudnate: This still needs to reflect the correct group ID. Use the `NSBundle` APIs to access the Info.plist keys/values. Because you can't ask which bundle UserDefaults belongs to, each client will need to determine this using a bundle-specific class like a view controller.

```swift
Bundle(for: MyClass.Type).object(forInfoDictionaryKey: "AppropriateKey")
```
Except I'd recommend creating a typed runtime property as an extension on `Bundle` to keep the key in one place, e.g. `.appGroupSuiteName` or something. 

Of course that means this property now has to take an argument... sorry about that.

bharat:
ok - I think I've resolved all your recent comments. Thanks for the detailed explanation of how to access Info.plist variables from the code!

One thing I did which might be slightly funky is that I created a 2nd Bundle extension in Loop to separate out the generally useful Bundle extensions (.shortVersionString, .bundleDisplayName, etc) from .appGroupSuiteName which has to be different for each target. I believe that this is necessary since otherwise I'd have two extensions setting the same variable (which I believe is a no-no) but happy to revise if this is not the right approach.

Oh, and I also had to add MainAppBundleIdentifier to the Info.plist for Loop and Loop Status Extension since I couldn't find another variable to draw from.

Done for now. Let me know what else I can change!

bharat:
Hm. Can you be a little more explicit? I'm trying to fill in the gaps here, but I'm not following you. Based on your last message, here's what I'm receiving:

1. I delete the NSBundle+ extensions I created
1. I add an instance property called `appGroupSuiteName` to Loop/Extensions/NSBundle.swift
1. The property's getter takes an argument (seems weird to me, but maybe that's normal in Swift?)
1. Whenever I request NSUserDefaults.statusExtensionContext, I pass in a view controller
1. NSUserDefaults.statusExtensionContext then passes that view controller to the .appGroup class var which in turns passes it to the Bundle extension which finds the right bundle and returns the right appGroupSuiteName

Is that right? It seems a little convoluted to me, so I assume that this isn't what you intend. Happy to code it up any way that you want, but need a little more clarity first... From my perspective, it would be really helpful to me if you'd write out the specific APIs you'd like to see, and the files you'd like to see them in. thanks!

Loop Status Extension/NSBundle+LoopStatusExtension.swift
```diff
@@ -0,0 +1,16 @@
...
+//  Loop
+//
+//  Created by Bharat Mediratta on 12/7/16.
+//  Copyright © 2016 LoopKit Authors. All rights reserved.
+//
+
+import Foundation
+
+extension Bundle {
+    static var appGroupSuiteName: String {
```
loudnate:
```swift
private var mainAppBundleIdentifier: String? {
    return object(forInfoDictionaryKey: "MainAppBundleIdentifier") as? String
}

var appGroupSuiteName: String {
    return "group.\(mainAppBundleIdentifier!)"
}
```

Loop/Extensions/NSUserDefaults+StatusExtension.swift
```diff
@@ -0,0 +1,35 @@
...
+
+import Foundation
+
+extension UserDefaults {
+    
+    private enum Key: String {
+        case StatusExtensionContext = "com.loopkit.Loop.StatusExtensionContext"
+    }
+
+    class var appGroup: UserDefaults? {
```
loudnate: You can strike this.

Loop Status Extension/StatusViewController.swift
```diff
@@ -0,0 +1,93 @@
...
+        subtitleLabel.alpha = 0
+        subtitleLabel.textColor = UIColor.secondaryLabelColor
+    }
+    
+    override func didReceiveMemoryWarning() {
+        super.didReceiveMemoryWarning()
+    }
+    
+    func widgetPerformUpdate(completionHandler: (@escaping (NCUpdateResult) -> Void)) {
+        guard let context = UserDefaults.appGroup?.statusExtensionContext else {
```
loudnate:
```swift
let bundle = Bundle(for: type(of: self))
guard let context = UserDefaults(suiteName: bundle.appGroupSuiteName)?.statusExtensionContext else {
    // ...
```

Loop/Managers/StatusExtensionDataManager.swift
```diff
@@ -0,0 +1,100 @@
...
+        NotificationCenter.default.addObserver(self, selector: #selector(update(_:)), name: .LoopDataUpdated, object: deviceDataManager.loopManager)
+    }
+
+    @objc private func update(_ notification: Notification) {
+
+        self.dataManager.glucoseStore?.preferredUnit() { (unit, error) in
+            if error == nil, let unit = unit {
+                self.createContext(unit) { (context) in
+                    if let context = context {
+                        UserDefaults.appGroup?.statusExtensionContext = context
```
loudnate:
```swift
let bundle = Bundle(for: type(of: self))  // or Bundle.main might be cleaner
UserDefaults(suiteName: bundle.appGroupSuiteName)?.statusExtensionContext = context
```

Loop/Extensions/NSBundle+Loop.swift
```diff
@@ -0,0 +1,16 @@
...
+//  NSBundle+Loop.swift
+//  Loop
+//
+//  Created by Bharat Mediratta on 12/7/16.
+//  Copyright © 2016 LoopKit Authors. All rights reserved.
+//
+
+import Foundation
+
+extension Bundle {
```
loudnate: Strike this file

Loop Status Extension/NSBundle+LoopStatusExtension.swift
loudnate: You can share this code between targets now (and it can go in the follow-up framework refactor later)

bharat:
thanks for the clear explanation. This is a much better approach than what I had before. I like that we're clearly separating the responsibility of accessing known values from UserDefaults from the responsibility of choosing the right bundle. One note: I used Bundle.main in both the main app and the extension since according to the docs it should always work:

https://developer.apple.com/reference/foundation/bundle
`The main bundle represents the bundle directory that contains the currently executing code. `

Ok - let me know what else you'd like to see!

loudnate:
There's two open comments on how NetBasal should be an immutable value type, but after that things look good. Please make sure to test in both Debug and Release build configurations so no users are caught off-guard

bharat:
Oops - sorry about that. I committed the NetBasal fix to the wrong branch. Merged it in correctly now - let me know if you need further work there.

What is the right way to test in Debug and Release build configs? The Loop scheme specifies Debug for the Run action and that's what I've generally been using. Should I temporarily switch to Release in that scheme or is there some better way?

loudnate:
Run Product > Archive, then either Upload to the App Store for TestFlight
distribution or export and Ad Hoc build and install using the Devices
window.

There should be some folks in the Gitter channel who can help with this
step too.

Loop Status Extension/StatusExtensionContext.swift
```diff
@@ -0,0 +1,162 @@
...
+}
+
+struct NetBasalContext {
+    var rate: Double
+    var percentage: Double
+    var startDate: Date
+}
+
+struct SensorDisplayableContext: SensorDisplayable {
+    var isStateValid: Bool
```
loudnate: These can be `let`

bharat:
ok - successfully built an archive, exported it for an Ad Hoc build and deployed it to my iPhone 6+. The app seems to be working fine - but the extension seems to be working off of old data and not updating. I don't see a crash in the crash logs - and the Loop Status Extension doesn't show up in the process list, even though I'm attached to my device and I'm looking at the widget. Any thoughts about what's going on here?

Loop Status Extension/StatusExtensionContext.swift
bharat: fixed.

bharat:
Hey team. I'm struggling to debug why this works fine for a Debug build but not a production build. I think at this point it makes more sense to merge it to dev and get more eyes on the problem. This would also free me up to do the follow-on refactors (ie, moving the shared views into their own framework, etc). What do you think?

Issue #267, opened 2016-11-20
Displaying decimals on Watch for mmol 


alternateal:
Neither the watch app nor the complication display decimal places. For users of mmol/L a whole number doesn't make sense. Is there any way of adding this depending on health kit units?

alternateal:
#338 should close this

PR #303, opened 2016-12-03
Issue #267 - Displaying decimals on Watch for mmol

alternateal:
Updated code to use significant digits to enable decimals to be shown on both watch app and complication.


WatchApp Extension/ComplicationController.swift
```diff
@@ -40,10 +40,13 @@ final class ComplicationController: NSObject, CLKComplicationDataSource {
     
     // MARK: - Timeline Population
 
-    private lazy var formatter = NumberFormatter()
+ //   private lazy var formatter = NumberFormatter()
 
     func getCurrentTimelineEntry(for complication: CLKComplication, withHandler handler: (@escaping (CLKComplicationTimelineEntry?) -> Void)) {
 
+        let formatter = NumberFormatter(); formatter.usesSignificantDigits = true; formatter.minimumSignificantDigits = 2; formatter.maximumSignificantDigits = 3
```
ps2: What does this display with mg/dl?

Need to avoid repeating this code three times.

ps2:
Actually, I think we will be able to use @bharat 's extension to do this: https://github.com/LoopKit/Loop/pull/281/commits/b41ceda7b6f7a3c90261e62a854f273297da1419

alternateal:
Happy to use @bharat's extension (provided it works on watch also) - I knew it was quick and dirty solution - just needed to make something workable for users of mmol/L. Happy to close the thread if you want?

ps2:
I would rather you keep it open and then update it to use the HKUnit unit extension when it's available!

loudnate:
Once the HKUnit extension is available, include it in the watch target along with `NSNumberFormatter.swift`, so you can use `NumberFormatter.glucoseNumberFormatter(for: HKUnit)`

ps2:
The HKUnit extension has been merged into dev, if you want to rebase and update this PR to use it.

ps2:
I believe this is resolved properly with #379

PR #310, opened 2016-12-09
If minBG is below 55, then shut off basal, even if eventualBG is in r…

### LINKED ISSUE #307: Safety Feature - Suspend on low BG regardless of projections
We ran into a potentially dangerous situation where bg went low but the projections predicted him to go high so it stopped low temping him.  I think that there should be in the algo, a limit that if the bg falls below, it will suspend the basal regardless of projections.  This could be because the carb count is incorrect, iob is incorrect, setting could be incorrect, or any number of different issues.   For oref0, we put a hard limit in place at 55.  Something similar would be good to implement in Loop.

You can see the discussion around this for oref0 at https://github.com/openaps/oref0/issues/84

Attached is the chart showing the scenario of what happened with Loop tonight

image

ps2:
If minBG is below 55 mg/dL, then recommendation should be to shut off basal, even if eventualBG is in range. Simplify code by removing allowPredictiveTempBelowRange flag.

Fixes #307


amazaheri:
👍

bharat:
good change, thanks Pete!

ps2:
This is now configurable: #365

PR #316, opened 2016-12-11
Properly support mmol/L in the status extension.

### LINKED ISSUE #311: Crash in dev when using mmol/L
When mmol/L units are selected as the default unit for blood glucose measurements in HealthKit, loop crashes on startup.

bharat:
Pass the raw HKUnit.unitString (eg. mmol<180.1558800000541>/L) through
the context, as opposed to the unitDisplayString which is uses the
glucose-centric version (eg. mmol/L). This way we can properly hydrate
the unit on the extension side and represent it in the extension.

Fixes #311.

### DIFF
--- a/Loop Status Extension/StatusExtensionContext.swift
+++ b/Loop Status Extension/StatusExtensionContext.swift
@@ -117,8 +117,8 @@ final class StatusExtensionContext: NSObject, RawRepresentable {
 
         raw["preferredUnitString"] = preferredUnitString
         
-        if let glucose = latestGlucose,
-           preferredUnitString != nil {
+        if preferredUnitString != nil,
+            let glucose = latestGlucose {
             raw["latestGlucose_value"] = glucose.quantity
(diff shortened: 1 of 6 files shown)


Loop Status Extension/StatusExtensionContext.swift
```diff
@@ -115,10 +115,10 @@ final class StatusExtensionContext: NSObject, RawRepresentable {
             "version": version
         ]
 
-        raw["preferredUnitDisplayString"] = preferredUnitDisplayString
+        raw["preferredUnitString"] = preferredUnitString
         
         if let glucose = latestGlucose,
-           preferredUnitDisplayString != nil {
+           preferredUnitString != nil {
```
loudnate: What is this check for?

loudnate: OK - reversing the order of the conditions would make that clearer.

bharat: good point. reversed the order.

PR #334, opened 2016-12-23
Resolve #319: tapping the widget opens the app

### LINKED ISSUE #319: Status widget - open app on tap
When you tap the widget, it should open the app.

bharat:
Register the "Loop" app url scheme in order to make the app an addressable target.

### DIFF
--- a/Loop/Extensions/NSBundle.swift
+++ b/Loop/Extensions/NSBundle.swift
@@ -32,7 +32,7 @@ extension Bundle {
 
     var mainAppUrl: URL? {
         if let mainAppBundleIdentifier = mainAppBundleIdentifier {
-            return URL(string: mainAppBundleIdentifier + "://")
+            return URL(string: "\(mainAppBundleIdentifier)://")
         } else {
             return nil
         }
--- a/Loop/Extensions/NSBundle.swift
+++ b/Loop/Extensions/NSBundle.swift
@@ -29,4 +29,12 @@ extension Bundle {
     var appGroupSuiteName: String {
         return "group.\(mainAppBundleIdentifier!)"
     }
+
+    var mainAppUrl: URL? {
+        if let mainAppBundleIdentifier = mainAppBundleIdentifier {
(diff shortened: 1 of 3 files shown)


bharat:
this resolves #319 - adding this comment to link the two.

Loop/Info.plist
```diff
@@ -2,6 +2,15 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>CFBundleURLTypes</key>
+	<array>
+		<dict>
+			<key>CFBundleURLSchemes</key>
+			<array>
+				<string>Loop</string>
```
loudnate: Lowercase would be preferred. Any idea if this will create conflicts with another app? 

Alternatively you could derive something unique using the bundle identifier, which will be nice for the few folks who have multiple installations.

bharat: I did some searching and couldn't find a canonical repository, and since "loop" is 4 characters I'm positive there's a conflict out there somewhere. I just pushed a change to use MAIN_APP_BUNDLE_IDENTIFIER, so the URL now looks like: "com.loopkit.Loop://".

It works in the simulator. Haven't tested it on my device yet.

Loop/Extensions/NSBundle.swift
```diff
@@ -29,4 +29,12 @@ extension Bundle {
     var appGroupSuiteName: String {
         return "group.\(mainAppBundleIdentifier!)"
     }
+
+    var mainAppUrl: URL? {
+        if let mainAppBundleIdentifier = mainAppBundleIdentifier {
+            return URL(string: mainAppBundleIdentifier + "://")
```
loudnate: `"\(mainAppBundleIdentifier)://"`

bharat: good catch. fixed.

Issue #320, opened 2016-12-13
Improve the Status Widget to update in real time as Loop data changes


bharat:
Currently, the status widget only updates when it first appears. It should continue to update in real time as the Loop app detects changes and makes adjustments.

bharat:
Looks like this might be non-trivial:
1. http://stackoverflow.com/questions/28736506/kvo-with-shared-nsuserdefaults-in-swift
1. http://www.atomicbird.com/blog/sharing-with-app-extensions

MMWormhole might work for us:
https://github.com/mutualmobile/MMWormhole

Pulling in a new framework for this seems heavy. Alternatively, it might just be easiest to have the extension poll every minute. It's not like the data changes that quickly.

ps2:
Could you update the description to describe the issue being solved?

bharat:
done. lmk if it needs more clarity.

ps2:
So this is only an issue if you keep the status screen up for a while.  Each time you bring it up it's is still refreshing correctly?

bharat:
That is correct. So IMO it's not a critical issue, but if you happen to stay on the status screen for a while you might be looking at expired data. For what it's worth, it also shows the timestamps for all readings so you'd at least have a visual way of knowing that it's out of date, but date stamps are subtle.

marionbarker:
I can't make the widget stay up long enough to test this.  It does refresh each time I call it up.
Can this issue be closed?

PR #335, opened 2016-12-24
Monitor changes and update the widget in real time.

bharat:
Use KVO to watch the UserDefaults object and update the view as
appropriate. This is a first pass at resolving #320.


Loop Status Extension/StatusViewController.swift
```diff
@@ -20,33 +20,65 @@ class StatusViewController: UIViewController, NCWidgetProviding {
...
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+
+        defaults = UserDefaults(suiteName: Bundle.main.appGroupSuiteName)
+        if let defaults = defaults {
+            defaults.addObserver(
+                self,
+                forKeyPath: defaults.statusExtensionContextObservableKey,
+                options: [.new, .initial],
+                context: nil)
```
loudnate: It's important to use a context here. Example:

https://developer.apple.com/library/content/documentation/Swift/Conceptual/BuildingCocoaApps/AdoptingCocoaDesignPatterns.html#//apple_ref/doc/uid/TP40014216-CH7-ID6

Loop Status Extension/StatusViewController.swift
```diff
@@ -20,33 +20,65 @@ class StatusViewController: UIViewController, NCWidgetProviding {
...
     
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+
+        defaults = UserDefaults(suiteName: Bundle.main.appGroupSuiteName)
+        if let defaults = defaults {
+            defaults.addObserver(
+                self,
+                forKeyPath: defaults.statusExtensionContextObservableKey,
+                options: [.new, .initial],
```
loudnate: If you aren't using the change dict in the observer than you should leave these blank.

Loop Status Extension/StatusViewController.swift
```diff
@@ -20,33 +20,65 @@ class StatusViewController: UIViewController, NCWidgetProviding {
...
+    override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+
+        if let defaults = defaults {
+            defaults.removeObserver(self, forKeyPath: defaults.statusExtensionContextObservableKey)
+        }
+    }
+    
+    override func observeValue(forKeyPath keyPath: String?, of object: Any?, change: [NSKeyValueChangeKey : Any]?, context: UnsafeMutableRawPointer?) {
+        update()
```
loudnate: Always check the context. If it doesn't match, then you must call super.

Loop Status Extension/StatusViewController.swift
```diff
@@ -20,33 +20,65 @@ class StatusViewController: UIViewController, NCWidgetProviding {
...
+    var context: StatusExtensionContext?
+    var defaults: UserDefaults?
+    
     override func viewDidLoad() {
         super.viewDidLoad()
         subtitleLabel.alpha = 0
         subtitleLabel.textColor = UIColor.secondaryLabelColor
     }
     
+    override func viewWillAppear(_ animated: Bool) {
```
loudnate: this can be called multiple times, and disappear can as well. These two aren't always balanced in the case of an interactive transition.

bharat: ok, that's good to know. I'll move these to viewDidLoad() and deinit, following the pattern in PredictionTableViewController

Loop Status Extension/StatusViewController.swift
bharat: Will do. The purpose of the context member was pretty opaque to me. I'm assuming it's some kind of semaphore that lets you know that the notification is relevant to this particular instance.

Loop Status Extension/StatusViewController.swift
bharat: can you elaborate on that? Not sure what you mean by `using the change dict in the observer`.

Loop Status Extension/StatusViewController.swift
bharat: Is it legit to use the `context` variable (defined as `var context: StatusExtensionContext?`) for this purpose? It's convenient. If not, I'll create a new var.

bharat:
Ok - I think I covered all the issues you raised. Thanks for the guidance!

I'm not sure how to properly test this in the simulator, because when I go to the lock screen, the Loop app isn't updating so I don't see any changes. Any thoughts on how to test it in the simulator? If not, I'll deploy it to my phone and just observe it for 5-10 minutes :-)

Loop Status Extension/StatusViewController.swift
loudnate: I wouldn't, foremost because it's convoluting the term "context" for future readers, but secondly because you're looking for a unique pointer value here, and it should feel safer to instead use one containing a simple integer value that never changes.

Loop Status Extension/StatusViewController.swift
loudnate: passing the `.new` key here is so it is included in the `observeValue` call below. `.initial` means the observer is called right away, which isn't what you want.

Loop Status Extension/StatusViewController.swift
```diff
@@ -20,33 +20,64 @@ class StatusViewController: UIViewController, NCWidgetProviding {
...
+    }
+    
+    deinit {
+        if let defaults = defaults {
+            defaults.removeObserver(self, forKeyPath: defaults.statusExtensionContextObservableKey)
+        }
+    }
+    
+    override func observeValue(forKeyPath keyPath: String?, of object: Any?, change: [NSKeyValueChangeKey : Any]?, context: UnsafeMutableRawPointer?) {
+        if context != &self.context {
```
bharat: good point. fixed!

PR #338, opened 2016-12-25
add more complication families and improve the display (adding trends)

ps2:


walker0:
Looks great! Thanks for working this through more thoroughly. I personally quite like the Modular Large complication and relative time instead of absolute. 
img_7453

loudnate:
@waker0 I think that looks great, though you could provide more info like IOB in a table there too. Would be great if you could submit follow-up PRs after this is merged.

Issue #325, opened 2016-12-13
Battery level missing from widget


alternateal:
1.2.0 looks to have some consistency issues  between the app and the widget - see attached photos 
![img_3137](https://cloud.githubusercontent.com/assets/17626700/21157835/cfbe6db0-c172-11e6-898b-c66b4b281263.PNG)
![img_3138](https://cloud.githubusercontent.com/assets/17626700/21157837/cff4173a-c172-11e6-8d5e-60d80af93d79.PNG)
![img_3139](https://cloud.githubusercontent.com/assets/17626700/21157836/cff4340e-c172-11e6-9e78-bd987aab9c45.PNG)

Battery needs to show on widget and app needs to show same precision as the widget for the "eventually" mmol/L levels.

bharat:
@alternateal: The "eventually" text is covered in #324. Can you rename this issue to something along the lines of "Battery level missing from widget" please?

wkcwells:
Pull request #342 fixes this for x22 pumps, and has been merged.  If this report is specific to x22 pumps, this issue can be closed.

alternateal:
Closed by pull request #342

loudnate:
Please leave issues open until they are merged to master (and subsequently released to users)

wkcwells:
Nate, thanks for keeping us on process.

PR #342, opened 2016-12-27
Fix widget battery level for x22 pumps

wkcwells:
This change fixes the lock screen widget battery level to work with MiniMed x22 pumps, which do not send MySentry packets.  I believe that this covers issue #325.  The change just makes use of the work done for issue #141 - thanks to the team for that great work!  (Note that I have not been able to test this for regression against x23 model pumps as I do not have one.)


shanselman:
THANK YOU ;)

Issue #100, opened 2016-08-04
Ability to use MM CGM with an x22 Pump


tomasboudr:
Looking for the ability to use a MM CGM with an x22 pump via loop.

loudnate:
Please file an issue in rileylink_ios as well, as the order of work is to parse sensor data pages. 

Are you willing to lend a hand to building this support out?

tomasboudr:
Hi Nate,
Yes of course. I will create an issue under rileylink_ios as well. I'm a
complete amateur at iOS development but have started the Standford class on
iTunes U and wanted to get into it for a while now, so this is a great
excuse.

tmecklem:
@tomasboudr it's been a while since you opened this issue. Are you still interested in the feature?

tomasboudr:
I got a G5, but I still think this is a useful feature and there are some folks working on it. So this should remain open until they PR those in.

oskarpearson:
This can be closed; functionality now exists since https://github.com/LoopKit/Loop/pull/378

Kdisimone:
closing issue, thanks @oskarpearson

PR #349, opened 2016-12-30
X22 enlite cgm support

tmecklem:
This PR adds initial Loop support for enlite cgm on pumps that do not support MySentry. https://github.com/LoopKit/Loop/issues/100

Overview:

1. Uses RileyLink MinimedKit support for fetching glucose history pages. Glucose fetching happens if needed after completion of assertCurrentPumpStatus.
2. Adds switch to activate pump cgm fetching
3. Adds sensor info to show "!" when sensor calibrations are required or a sensor error or weak signal happens

PR changes requested:
- [x] Rename functions to enlite for clarity
- [x] Group related enlite data and functions together
- [x] Remove tight coupling between pump status and enlite data fetching

### DIFF
--- a/Loop/Extensions/NSUserDefaults.swift
+++ b/Loop/Extensions/NSUserDefaults.swift
@@ -25,6 +25,7 @@ extension UserDefaults {
         case MaximumBasalRatePerHour = "com.loudnate.Naterade.MaximumBasalRatePerHour"
         case MaximumBolus = "com.loudnate.Naterade.MaximumBolus"
         case PreferredInsulinDataSource = "com.loudnate.Loop.PreferredInsulinDataSource"
+        case PumpFetchGlucoseEnabled = "com.loopkit.Loop.FetchPumpGlucoseEnabled"
         case PumpID = "com.loudnate.Naterade.PumpID"
         case PumpModelNumber = "com.loudnate.Naterade.PumpModelNumber"
         case PumpRegion = "com.loopkit.Loop.PumpRegion"
@@ -211,6 +212,15 @@ extension UserDefaults {
         }
         set {
             set(newValue, forKey: Key.G4ReceiverEnabled.rawValue)
+        }
+    }
+
+    var fetchPumpGlucoseEnabled: Bool {
+        get {
+            return bool(forKey: Key.PumpFetchGlucoseEnabled.rawValue)
+        }
+        set {
+            set(newValue, forKey: Key.PumpFetchGlucoseEnabled.rawValue)
         }
     }
 
--- a/Loop/Extensions/NSUserDefaults.swift
+++ b/Loop/Extensions/NSUserDefaults.swift
@@ -25,7 +25,7 @@ extension UserDefaults {
         case MaximumBasalRatePerHour = "com.loudnate.Naterade.MaximumBasalRatePerHour"
         case MaximumBolus = "com.loudnate.Naterade.MaximumBolus"
         case PreferredInsulinDataSource = "com.loudnate.Loop.PreferredInsulinDataSource"
-        case PumpFetchGlucoseEnabled = "com.loopkit.Loop.FetchPumpGlucoseEnabled"
+        case FetchEnliteDataEnabled = "com.loopkit.Loop.FetchEnliteDataEnabled"
         case PumpID = "com.loudnate.Naterade.PumpID"
         case PumpModelNumber = "com.loudnate.Naterade.PumpModelNumber"
         case PumpRegion = "com.loopkit.Loop.PumpRegion"
@@ -215,12 +215,12 @@ extension UserDefaults {
         }
     }
 
-    var fetchPumpGlucoseEnabled: Bool {
+    var fetchEnliteDataEnabled: Bool {
         get {
-            return bool(forKey: Key.PumpFetchGlucoseEnabled.rawValue)
+            return bool(forKey: Key.FetchEnliteDataEnabled.rawValue)
         }
         set {
-            set(newValue, forKey: Key.PumpFetchGlucoseEnabled.rawValue)
+            set(newValue, forKey: Key.FetchEnliteDataEnabled.rawValue)
         }
     }
 
--- a/Loop/Managers/DeviceDataManager.swift
+++ b/Loop/Managers/DeviceDataManager.swift
@@ -54,6 +54,15 @@ final class DeviceDataManager: CarbStoreDelegate, CarbStoreSyncDelegate, DoseSto
         }
     }
 
+    var fetchPumpGlucoseEnabled: Bool {
+        get {
+            return UserDefaults.standard.fetchPumpGlucoseEnabled
+        }
+        set {
+            UserDefaults.standard.fetchPumpGlucoseEnabled = newValue
+        }
+    }
+
     var sensorInfo: SensorDisplayable? {
         return latestGlucoseG5 ?? latestGlucoseG4 ?? latestGlucoseFromShare ?? latestPumpStatusFromMySentry
     }
@@ -367,7 +376,9 @@ final class DeviceDataManager: CarbStoreDelegate, CarbStoreSyncDelegate, DoseSto
         device.assertIdleListening()
 
         assertCurrentPumpStatus(device: device) {
-            self.assertCurrentPumpCGMData(device: device);
+            if UserDefaults.standard.fetchPumpGlucoseEnabled {
+                self.assertCurrentPumpCGMData(device: device)
+            }
         }
     }
 
--- a/Loop/Managers/DeviceDataManager.swift
+++ b/Loop/Managers/DeviceDataManager.swift
@@ -64,7 +64,7 @@ final class DeviceDataManager: CarbStoreDelegate, CarbStoreSyncDelegate, DoseSto
     }
 
     var sensorInfo: SensorDisplayable? {
-        return latestGlucoseG5 ?? latestGlucoseG4 ?? latestGlucoseFromShare ?? latestPumpStatusFromMySentry ?? latestGlucoseFromPumpHistory
+        return latestGlucoseG5 ?? latestGlucoseG4 ?? latestGlucoseFromShare ?? latestPumpStatusFromMySentry ?? latestEnliteData
     }
 
     var latestPumpStatus: RileyLinkKit.PumpStatus?
@@ -176,8 +176,6 @@ final class DeviceDataManager: CarbStoreDelegate, CarbStoreSyncDelegate, DoseSto
     // MARK: Pump data
 
     var latestPumpStatusFromMySentry: MySentryPumpStatusMessageBody?
-
-    fileprivate var latestGlucoseFromPumpHistory: PumpGlucoseHistorySensorDisplayable?
 
     /**
      Handles receiving a MySentry status message, which are only posted by MM x23 pumps.
@@ -379,7 +377,7 @@ final class DeviceDataManager: CarbStoreDelegate, CarbStoreSyncDelegate, DoseSto
 
         assertCurrentPumpStatus(device: device) {
             if UserDefaults.standard.fetchPumpGlucoseEnabled {
-                self.assertCurrentPumpCGMData(device: device)
+                self.assertCurrentEnliteData(device: device)
             }
         }
     }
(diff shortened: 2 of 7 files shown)


Loop/View Controllers/SettingsTableViewController.swift
```diff
@@ -189,6 +190,17 @@ final class SettingsTableViewController: UITableViewController, DailyValueSchedu
                 return switchCell
             }
 
+            if case .fetchPumpGlucose = ConfigurationRow(rawValue: indexPath.row)! {
+                let switchCell = tableView.dequeueReusableCell(withIdentifier: SwitchTableViewCell.className, for: indexPath) as! SwitchTableViewCell
+
+                switchCell.`switch`?.isOn = dataManager.fetchPumpGlucoseEnabled
+                switchCell.titleLabel.text = NSLocalizedString("Fetch Pump Glucose", comment: "The title text for the fetch pump glocose enabled switch cell")
```
tmecklem: I'm not a fan of the switch, but I don't have a better solution to enable/disable glucose history fetching by deduction yet. 

Is the phrase "Fetch Pump Glucose" clear enough?

tmecklem:
This is looking very solid now in testing. There are no more crashes after the changes ensuring only unique entries are inserted in GlucoseStore, and the "!" display shows correctly only when there's a sensor error or warning.

cc @ps2

bharat:
Thanks for the hard work, Tim! I'm going to review this in two
passes. The first pass for structure and design, then the second pass
I'll dig into APIs, style, etc. This way we don't wind up polishing
until there's agreement on direction.

First, from reading the code (and I'm new to Loop so please bear with me!)
here's what I understand of the old and new logic.

The old logic for assertCurrentPumpData()
1. Check to see if we need pump status
2. Read pump data (via RileyLink)
3. Update pump data in relevant places (reservoir, battery, etc)

The new logic for assertCurrentPumpData():
1. Check to see if we need pump status
2. Call out to assertCurrentPumpStatus()
  3. Read pump data (via RileyLink)
  4. Calls (optionally, via closure) to assertCurrentPumpCGMData, which
    5. fetches sensor values from the pump and..
    6. pushes those values into latestGlucoseFromPumpHistory and..
    7. pushes those values up to NightScout and..
    8  adds those values to the GlucoseStore

If this is a correct reading, then there are two significant
structural issues that we should work through.

The first issue is that it's unclear what happens in the case when
Enlite is enabled in addition to any other existing method (G5 or
G4ShareSpy). In that case from my reading, we would wind up
storing both data sources into the glucose store, which seems like
a pretty bad idea.

I think the easiest fix for this would be to change the settings
code to disallow letting any two of those settings to be enabled.
Currently I think it's the case that you can enable both G4 and G5
and (if you slap both of those sensors on your body) you'd get a
conflict - but it's such an unlikely case that it probably hasn't
arisen yet. If you agree, we could make this a separate refactor that
I (or somebody else) does ahead of this change going in so that we
don't slow you down.

The second issue is that from an abstraction perspective, it's
non-intuitive to piggyback loading the Enlite data at the same time
as we're getting the general pump data. To clarify this, I think the
right approach is to fully out these two pieces of functionality:

1. Create a new section in DeviceDataManager for Enlite and mark it
   with a "// MARK: Enlite" marker 
2. Move your assertCurrentPumpCGMData() method which does the bulk of
   the work, into that new section and rename it something like
   fetchEnliteData() to indicate that it's actively reaching out to
   get the data.
3. Have fetchEnliteData() get its own RileyLink from the
   rileyLinkManager directly, so that you don't have to take an
   argument here.
4. return assertCurrentPumpData to its original state()

Now that you've separated them out, the next step is to find the right
place to call it from. Assuming that we've adjusted the UI so that
it's not possible to fetch Enlite data at the same time as using G4 or
G5, this means that the only heartbeat that is actually useful here is
the one that happens in receivedRileyLinkTimerTickNotification(), so 
inside the closure there you can conditionally call fetchEnliteData()
if it's enabled.

Note also that I am implicitly renaming this feature to align with the
product name ("Enlite") instead of the more generic "pump
glucose". This is more consistent with our other naming so far (G4,
G5, etc). 

Hope that's helpful! I'm going to pull out a feature request for the settings
work and see if I can get a PR up so that your code can slot in.

tmecklem:
@bharat thanks for having a look at the PR. The Loop ecosystem is my first excursion into Swift for iOS development, so I appreciate your patience as we work through this. I've separated out the following by first and second issue as you described them in your review.

####First Issue####

I appreciate you offering to take the multiple CGM scenario as a separate issue. I see that as an important, independent and somewhat larger concern than the work in this PR. I had some user confusion in setting up CGM when I first installed Loop, so I'd love to see a unified experience in choosing and enabling only a single CGM option at a time. I also have a concern that people will not have enough information to make an informed decision about the difference between receiving glucose data via MySentry vs fetching Enlite data from glucose history. I think there's room for iterative improvement in the wording and guidance Loop can give users. It would be too much to try to solve it all here.

####Second Issue####

I agree that the chaining of the pump history to cgm history fetching couples the two together in an unfortunate way. I'll implement your four numbered suggestions for this issue. 

Here's some background for context. My primary goals at the time I wrote the code were:

1. Fetch pump history and glucose history as closely together in time as possible, to have contemporary data for Loop's algorithm and use pump comm sessions efficiently
2. Avoid contention for the scarce pump resource that I believed might happen by triggering both fetches independently from the same timer event.

As I understand now from recent Gitter conversations, direct contention isn't a concern since there is a queueing mechanism for pump messages. With that new information in mind, I feel comfortable with what you've proposed.

####An aside####
From a subjective perspective, it feels like adding an Enlite MARK is a cosmetic change to a deeper problem. I'm very uncomfortable with the DeviceDataManager class. It has outgrown a single responsibility and it's difficult to maneuver in that class and make isolated and testable changes. Each of the MARKs looks like they could be a separate class with a singular focus, and I don't feel good about adding another responsibility with the addition of this cgm fetch. Making the larger change I'm suggesting is beyond the time and energy I can devote to the project right now though.

You've doing a good job in your analysis, and the depth of your review impresses me 👍. Keep up the good work!

Loop/View Controllers/SettingsTableViewController.swift
ps2: Similar to what @bharat suggested, 'Fetch Enlite Data' would probably be better here.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -391,10 +410,58 @@ final class DeviceDataManager: CarbStoreDelegate, CarbStoreSyncDelegate, DoseSto
                     nsPumpStatus = nil
                 }
                 self.nightscoutDataManager.uploadDeviceStatus(nsPumpStatus)
+                completion?()
             }
         }
     }
 
+    private func updatePumpCGMSensorStatus(_ events: [TimestampedGlucoseEvent]) -> Void {
```
ps2: Agree about the `// MARK: Enlite` being a cosmetic change, but it would be nice here. This file is the high level thin layer of state above all the device data that we manage; as long as the code here is more about how that data fits/integrates into Loop, and not just managing device details, then I think it's ok.

ps2: And wrt function and variable names, I agree with renaming 'pump cgm data' to 'enlite data'

ps2: I think @bharat has some good ideas about further improvements, and is doing the right thing by opening separate issues to track those. I think we can get this patch merged in with just the naming and mark changes.

ps2:
The way Tim has it currently (calling assertCurrentPumpCGMData from the assertCurrentPumpStatus completion handler) has the benefit of only triggering an Enlite data fetch once every 5 minutes, and doing it while the pump is already awake.  The receivedRileyLinkTimerTickNotification is ticked every 1 minute, so we don't want to trigger the fetch each time we get that signal.

bharat:
> The way Tim has it currently (calling assertCurrentPumpCGMData from the assertCurrentPumpStatus completion handler) has the benefit of only triggering an Enlite data fetch once every 5 minutes, and doing it while the pump is already awake. The receivedRileyLinkTimerTickNotification is ticked every 1 minute, so we don't want to trigger the fetch each time we get that signal. 

@ps2 thanks for weighing in. I hadn't considered that issue.

It looks like `assertCurrentPumpData` wakes up the device (with `device.assertIdleListening`) then decides whether to poll for pump status. This means that it's waking up the device on every RL heartbeat, right?

We also know that it's a requirement that `assertCurrentPumpData` is the only trigger for a loop.

Combining that info, I'm thinking that the right approach is to have `fetchEnliteData` have some similar logic to wake up the pump and fetch new data if the current data is stale, then call `assertCurrentPumpData` afterwards.  Maybe the right API there is to have an `assertCurrentEnliteData` wrapper around `fetchEnliteData`, and call that before calling `assertCurrentPumpData`?

ps2:
> It looks like assertCurrentPumpData wakes up the device (with device.assertIdleListening) then decides whether to poll for pump status. This means that it's waking up the device on every RL heartbeat, right?

No, that just makes sure that we're running a 'listen' command on the RL, if idleListeningEnabled. 

Running a loop is ultimately triggered by updateReservoirVolume, so we should fetch enlite data before that.

Loop/Extensions/NSUserDefaults.swift
```diff
@@ -214,6 +215,15 @@ extension UserDefaults {
         }
     }
 
+    var fetchPumpGlucoseEnabled: Bool {
+        get {
```
ps2: fetchEnliteDataEnabled

Loop/Extensions/NSUserDefaults.swift
```diff
@@ -25,6 +25,7 @@ extension UserDefaults {
         case MaximumBasalRatePerHour = "com.loudnate.Naterade.MaximumBasalRatePerHour"
         case MaximumBolus = "com.loudnate.Naterade.MaximumBolus"
         case PreferredInsulinDataSource = "com.loudnate.Loop.PreferredInsulinDataSource"
+        case PumpFetchGlucoseEnabled = "com.loopkit.Loop.FetchPumpGlucoseEnabled"
```
ps2: FetchEnliteDataEnabled

Loop/Managers/DeviceDataManager.swift
```diff
@@ -366,6 +375,14 @@ final class DeviceDataManager: CarbStoreDelegate, CarbStoreSyncDelegate, DoseSto
 
         device.assertIdleListening()
 
+        assertCurrentPumpStatus(device: device) {
+            if UserDefaults.standard.fetchPumpGlucoseEnabled {
+                self.assertCurrentEnliteData(device: device)
```
ps2: This needs to be synchronized with fetching pump data.  As it stands, it looks like pump data and enlite data could both get onto their own independent 5 min cycles.

Loop/Extensions/NSUserDefaults.swift
ps2: Also in DeviceDataManager and SettingsTableViewController

bharat:
> >It looks like assertCurrentPumpData wakes up the device (with device.assertIdleListening) then decides whether to poll for pump status. This means that it's waking up the device on every RL heartbeat, right?
>No, that just makes sure that we're running a 'listen' command on the RL, if idleListeningEnabled.
>
>Running a loop is ultimately triggered by updateReservoirVolume, so we should fetch enlite data before that.

Thanks for the explanation! So then as I understand it, our constraints are:
1. We can only count on a RileyLink heartbeat event, so our trigger has to go there
1. We want to make sure that we don't wake up the pump twice
1. We want to load the Enlite glucose values before we load the pump state, since loading the pump state will call out to `updateReservoirValue` and trigger a loop

Do those constraints sound about right? Assuming that they are for now, one possible solution is to abstract the current trigger logic out of `assertCurrentPumpData`:

```
        // How long should we wait before we poll for new pump data?
        let pumpStatusAgeTolerance = rileyLinkManager.idleListeningEnabled ? TimeInterval(minutes: 11) : TimeInterval(minutes: 4)

        // If we don't yet have pump status, or it's old, poll for it.
        if  doseStore.lastReservoirValue == nil ||
            doseStore.lastReservoirValue!.startDate.timeIntervalSinceNow <= -pumpStatusAgeTolerance ```

...into a new function. Let's call it `pumpDataIsStale`. Then we can call it from both `assertCurrentPumpData` as well as `assertCurrentEnliteData` so that we have the same trigger condition for both. Then we just need to make sure that we call `assertCurrentEnliteData` before calling `assertCurrentPumpData` in the heartbeart:

```
    @objc private func receivedRileyLinkTimerTickNotification(_ note: Notification) {
        backfillGlucoseFromShareIfNeeded() {
            if UserDefaults.standard.fetchPumpGlucoseEnabled,
                self.assertCurrentEnliteData()
            }
            self.assertCurrentPumpData()
        }
    }
```

I believe this meets all constraints. We only call it from the RL heartbeat, we only wake up the pump when its data is stale, and we load Enlite before asserting current pump data.

thoughts?

bharat:
ok - I just submitted https://github.com/LoopKit/Loop/pull/364 which creates exclusive CGM data sources. once that's approved and pulled, you can merge it into your branch and add Enlite as another exclusive option.

ps2:
Re @bharat's ideas for specific implementation of assertCurrentEnliteData ordering: I think that sounds like it would work.

bharat:
hey @tmecklem - I've been traveling but just got back to town. Is this ready for another look? If not, just ping me in this pull request when you're ready and I'll dive in.

tmecklem:
Yes, just waiting on https://github.com/LoopKit/Loop/pull/364 as far as I know, although this could ship first and that PR changed to address the new toggle. 

This branch has continued to work in production for the T1D I'm supporting with the changes you and Pete have suggested so far. Thanks for the feedback to help address the naming and coupling concerns.

Unfortunately, due to other work and home priorities I don't have any more time to devote to this PR. I recommend the code be merged as is and improved iteratively through future PRs. If it needs more work before it can be merged, I'll be happy to give the branch (and credit) to anyone who has the time to push it through to production for a wider audience.

ps2:
Continuing work on this in #378

ps2:
Thanks @tmecklem! Merged this in to dev under #378!

PR #330, opened 2016-12-22
Refactor HUD views into LoopUI framework (#284)

bharat:
First approximation of this refactor. Happy to make whatever changes are necessary.

PR #352, opened 2016-12-31
Refactor files with membership in multiple targets into a Common hierarchy

bharat:
Note that this is stacked up behind #330 so the diff is going to be wonky until that is merged into dev. Basically all I'm doing is `git mv` on every file into a new hierarchy and then fixing up the location in Xcode. There are no code changes.


bharat:
ok - this is now ready for a look.

Issue #360, opened 2017-01-04
Status widget - tapping on widget does not open app except for tapping on mini icon


wkcwells:
Related to/regression of PR #334.

Something fishy seems to be going on here since this was addressed by the above PR, but never worked for me.  (And I can't figure out anything that I could have done to break it.)  It works if you tap on the small Loop icon in the upper left of the widget, but that seems to be using some automatic functionality of widgets.

This code in `Loop/Extensions/NSBundle.swift` is always returning `nil` for me:

```
    private var mainAppBundleIdentifier: String? {
        return object(forInfoDictionaryKey: "MainAppBundleIdentifier") as? String
    }
```

wkcwells:
I have a simple change that appears to fix this.  Will PR shortly.

wkcwells:
Diff is here:
https://github.com/LoopKit/Loop/compare/dev...wkcwells:make-widget-tappable?expand=1

DM'd @bharat to take a look given.

bharat:
That looks like the right fix to me. I know I committed that same change, but I suspect I committed it to the wrong branch (my deploy branch) and it didn't go into my PR. Good catch, @wkcwells!

Kdisimone:
Closed, fix was merged.

PR #363, opened 2017-01-05
Add entry to widget info.plist to make tapping on widget launch Loop.app

wkcwells:
Confirmed that this is the right change with @bharat


wkcwells:
Fixes issue #360

Issue #361, opened 2017-01-05
Don't allow multiple CGM data sources to be active simultaneously


bharat:
Currently it's possible for you to enable the G4 and the G5 data sources concurrently. That's probably ok because who's going to slap a G4 and a G5 on their body? But we've got a change coming to add Enlite support in #349 which will complicate things.

Let's create a UI that allow the user to choose one and only one data source.

bharat:
Ok, it looks like it should be relatively straightforward to do a refactoring of the settings to pull out a couple of new groups. One for the pump (since now the Pump ID and the new Pump battery type are a related set) and one for the CGM. It'd look like this:

![screen shot 2017-01-04 at 7 54 22 pm](https://cloud.githubusercontent.com/assets/54570/21668552/ca5beda6-d2b7-11e6-9f48-a791b33a4926.png)

We can then create rules for the CGM section such that:
1. G5 Transmitter cannot be enabled unless the G5 Transmitter ID is well formed
2. Only one of the CGM options can be enabled at any given time

This creates space for us to add future Enlite support.

Kdisimone:
Resolved with https://github.com/LoopKit/Loop/pull/364 (plus new CGM manager work in current dev changes this selection as well)

PR #364, opened 2017-01-07
Only allow one active glucose data source

bharat:
Separate CGM sources out into their own section and enforce zero or one
active sources at any given time. This requires a new boolean value to
track whether the G5 transmitter is enabled independently of the actual
transmitter ID.

While we're at it, create a separate config section for the pump to
group it with the pump battery type since those are not really
configuration values and go together.

### DIFF
--- a/Loop/View Controllers/SettingsTableViewController.swift
+++ b/Loop/View Controllers/SettingsTableViewController.swift
@@ -660,7 +660,7 @@ final class SettingsTableViewController: UITableViewController, DailyValueSchedu
         if dataManager.transmitterEnabled == false {
             dataManager.transmitterEnabled = true
             disableReceiver()
-            tableView.insertRows(at: [IndexPath(row: CgmRow.transmitterID.rawValue, section:Section.cgm.rawValue)], with: .bottom)
+            tableView.insertRows(at: [IndexPath(row: CgmRow.transmitterID.rawValue, section:Section.cgm.rawValue)], with: .top)
         }
     }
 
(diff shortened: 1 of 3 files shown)


bharat:
This resolves #361 and makes room for us to add Enlite in https://github.com/LoopKit/Loop/pull/349

bharat:
screen shot 2017-01-07 at 3 28 00 pm

Some notes:
1. Enabling G5 shows the transmitter ID row (it animates out)
1. Enabling either G4 or G5 disables the other one
1. Disabling G5 (either directly, or by enabling G4) hides the transmitter ID row
1. Moved pump id and battery chemistry into their own section
1. handle the upgrade case (now transmitter enabled is explicit, not implicitly dependent on having a well-formed transmitter id)

Loop/View Controllers/SettingsTableViewController.swift
```diff
@@ -93,20 +95,31 @@ final class SettingsTableViewController: UITableViewController, DailyValueSchedu
...
         case pumpID = 0
-        case transmitterID
-        case receiverEnabled
-        case glucoseTargetRange
+        case batteryChemistry
+
+        static let count = 2
+    }
+
+    fileprivate enum CgmRow: Int {
```
loudnate: `s/Cgm/CGM`

bharat:
pull request am i
quietly i languish here
save me, @ps2

ps2:
This looks good to me.  Are there parts of the guide that should be updated?

Kdisimone:
I can do this. Just updating Anna's app now. This went into dev branch though...so I will note that in guide and when it goes to master I'll update the guide for that too.

PR #365, opened 2017-01-08
Minimum bg guard

ps2:
This branch adds a new configuration parameter: Minimum BG Guard. Bolus and temp basal recommendations will now be allowed as long as no part of the forecasted BG drops below Minimum BG Guard, even if parts are below the glucose target range.

To give the user more awareness of when recommendations are being given in this new context, the bolus interface has been updated to show notices of bg below target range. 

The bolus interface has also been updated to show Active Insulin, Active Carbohydrates, Eventual BG, and Pending Insulin.  Pending insulin includes boluses that have been issued to the pump, but have not been confirmed in pump history, as well as temp basals that have started but have not finished.


ps2:
img_0128

danibjor:
Love it

amazaheri:
Awesome, going to give it a spin ...

jlucasvt:
Just started running this branch.  I have a question about the "Message" area at the top of the bolus entry.  Is the text area mainly the last status message?  I am trying to understand what I should expect.  on first running it obviously points to "Configuration Error" (right i had no bg guard set) and then some "Comm Errors" also very understandable as I am just getting running, and need to sync up with dex and pump.  Now even after running for a bit and loop is closed I still see the warning iconography with no text.  

img_6363

If there is no warning should this be hidden or is the purpose to warn "Hey you are about to deliver a bolus!".  Which if that's the purpose it absolutely makes sense to me.  Just first impressions.  I love the data Active Carbs and Active and Pending Insulin.  Truly will be great.    More thoughts to come.  Great work @ps2

ps2:
Yes, thanks Jeremy, that's something that I need to fix.  In the case of no errors, the warning symbol is still being shown, and it shouldn't be.

jlucasvt:
I've been using this for a few days now and the Information and recommendation data has made my decisions better with regards to Bolus Calculation as well as daily BG Average outcomes.   My only take away and it is not really an issue is that the text (font) for Active Carb and Active Insulin would be much better if it was the same size as "Recommended" and "Bolus" title text.   I know I'm being picky and I probably need to stop pretending I am young and get a (plus) model phone.  I can make this adjustment on my version so it's not a big deal.

elnjensen:
I just installed this branch to check it out - have been wanting to see active insulin on the bolus screen, so this should be very useful.  Let me know if there's anything particular you want me to look out for, or just to run it as usual and see if anything of note comes up.

elnjensen:
Just noticed one thing - not sure if it's a coincidence, but it seemed like it was taking a long time for Loop to turn green after installing this new version.  Did a full history fetch, etc., as usual, but wasn't getting closed loop.  Then I scrolled through settings, and noticed that there was the new minimum BG guard parameter, for which I didn't yet have a value.   So I set a value, and then shortly after got a closed loop.  Not sure if that was related, but thought I'd mention it.

elnjensen:
img_1984

Very small cosmetic issue - part of "minimum" is cut off on iPhone 5s screen.

ps2:
@elnjensen Thanks for testing! Yes, Loop will be in an invalid configuration state until you set a minimum bg.  I didn't want to default this; we'll add it to the docs for new users, and put it as part of the release notes for those upgrading.

Thanks for the feedback on the clipped text on iphone5s.  That's fixable, as is the warning icon being always present.

PR #377, opened 2017-02-12
Show the glucose chart in the widget

bharat:
### 3/21/2017
1. added new screenshots
2. working on new to-do items at the bottom of this description.
### 3/23/2017
1. added higher res screenshots

## Details
This is the basic framework for exposing the glucose history and the predicted glucose in the widget. This takes advantage of the widget display modes that lets you zoom open the widget and show more details.

compact form looks like normal, but note that there's a [show more] link in the top right:
screen shot 2017-03-23 at 7 39 56 am

expanded form looks like this:
screen shot 2017-03-23 at 7 40 06 am

Work completed:
- [x] show the target BG range
- [x] move the chart code into the LoopUI framework
- [x] update screenshot to show latest layout
- [x] tighten up prediction values to be start date, interval and array of doubles.
- [x] reduce historical glucose data to 30 minutes so that we have a consistent experience regardless of whether the last data update came in when the device was locked or unlocked
- [x] Don't store new property in StatusChartsManager extensions - use a protocol or delegate in the common class and inject appropriate object from view controllers in Loop and extension
- [x] only serialize the most recent SensorDisplayableContext since that's the only thing we show in the extension UI
- [x] don't hardcode chart end date in the widget, use the last prediction value -- **punt this for now**
- [x] get rid of eventual glucose and use the last prediction value


elnjensen:
Looks great!  What sets the amount of time displayed on the x axis? 

If the show more / show less is toggled once, will it stay that way next time you display?  Or will you have to open it each time?  The former seems preferable to me if it's possible. 

Also, I'll put in another plug for displaying IOB ("Active insulin") in the status widget if possible.  Separate issue, I know, but it would be useful to see it there.

ps2:
@bharat Let me know when you think this is ready for review.

bharat:
It's worth a review just to make sure I'm headed in the right direction - but I haven't had much time to dig into this recently so I haven't advanced the ball beyond the initial commit...

bharat:
ok. I've cleaned up most of the major issues and synced this up with the latest changes in dev. We're down to two extension unsafe dylib warnings, and I still need to work on the BG range code - but it's headed in the right direction. As far as I can tell. Feel free to lob in criticism!

elnjensen:
Following this with interest - let me know if you need any testing of anything.  Thanks for your work on this.

bharat:
ok - this is now functionally complete and doesn't generate any warnings.. yay! I need to do a pass over it just to make sure there isn't anything dumb in there and then it'll be ready for a real review.

bharat:
to do:
- [ ] update screenshot to show latest layout
- [ ] tighten up prediction values to be start date, interval and array of doubles.

bharat:
@ps2 - not sure why the Travis builds are failing. I can't repro the failure in my dev environment and I don't see clues in the Travis output. :-/

sarahspins:
I know it's been requested before, but any chance to add a tiny IOB chart as well?

bharat:
@sarahspins that's a good idea (and I know that others want it as well). would you please file a separate feature request for that? I'd like to do it separately since I'm trying to avoid growing the scope of this pull request.

sarahspins:
Will do :)

bharat:
@ps2 - I think this is good to go!

PR #379, opened 2017-02-16
Add mmol watchapp units support. Add modularLarge complication. Change complication time to relative time.

walker0:
-Adds support for decimals in the watchApp for mmol users. 

-Adds modularLarge complication. ( currently in modularLargeTallBody form with only BG but happy to change it to add more info the future) 

-Changes complication time to relative time for ease of interpretation when glancing at the watch.


Loop/gallery.ckcomplication/Base.lproj/ckcomplication.strings
```diff
@@ -1,16 +0,0 @@
-/*
```
ps2: Any change made to the complication code requires a manual regeneration of this bundle file, which can be done in the watch simulator.

WatchApp Extension/ComplicationController.swift
```diff
@@ -81,14 +81,19 @@ final class ComplicationController: NSObject, CLKComplicationDataSource {
         let template: CLKComplicationTemplate?
         
         let glucoseText = CLKSimpleTextProvider.localizableTextProvider(withStringsFileTextKey: "120↘︎", shortTextKey: "120")
-        let timeText = CLKTimeTextProvider(date: Date())
+        let timeText = CLKRelativeDateTextProvider(date: Date(), style: .natural, units: .minute)
```
ps2: Does this fit on the smaller complications?

Loop/gallery.ckcomplication/Base.lproj/ckcomplication.strings
ps2: https://developer.apple.com/library/content/releasenotes/WatchKit/AddingComplications/index.html#//apple_ref/doc/uid/TP40017258

alternateal:
Tested the changes in my Dev setup and it works well - Thanks Matt

WatchApp Extension/ComplicationController.swift
walker0: I believe it does. It is reactive to available space so it should fit. I haven't ever been able to get the utilitarian large complication to work to test that one, nor have I looked into why. This change definitely works for modular small as well as the large version though.

Loop/gallery.ckcomplication/Base.lproj/ckcomplication.strings
walker0: I'm having problems with it generating the correct files at the moment. I did regenerate them before I added this PR but it clearly didn't add in the other file. I will look into this now and resolve it.

ps2:
In testing, I am seeing that complications are sometimes blanking out.  And it seems like the utilitarian complications aren't filling in values properly: I just get strings like "Loop", or "@@ @".  These may be existing issues, but could you look into them? 

Regarding merging this PR, the one concern I have left is that the change from absolute time to relative time introduces more uncertainty about how recent the measurement actually is, since complications don't always update in a timely manner.

walker0:
@ps2 the @@@ on utilitarian complications are an existing issue. I'll see if I can figure them out. Do you know of any of the other complications affected? 
Not showing until loop was green is how it has always been, it probably would be nice to change this.

I'm unsure what you mean by this third comment. The relative time display counts even when it isn't updated. It will always show the delta between current time and the BG time

ps2:
Ok, yes, you're right about relative time. That sounds good then.

PR #393, opened 2017-03-06
Charts refactor

ps2:
- Refactoring charting code to remove unnecessary work and unnecessary caching
- userDefaults["ModifiedCarbEntries"] never cleared for non-nightscout users
- Fixing a black screen during launch animation
- Disabled Amplitude's attempted CLLocation usage

### DIFF
--- a/Common/Extensions/NumberFormatter.swift
+++ b/Common/Extensions/NumberFormatter.swift
@@ -20,7 +20,7 @@ extension NumberFormatter {
         return numberFormatter
     }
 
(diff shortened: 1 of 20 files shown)


ps2:
@bharat You should take a look at the chart changes in this PR, and let me know if you have any comments.

bharat:
thanks for the heads up. this looks like it's moving in the right direction for me to use StatusChartManager in Loop Status Extension without it dragging in LoopKit and friends. I haven't had a chance to review it in detail, and it looks like it's not quite there yet (because StatusChartManager still requires LoopKit) - but I think this will get us most of the way there.

PR #395, opened 2017-03-10
Add a feature to limit maximum Insulin On Board

trixing:
With this settings the Bolus and Basal recommendation will be limited to a certain amount of Insulin.  It does not reject a Bolus above this amount if manually entered in the Bolus view.  If an exceeding amount is detected, the Basal rate will be set to the default rate.

[first pull request on github, please be kind]


trixing:
In case it is not obvious, but this is factored out of my master branch at https://github.com/trixing/Loop - that's where the merge conflicts come from.

Loop/Extensions/NSUserDefaults.swift
```diff
@@ -24,6 +24,7 @@ extension UserDefaults {
         case GlucoseTargetRangeSchedule = "com.loudnate.Naterade.GlucoseTargetRangeSchedule"
         case MaximumBasalRatePerHour = "com.loudnate.Naterade.MaximumBasalRatePerHour"
         case MaximumBolus = "com.loudnate.Naterade.MaximumBolus"
+        case MaximumIOB = "com.loudnate.Naterade.MaximumIOB"
```
ps2: New defaults keys should have the com.loopkit prefix.

DoseMathTests/DoseMathTests.swift
```diff
@@ -632,17 +712,38 @@ class RecommendBolusTests: XCTestCase {
...
+                                                               maxBolus: maxBolus,
+                                                               maxIOB: maxIOB,
+                                                               glucoseTargetRange: glucoseTargetRange,
+                                                               insulinSensitivity: self.insulinSensitivitySchedule,
+                                                               basalRateSchedule: basalRateSchedule,
+                                                               insulinOnBoard: 20.0
+        )
+        
+        XCTAssertEqual(0.0, dose)
+    }
```
ps2: There should be tests around the case when maxBolus is nil

ps2: And also for when a low temp is needed but IOB is above max

ps2: It should validate that it still low-temps correctly.

Loop/View Controllers/TextFieldTableViewController.swift
```diff
@@ -85,4 +85,18 @@ extension TextFieldTableViewController {
 
         return vc
     }
+    
+    static func maxIOB(_ value: Double?) -> T {
```
ps2: The setting VC should have explanatory text to explain how the feature works and that it can be left blank to disable the feature.

DoseMathTests/DoseMathTests.swift
trixing: What should happen for maxIOB = nil?  throw an error like all the other missing preferences or not have a limit?

trixing:
"It's seems like the code is validating current IOB, not the potential IOB from enacting a dose. You could be at max - n but then allow a dose that puts you > max. It seems to me that this feature should be preventing that. "

Are you referring to the temp basal or bolus calculation?

For the temp basal calculation:
here
            if iob > maxIOB {
I guess you want me to check against (iob + determinedRate/2)?  To be even more correct I'd also need to pass in the lastBolus as long as it is non-nil?

For the bolus calculation - I think that already takes the last basal rate into account.  I guess I should take the lastBolus amount into account here as well?

Thanks for the feedback.

ps2:
Please re-open this as a PR against dev.

scottleibrand:
@ps2 If you're not already aware, Github recently added a feature that lets you (as the maintainer) change the branch a PR targets, via an edit button at the top.  That lets you preserve all the comment history while still getting it retargeted as needed.  Not sure if that's what you need in this case, but it took me quite awhile to figure out that feature existed, so not sure if you know about it yet.

ps2:
Cool, no, I didn't know that. Will have to check it out. Thanks!

PR #397, opened 2017-03-11
Add a switch to turn on/off Glucose Momentum

trixing:
Motivation is mostly that it duplicates the retrospectiveCorrection.  Having both on can double the effect leading to over and under-correction.  This way the user can easily chose and compare what works best.


loudnate:
Hey @trixing, thank you for your willingness to submit these PRs. I think it's necessary for me to address this one with an author's perspective.

It's not correct to say momentum duplicates retrospective correction. They are not intended to be interchangeable, and they work in concert: The effects of momentum are not added to the effects of insulin/carbs/retrospective correction, but blended with those effects. 

Within 5 minutes from the last glucose reading, momentum is 100% responsible for the next prediction, and the other effects are 0% representative. At 15 minutes, momentum is 50% responsible. At 30 minutes, 0%. Using momentum to transition between measured glucose and predicted glucose prevents visual discontinuity between the two lines and, in my experience, more accurately reflects how glucose will change.

I believe adding additional settings to this "mainline" repo should be taken very seriously, both because of the risk of creating confusion, and the increased difficulty of validating each possible configuration. 

My recommendation at this time is that this particular change is better kept in your personal fork.
YDMV, which is why Loop and LoopKit are open-source projects, originated with a focus on code clarity and organization to allow personalized experimentation.

trixing:
Hi Nathan,

thank you for your thoughts on this subject.  I actually agree that keeping the number of settings to a minimum is a good thing.  Otoh you say yourself the point of this project is to allow users to experiment.  I think the super conservative approach should not be to add any knobs which are clearly understood and Loop should not do any prediction for that reason but only operate on current IOB/COB by default.

In our experience (and I think there is another bug report about this) both  momentum and retrospective correction together tend to overshoot unexplained raises and lows.  Both actually have the roughly same effect in terms of predicting future changes, though I think the retrospective one is a bit more clever.

I think the phasing in you are referring to could probably be applied to the retrospective values as well as momentum to avoid the discontinuity,  e.g. in predictGlucose in LoopMath, 'momentum' could just be another effect and blend in logic could operate on the sum of effects.

Anyways, I'm fine keeping this in my own branch if you feel strongly about this.

ps2:
@trixing reading your above statement, I'm not sure that you understand the math of how these features are integrated into the forecast.  They do not add cumulatively.  Retrospective correction is added cumulatively, like all the other effects timelines, but momentum is just a linear effect that *overrides* all the other forecast components, with a quick fade out. 

When you say "phasing in ... could probably be applied to the retrospective values": momentum is not phased in, but phased out.

Turning off momentum seems like an interesting thing to experiment with, and I'd be curious about others' experiences with doing so. That can be done on a personal customization basis.  If it turns out that this is a useful change then let's look again at how we roll it out to more users.

PR #400, opened 2017-03-12
Guard against unsupported and too high bolus amounts

trixing:
- Negative and non 0.1 increment are flat out refused
- bolus above recommended needs the user to confirm the amount (to prevent dangerous mistakes, ie. 11 instead of 1.1 or so)


ps2:
Boluses of 0.025 and 0.05 increments are perfectly reasonable on x23+ pumps, and a feature desired by folks with high insulin sensitivities.  

Also, could you describe the user scenario that you find the warning for exceeding recommended bolus useful?  I generally feel that when a user has taken specific action to do something outside the norm, that you don't need to warn, but maybe I don't understand the scenario this is trying to solve for.

trixing:
Okay, I didn't know about the 0.025 - going to change.  This (and the negative part) is purely to prevent unnecessary "Bolus failed" messages.

The exceeding recommended is useful in two scenarios:
- typos, ie. mixing up 11 and 1.1 for example
- think'o's - it makes the user (parents, teachers, patient, etc) think twice about giving more:  Usually loop should be fine and imho in Loop you'd rather enter extra carbs to achieve the same effect anyways, as otherwise it will just 'low temp' after the excessive Bolus to offset the extra insulin.  Maybe we should add this to the alert text.
- kids: I want my kid to use this as much unintended as possible and rather prevent any avoidable mistakes.  I'm actually considering in my private branch to remove the ability alltogether for the reason above.

loudnate:
> This (and the negative part) is purely to prevent unnecessary "Bolus failed" messages.

Values that are less than 0 or use an unsupported precision do not trigger "Bolus failed" messages in any way. I don't believe there's even a very easy way for a user to enter a negative value anyway.


> typos, ie. mixing up 11 and 1.1 for example

The maximum bolus setting could easily handle this specific case.

trixing:
Well, there is a '-' sign on the keypad.  They do cause a Bolus to go silently non working.

Instead of 11 and 1.1 it could also be 3.0 vs. 0.3 or 2.1 vs. 1.2

Think of it as robustness features making the app easier to use for people who do not understand the inner workings.

walker0:
@trixing I'm not sure if there is something different with our configurations but there is no - sign on my keypad. 

Could a possible solution be to force entry of a decimal. So if you wanted 1 unit you would have to type  '1.0' . Annoying but it solves part of the issue in terms of incorrect entry.

trixing:
Hmm, you are right, I just double and can't see it either anymore.  Not sure then what made me add this then.  Let's close this pull request and I'll just keep it in my private branch.  It's easy enough.

PR #405, opened 2017-03-14
Do not show glucose in HUD view if older than 15 minutes

trixing:
to not mislead people into wrong glucose levels.


ps2:
Please reopen this against dev.

PR #409, opened 2017-03-19
Update Carthage to work with ios10.3 & Xcode 10.3

walker0:


ps2:
Thanks Matt. I can't merge this in because the frameworks are compiled with a beta version of Xcode, but once GM is out, we'll need this.

PR #418, opened 2017-03-28
Rebuild all frameworks with Swift 3.1

bharat:
Comand used: $ carthage build --platform iOS,watchOS


bharat:
this is not ready for merge - there are a bunch of compilation failures. I'll dig into those.

bharat:
I'm going to take a different route here. disregard this PR, I'll submit a new one.

PR #419, opened 2017-03-29
Clean build on Swift 3.1

bharat:
Using two forks temporarily.

Using bharat/SwiftCharts because we need the extension safe commits from
master, but don't want the rest of the master branch yet.

Using bharat/LoopKit because we want the warning fixes from
https://github.com/LoopKit/LoopKit/pull/111 which is not yet merged into
LoopKit/dev


bharat:
new-overlords-i-for-one-welcome-our-new-swift-31-overlords

ps2:
I kicked the build, and I *think* it passed?  But it's not showing up as passing here yet for some reason.

Loop/Managers/CGM/DexCGMManager.swift
```diff
@@ -100,7 +100,7 @@ final class ShareClientManager: CGMManager {
     var debugDescription: String {
         return [
             "## ShareClientManager",
-            "latestBackfill: \(latestBackfill)",
+            "latestBackfill: \(latestBackfill.debugDescription)",
```
ps2: It's preferable to call String(reflecting: x) vs calling .debugDescription directly because that works with any instances, including those cases when the runtime will synthesize the string representation for you.

bharat: will fix everywhere.

ps2:
The frameworks in here need to be built from a LoopKit release tag.

bharat:
@ps2 when you say `the frameworks in here need to be built from a LoopKit release tag` I assume that you mean that the Cartfile should contain a line like this:

```
github "LoopKit/LoopKit" ~> 1.2.0
```

happy to do that, but I wasn't sure which release tag for LoopKit would cover this commit:
https://github.com/LoopKit/LoopKit/commit/b480b45a28b173906e86b9f302c2f86e4b3ac389

not sure how carthage grabs those version numbers and I don't see a commit in LoopKit/dev that bumps the version up. can you advise?

ps2:
The LoopKit warnings are not preventing a carthage build, right? We can merge this for people to run with Xcode 8.3, even still pointing to the LoopKit with warnings.

bharat:
I think so, but let me verify by going back to LoopKit 1.2.0 and see.

bharat:
seems to be fine.

bharat:
I take it back - that won't compile because I've already updated the various APIs to track what's in LoopKit/dev. Reverted the revert - this PR is now pegged to LoopKit/dev again.

ps2:
For the sake of keeping things organized, and more easily revertible, please split this into at least two PRs; one for purpose of updating Loop to Swift 3.1, which includes fixing the warnings, and recompiling frameworks, without updating those frameworks.  Updates to external frameworks, and in particular, those that bring in new functionality, like the LoopKit v1.2.1 change, should be done separately.

bharat:
ok - so to be clear - you want:
1) one PR that doesn't change the Cartfile but compiles with swift 3.1
2) one PR that updates the Cartfile (LoopKit 1.2.1,dexcom-share-client-swift 0.2.1, G4ShareSpy 0.3.2)

is that right? happy to oblige.

ps2:
Since LoopKit 1.2.1 has functionality changes in it, and requires functional changes in Loop, that seems like it should be its own thing too.   Are dexcom-share-client-swift 0.2.1, G4ShareSpy 0.3.2 just swift updates?  They can maybe be postponed until we need them.

bharat:
The updates in G4ShareSpy and dexcom-share-client-swift are relatively minor compatibility changes - no API changes. I'll roll up a separate PR for those two - should be fine to do them together.

bharat:
ok - (1) is now in https://github.com/LoopKit/Loop/pull/421

bharat:
ok, here's what we have now:

1) https://github.com/LoopKit/Loop/pull/421 - rebuild all frameworks with swift 3.1 and clean up warnings including moving off of using some deprecated APIs
2) https://github.com/LoopKit/Loop/pull/422 - update dexcom-share-client-swift and G4ShareSpy frameworks to new swift 3.1 clean versions
3) https://github.com/LoopKit/Loop/pull/423 - update LoopKit to v1.2.1 

I believe I've accounted for everything - but wouldn't be surprised if I missed something in there. Will close this PR as those three supercede it.

PR #422, opened 2017-03-30
Swift 3.1 - update dexcom-share-client-swift and G4ShareSpy frameworks

bharat:
This should be independent of other framework update PRs.


bharat:
I'll resolve conflicts on this when #423 and #424 get merged.

PR #434, opened 2017-04-03
Widget should always display data using the preferred dispay unit,

### LINKED ISSUE #433: Unit display on new glucose graph on lock screen
Testing dev https://github.com/LoopKit/Loop/commit/22c42cebc743390cc5d80b70cd7d7c13aee96ba8 and the new glucose graph on the lock screen is using mg/dl unit even though I have selected mmol/L

image

bharat:
regardless of the units used in the StatusExtensionContext.

This resolves #433


bharat:
Note that I've only tested this in the simulator so far, but that was a useful test case because Loop hardcodes mg/dL predicted glucose values in the simulator.

bharat:
I'm seeing some other weirdness on my device when I switch units - the chart range doesn't update properly, I'm seeing the historical glucose in one unit and the predicted in another, etc. there's still work to do here.

PR #477, opened 2017-05-18
Reinstate upload of enlite bg history upload to ns

ps2:


Telefingers:
How and where do I do this?  I installed it using the easy way...does this mean I should be using the GitHub installation in order to install this so that NS displays?

ps2:
I've cut a small patch release, so you can just build with the latest master now. https://github.com/LoopKit/Loop/releases/tag/v1.3.3

PR #491, opened 2017-06-03
implemented setting to configure default carb absorption times

mpangburn:
Will require [LoopKit #132](https://github.com/LoopKit/LoopKit/pull/132) order to build/run.
Adds a cell in the configuration section of the settings tab that allows users to configure the three default carb absorption times.


Loop/Extensions/NSUserDefaults.swift
```diff
@@ -10,15 +10,17 @@ import Foundation
...
 extension UserDefaults {
-
+    
     private enum Key: String {
         case basalRateSchedule = "com.loudnate.Naterade.BasalRateSchedule"
         case batteryChemistry = "com.loopkit.Loop.BatteryChemistry"
         case cgmSettings = "com.loopkit.Loop.cgmSettings"
         case carbRatioSchedule = "com.loudnate.Naterade.CarbRatioSchedule"
         case connectedPeripheralIDs = "com.loudnate.Naterade.ConnectedPeripheralIDs"
+        case defaultAbsorptionTimes = "com.loudnate.Naterade.DefaultAbsorptionTimes"
```
ps2: Could yout put this data into loopSettings? That will limit keys to make migration easier.

Loop/Extensions/NSUserDefaults.swift
```diff
@@ -10,15 +10,17 @@ import Foundation
 import LoopKit
 import MinimedKit
 import HealthKit
+import CarbKit
 
 extension UserDefaults {
-
+
```
ps2: Would you clean up these trailing whitespace diffs?

Issue #416, opened 2017-03-27
Add option to use Dexcom Share server outside US


peterlynton:
Add option in Dexcom Share settings to use existing US server our non US server (US:share1.dexcom.com, non US:shareous1.dexcom.com)

I have tried manually editing Loop>Loop>Managers>KeychainManager+Loop.swift to update DexcomShareURL.
This gives the error 'The operation couldn't be completed. (ShareClient.ShareError error 1.) loginError("SSO_AuthenticatePassword invalid"), but keeps the account details.

dabear:
Hi. Workaround: 
This would require an updated dexcom-share-client framework. 
You can manually change this by forking the framework from here: https://github.com/mddub/dexcom-share-client-swift
You would have to edit ShareClient/ShareClient.swift and replace the two references to dexcomServerUS with dexcomServerNonUS, and then in Loop directory update the carthage file from
`github "mddub/dexcom-share-client-swift" ~> 0.2.1` to `github "yourgithubuser/dexcom-share-client-swift" "master"`
issue a `cartage update dexom-share-client` and then recompile Loop.  See also my project here, if you want to use nightscout instead of dexcom share for fetching glucose data: https://github.com/dabear/NightscoutShareServer

dabear:
This is now possible in Loop as of https://github.com/LoopKit/Loop/commit/d814d8847532b6b66768c2d1a00531797de84e94

![image](https://user-images.githubusercontent.com/442324/27803011-95887c18-6027-11e7-9a77-b044071aa59b.png)

This can probably be closed now @upsetter ?

ps2:
We usually wait until a feature is in a release branch to close the ticket, but then yes, I believe this can be closed!

Kdisimone:
Added in https://github.com/LoopKit/Loop/pull/535

PR #492, opened 2017-06-04
Make shareserver URL configurable in the GUI

dabear:
## Synopsis
This makes it possible to select which shareserver to use in the Loop UI and thereby solves issue #416 .

This requires updating the dexcom-share-client-swift to the latest master version, rather than specifying version 0.2.1 , as there isn't a tag available for the latest changes, which this depends on.

## Use cases
* It allows you to change to the NON_US dexcom share server. And users don't have to alter the Loop code to do so.
* You can use a fake shareserver for testing purposes. One such server, made by my self, is freely available at http://mockedshareserver.azurewebsites.net .
* When Looping with another cgm solution such as freestyle libre, librealarm and xdrip+ you wouldn't typically have access to a dexcom transmitter id. Without a dexcom transmitter id you can't use dexcom share. By using a proxy such as https://github.com/dabear/NightscoutShareServer you can still use Loop and share.

## Screenshot
img_0022


ps2:
Because it appears that the goal is to allow arbitrary urls, the ShareClient framework should be updated to just take a non-nil url.  The known URLs can be implemented as a public string backed enum in the framework. This will work better with the implementation in Loop.

For the UI, we should keep it simple for the common use case (selecting between US and EU servers), and make that a radio selection. To support custom URLs, which is a less common use case, users could edit the code for now.  To make it easier, we could add to the array of Name/URL tuples with a commented out third option and "type your server URL here and uncomment this line" block, then the radio selector would change from 2 to 3 options.

Lastly, RemoteDataManager needs to be updated to use the saved configuration; it looks like right now, after restarting Loop, it would revert to using US serversl

mddub:
Thanks a ton for doing this, @dabear. I just bumped the version of dexcom-share-client-swift with a `v0.3.0` tag, so you can specify that version explicitly in the Cartfile.

dabear:
@mddub thanks, however, I have some additional changes for you, please see https://github.com/mddub/dexcom-share-client-swift/pull/13 which should fix @ps2 first feedback here regarding "The known URLs can be implemented as a public string backed enum in the framework."

mddub:
@dabear okay, `v0.4.0` it is!

https://github.com/mddub/dexcom-share-client-swift/releases

dabear:
Closing this pull request. Most of this feature was added to the following branch: https://github.com/LoopKit/Loop/compare/share-server

PR #509, opened 2017-07-01
Take2: allow NON_US share server and custom share server

dabear:
This adds an option to select US and NON_US share servers in the GUI. Also, if the customServer variable is filled out, a third option for selecting "Custom" share server will appear in the GUI


Loop/Models/ServiceAuthentication/ShareService.swift
```diff
@@ -36,14 +38,23 @@ class ShareService: ServiceAuthentication {
                 value: url?.absoluteString,
                 options: [
                     (title: NSLocalizedString("US", comment: "U.S. share server option title"),
-                     value: DexcomShareURL.absoluteString)
+                     value: DexcomShareURL.absoluteString),
```
ps2: DexcomShareURL should be replaced everywhere with the new string constants from Mark’s lib

Loop/Models/ServiceAuthentication/ShareService.swift
```diff
@@ -36,14 +38,23 @@ class ShareService: ServiceAuthentication {
...
+        if customServer.characters.count > 0 {
+            credentials[2].options?.append(
+                (title: NSLocalizedString("Custom", comment: "Custom share server option title"),
+                value: customServer))
+        }
+
         if let username = username, let password = password, url != nil {
             isAuthorized = true
-            client = ShareClient(username: username, password: password)
+            client = ShareClient(username: username, password: password, shareServer: url!.absoluteString)
```
ps2: Should avoid use of ! with let url = url, in the if statement above

Loop/Models/ServiceAuthentication/ShareService.swift
```diff
@@ -74,9 +85,10 @@ class ShareService: ServiceAuthentication {
             return
         }
 
-        let client = ShareClient(username: username, password: password)
+        let client = ShareClient(username: username, password: password, shareServer: url!.absoluteString)
```
ps2: Avoid !

Loop/Models/ServiceAuthentication/ShareService.swift
```diff
@@ -36,14 +38,23 @@ class ShareService: ServiceAuthentication {
...
-                     value: DexcomShareURL.absoluteString)
+                     value: DexcomShareURL.absoluteString),
+                    (title: NSLocalizedString("Outside US", comment: "Outside US share server option title"),
+                     value: KnownShareServers.NON_US.rawValue)
+
                 ]
             )
         ]
 
+        if customServer.characters.count > 0 {
```
ps2: Using nullable/optional var here would be better than counting chars.

Loop/Models/ServiceAuthentication/ShareService.swift
```diff
@@ -17,6 +17,8 @@ class ShareService: ServiceAuthentication {
     let title: String = NSLocalizedString("Dexcom Share", comment: "The title of the Dexcom Share service")
 
     init(username: String?, password: String?, url: URL?) {
+        // change the contents of the variable below to point to your custom share server
+        let customServer = ""
```
ps2: Feel free to expand upon how to do this; like linking to your repo/blog.  Also, do you want to describe how to change the name as shown in the UI from "Custom"?

Loop/Models/ServiceAuthentication/ShareService.swift
dabear: I started doing that, but realized having an optional set to an initial state of nil would be more difficult to grasp for novice users not accustomed to editing code, instead of editing just the contents between two quotation marks. 

I guess i could use an optional with a more extensive description

Loop/Models/ServiceAuthentication/ShareService.swift
dabear: Yeah, agreed. Can you explain why the url != nil was added in the first place (rather than let url = url)..?

Loop/Models/ServiceAuthentication/ShareService.swift
ps2: Or maybe don't even set a string at all in the no-custom-server case; just have a block of code that is completely commented out.

Loop/Models/ServiceAuthentication/ShareService.swift
ps2: url wasn't used at that point; it would not compile with a let (would be an unused variable), so it was just a hint.

dabear:
The requested changes should now been added. Please verify  :)

PR #515, opened 2017-07-06
Fix share server select issue

### LINKED ISSUE #514: Have to tap and select Server option to add Dexcom Account
Just built Loop with dynamic-carb-absorption branch #507 and when I went through initial setup to add Dexcom Share account I noticed when I would input my username & password it wouldn't let me save the account before first tapping on Server and selecting US (the only option at the moment)

Video demonstrating issue with me narrating - http://granu.be/lPz5

ps2:
Fixes #514


loudestnoise:
👍 Tested and this does allow you to tap Add Account without having to first tap on Sever first.

PR #539, opened 2017-07-31
Insulin models testing branch

ps2:
This PR is for testing a new [exponential insulin model](https://github.com/LoopKit/Loop/issues/388#issuecomment-317938473). You can see [how it is implemented in LoopKit](https://github.com/LoopKit/LoopKit/blob/exponential-insulin-model/InsulinKit/ExponentialInsulinModel.swift)

The example settings currently in this branch are a duration of 360 minutes, with peak activity, which looks like this when compared to the existing Walsh curve in loop:

loopiob_scalableexp_ipynb_at_master_ _ps2_loopiob

&#x1F534; **Configuration**: The Insulin Action Duration setting in Loop config is ignored in this branch: you should instead [configure your own custom settings](https://github.com/LoopKit/Loop/blob/insulin-models-testing/Loop/Managers/LoopDataManager.swift#L70-L73).

I have created a python notebook you can use to [visually compare different settings](https://github.com/ps2/LoopIOB/blob/master/ScalableExp.ipynb) to your current Loop curve.


francesc0-cgm:
Some sample to test? In your opinion for novolog on a kid?

dm61:
Running exponential-insulin-model branch now. Compiler comes up with bunch of warnings, which I assume we can safely ignore. In Settings, Insulin Action Duration came in set to 6 hours, overriding my earlier setting. Action duration and peak action time can be manually adjusted here:  https://github.com/LoopKit/Loop/blob/insulin-models-testing/Loop/Managers/LoopDataManager.swift#L70-L73. As of now, I intend to keep 360min and 75min (for Novolog) for some time, and observe.

ps2:
Yes, you can ignore the warnings. I've updated the issue description to be more clear about the Insulin Action Duration setting being ignored, and how you should configure it.

ps2:
@francesc0-cgm See the [example settings](https://github.com/LoopKit/Loop/blob/insulin-models-testing/Loop/Managers/LoopDataManager.swift#L70-L73) for an example of Novolog/Humalog for a child. That represents my best guess, based on observed data for four kids that I ran through a linear regression model.

francesc0-cgm:
Thank you @ps2 in a couple of days i will start to test it!

francesc0-cgm:
After a 10 days test using it with 60 of time of peak and 300 of td as ps2 adviced, i found it is quite similar to real insulin absorption. I had better results, less dip after meal with erroneous 0 temping, and no problem at all with bolus dosing (i dropped a point in csf for other reason). For me with a setting menu to choose peak and td in thr options menu is ready to dev.

dm61:
After two weeks of running this branch: counteraction (Glucose Change) graphs look cleaner, no phantom dips or negatives. Overall better predictions and improved dosing. 

I've spotted just one strange anomaly, just once, and probably unrelated to any changes in this particular Loop branch: after a carb entry (80g), the bg prediction did not reflect the entry properly. The screenshot clearly showed the newly entered carbs, but the bg graph remained he same as before the entry was made:
img_4587
Here is the issue report captured at that point: 
[issue_report_erroneous_prediction_after_carb_entry.txt](https://github.com/LoopKit/Loop/files/1220714/issue_report_erroneous_prediction_after_carb_entry.txt)

I then went back to the carb page and edited the entry - changed carbs from 80g to 81g.  After that, the entry was properly taken into account in the bg prediction, as shown in the screenshot taken just after the edit was made:
img_4588
Here is the issue report captured  at that point: 
[issue_report_correct_prediction_after_carb_entry_edit.txt](https://github.com/LoopKit/Loop/files/1220715/issue_report_correct_prediction_after_carb_entry_edit.txt)

Possibly a really unusual timing issue?

rsilvers129:
I was having trouble with the compiler saying that the watch app had the wrong name - it was still picking up .com.loopkip.loop and not my own name. I deleted the watch app to get it to compile.

Kdisimone:
@rsilvers129 saving the Xcode project after changing the main app bundle id will update those app names. Xcode sometimes gets them right if you don't save...but glitches often enough that it's worth saving before building so that the updates for sure happen.

PR #572, opened 2017-09-03
Prevent repeat ICE values

ps2:


rsilvers129:
I am running the latest dev as of yesterday, and when using iosxdripreader I am getting an issue that may be related to this:

[text-0EF033CFA129-1.txt](https://github.com/LoopKit/Loop/files/1282368/text-0EF033CFA129-1.txt)
img_8930

Issue #550, opened 2017-08-07
Idea: Add option to have dexcom share as primary


joeyovens:
Add option to have dexcom share as primary glucose source and use transmitter as backup. Live in an area where cell service isnt great for data, but it would allow for more reliable connection when it is or on wifi.

jeremybarnum:
We have defaulted to forcing to the cloud; if fallback to the could were more reliable this would be less of an issue but as it stands for us anyway it would be useful.  Need to go back and check the rileylinktimer modification which I think might be another way to approach this.

Issue #547, opened 2017-08-03
Deleting carb entry in Dynamic Carb Absorption doesn't remove carbs from carbs on board


eszcloud:
When I entered carbs for food from both my watch and on the phone and then deleted one entry, the carbs of the deleted entry still seem to be included in the calculation of carbs on board. (v1.3.4dev)

The screenshot below shows an entry for 37 g carbs (and 37 g total carbs) but 65 g COB. 

![img_0956](https://user-images.githubusercontent.com/14155343/28939887-a810a2f6-7847-11e7-90bf-b269959a4e7a.jpg)

ps2:
This is just a display/cache issue. It will be fixed once the next version of LoopKit is incorporated.

eszcloud:
Good to know. Thanks for following up and having a fix already.

Issue #523, opened 2017-07-17
Font issue in HUD


jmatheson:
If user has increased font size in their iOS settings it ends up clipping in the HUD.

![image](https://user-images.githubusercontent.com/9138265/28293793-54d3d15e-6b99-11e7-9ab3-7934627f818b.png)

russfg:
I see the issues above, but also the time under my charts is not spaced proportionally. I have the regular size iPhone 7,  with the Settings>Display&Brightness>Text Size  1 below the max.
![890719b2-3937-4018-a54d-38c1e081b07f png](https://user-images.githubusercontent.com/25555257/28982051-ef04f5fe-7919-11e7-8316-82c489e88426.PNG)
![654fd2e4-5e30-4512-bbdc-56905f7b5530 png](https://user-images.githubusercontent.com/25555257/28982056-f3b197b0-7919-11e7-9739-b9f95e107c6c.PNG)

gkirma:
I had this issue as well and decreased the font size but it didn't correct on my Loop screen. Please offer some suggestions? Thank you!

Issue #438, opened 2017-04-04
Issue Report does not include lastBolus


ps2:
To diagnose bolus recommendation issues, it's useful to know if there are any pending boluses, as well as total pending insulin.  This is not included in the current diagnostic report from LoopDataManager.

PR #584, opened 2017-09-20
v1.5.0

ps2:
* The iOS 11 and watchOS 4 upgrades are optional, but the Xcode 9 upgrade is mandatory
* Carthage and Brew setup are automatic for new installs

### Features
* **Pre-meal correction targets**: Set a new correction range override in settings for pre-meal times. This can be enabled from the toolbar. This override is active for 1 hour or until carbohydrates are added. Fixes LoopKit/Loop#177
* **Insulin Delivery in Health**: Insulin delivery data is stored using HealthKit and can be viewed in the Health app and other HealthKit-enabled apps.
* **Safer correction calculations**: Insulin corrections are now calculated ensuring the minimum glucose value does not drop below the correction target, rather than only the final glucose value. Thanks to dm61 and his work on LoopKit/Loop#533

### Settings
* New models of insulin effects are available in settings along with an easy way to compare their curve shapes. The Walsh/Minimed DIA model is now considered deprecated. Fixes LoopKit/Loop#388
* Retrospective correction is now enabled by default. It is designed to work closely with other Loop features and it’s recommended for all users.
* Target Range is now known as Correction Range to avoid confusion with the ranges often used for "Time in Target"
* Minimum BG Guard is now known as Suspend Threshold. It's now optional; if not set, the lowest value in the active Correction Range is used.

### Reliability
* Subscription to the RileyLink connection timer is now updated dynamically based on the connection staleness of CGM and pump data. This should help resolve intermittent connection issues more quickly.
* Loop now takes more opportunities to fetch data from Share. If you use the Dexcom G5 app for your transmitter, *don't delete the transmitter ID to try and improve communication*. It won't have any effect. Fixes LoopKit/Loop#550
* Pump history reconciliation is improved to identify insulin delivery as being suspended during No Delivery alarms and Rewind states
* More Loop failure cases are tracked and can be investigated by tapping a red Loop indicator. This includes a mis-calibrated CGM and failure to read Pump History. 
* Bolus delivery is more tightly-integrated into the flow sheet screen. While a bolus is starting, an indicator is shown at the top of the screen. After a bolus has started, the Glucose, Active Insulin, and Insulin Delivery charts will optimistically update to reflect the programmed amount. Fixes LoopKit/Loop#438 LoopKit/Loop#273

### Bug Fixes
* Rotating the charts can get into a state where too little or too much data is shown
* System font size changes result in overlapping text labels in many places. Fixes LoopKit/Loop#523
* Carbs on Board values can be stale for a few minutes after deletion. Fixes LoopKit/Loop#547


francesc0-cgm:
Thank you @ps2 for your awesome work. The third new feature means we can bolus all the suggested amount safely also using longer-DIA exponential curves?

ps2:
@francesc0-cgm Yes, with large meals, and new insulin curves, Loop will reduce the bolus amount so that the forecast does not drop below your suspend threshold.   A reasonable suspend threshold is something > 65mg/dl.  That limit may be enforced in a future release.

francesc0-cgm:
Thank you @ps2

PR #631, opened 2017-11-22
A few fixes to avoid compile-time warnings 

elnjensen:
@ps2  Just a couple of fixes that remove warnings that Xcode throws when compiling - one fix for character count syntax change in Swift 4 (in two places), and one to add a 1024x1024 icon that it seems to want.  For the latter I just scaled up one of the other icons.  Seems somewhat unnecessary but nice to avoid warnings that could trip people up.  Feel free to cherry-pick any part that's useful here.


ps2:
app store copy 2x

Full res asset.

PR #635, opened 2017-12-03
Add active insulin (IOB) value in Today widget

elnjensen:
Show IOB value above the glucose chart in the lock screen status widget (Today widget).


Common/Models/StatusExtensionContext.swift
```diff
@@ -138,6 +144,32 @@ extension NetBasalContext: RawRepresentable {
...
+            "error": error
+        ]
+    }
+
+    init?(rawValue: RawValue) {
+        guard
+            let value     = rawValue["value"] as? Double,
+            let startDate = rawValue["startDate"] as? Date,
+            let error     = rawValue["error"] as? String
+            else {
```
bharat: nit: indent the else statement to line up under the guard statement.

Common/Models/StatusExtensionContext.swift
```diff
@@ -274,6 +306,7 @@ struct StatusExtensionContext: RawRepresentable {
     var loop: LoopContext?
     var netBasal: NetBasalContext?
     var batteryPercentage: Double?
+    var activeInsulin: ActiveInsulinContext?
```
bharat: you should bump the private version number above to reflect the fact that you've added a new member to the context

Loop Status Extension/StatusViewController.swift
```diff
@@ -205,11 +205,24 @@ class StatusViewController: UIViewController, NCWidgetProviding {
...
 
+                    var activeInsulinString = "--"
+                    if let activeInsulin = context.activeInsulin {
+                        // If the returned value doesn't indicate an error,
+                        // assign it to the label:
+                        if activeInsulin.error == "" {
+                            activeInsulinString = String(format: "%.1f", activeInsulin.value)
+                        } else {
+                            // This will be ugly but maybe (?) informative: 
+                            activeInsulinString = activeInsulin.error
```
bharat: what will this look like? I haven't looked up what doseStore errors look like, but we don't want it to be misleading (ie, a small integer value here could be seriously confusing or potentially very harmful). I think I'd want to see us put in a non-numeric value of known size, like "??" or "ERR"

Common/Models/StatusExtensionContext.swift
```diff
@@ -29,6 +29,12 @@ struct NetBasalContext {
     let end: Date
 }
 
+struct ActiveInsulinContext {
+    let value: Double
+    let startDate: Date
+    let error: String
```
bharat: I'm not sure that it makes sense to pass the actual error string through to this point. showing it in the widget likely won't make a lot of sense. I think you have a few options:

1) if there's an error, don't add an ActiveInsulinContext and leave it blank in the widget
2) if there's an error, don't add an ActiveInsulinContext and report "ERR" in the widget
3) let error be an optional String so that you're not checking for "", but for the existence of the error

Loop Status Extension/StatusViewController.swift
```diff
@@ -205,11 +205,24 @@ class StatusViewController: UIViewController, NCWidgetProviding {
                 if let eventualGlucose = predictedGlucose.last {
                     let formatter = NumberFormatter.glucoseFormatter(for: eventualGlucose.unit)
 
+                    var activeInsulinString = "--"
+                    if let activeInsulin = context.activeInsulin {
+                        // If the returned value doesn't indicate an error,
+                        // assign it to the label:
+                        if activeInsulin.error == "" {
```
bharat: see my comment in `struct ActiveInsulinContext` - instead of checking for empty string, I think you should let error be an optional string value and check for nil.

Loop/Extensions/NSUserDefaults.swift
```diff
@@ -80,7 +80,7 @@ extension UserDefaults {
                     return .enlite
                 }
 
-                if let transmitterID = string(forKey: "com.loudnate.Naterade.TransmitterID"), transmitterID.characters.count == 6 {
+                if let transmitterID = string(forKey: "com.loudnate.Naterade.TransmitterID"), transmitterID.count == 6 {
```
bharat: this compatibility fix (and others) should probably be in its own PR to avoid merge conflicts, but I'm not going to be a stickler (maybe somebody else will be :-) )

Loop/Managers/StatusExtensionDataManager.swift
```diff
@@ -134,6 +135,30 @@ final class StatusExtensionDataManager {
...
+                // isn't used in the code, but could e.g. check how
+                // recent it is. 
+                switch result {
+                case .success(let iobValue):
+                    context.activeInsulin = ActiveInsulinContext(value: iobValue.value,
+                                                                 startDate: iobValue.startDate,
+                                                                 error: "")
+                    
+                case .failure(let error):
+                    // Use a large negative value to flag that we
```
bharat: signaling with invalid values is a dangerous pattern, especially in systems like this. If we're concerned about the possible downstream uses of this value (and we should be) then I think the only safe thing to do in an error situation is to not return an ActiveInsulinContext.

So, instead of making error an optional String (as I suggested elsewhere), you can leave context.activeInsulin nil in case of error. Since we know that this value is expected in this version of the code, the fact that it's nil indicates that there's an error. That will simplify your downstream code and avoid fake values.

ps2: 👍

ps2:
Thanks, @elnjensen! 

And comments look good, @bharat!

Loop Status Extension/StatusViewController.swift
ps2: This should use NumberFormatter, not "%.1f "

Loop Status Extension/StatusViewController.swift
```diff
@@ -205,11 +205,24 @@ class StatusViewController: UIViewController, NCWidgetProviding {
...
+                            activeInsulinString = activeInsulin.error
+                        }
+                    }
+
                     if let eventualGlucoseNumberString = formatter.string(from: NSNumber(value: eventualGlucose.quantity.doubleValue(for: unit))) {
                         subtitleLabel.text = String(
                             format: NSLocalizedString(
-                                "Eventually %1$@ %2$@",
-                                comment: "The subtitle format describing eventual glucose. (1: localized glucose value description) (2: localized glucose units description)"),
+                                "%1$@ U IOB;   Eventual BG %2$@ %3$@",
```
ps2: The preferred wording here is "Eventually", to match the app.

Loop Status Extension/StatusViewController.swift
elnjensen: @ps2  Happy to switch to NumberFormatter, but for my own education, what's the advantage there?  Is it that it handles locale information?

Loop/Extensions/NSUserDefaults.swift
elnjensen: Agreed - it actually got in there by accident.  Wasn't working from the branch I thought I was...

Loop/Managers/StatusExtensionDataManager.swift
elnjensen: OK, agreed with all of the above.  I will tackle these changes soon.  Thanks for the feedback!

Loop Status Extension/StatusViewController.swift
elnjensen: That's what I had originally, but here I think the situation is a bit different because of the addition of the IOB value.  In the app, it's clear that the number after "Eventually" refers to BG, since there it says "Glucose" earlier on the line and it's right above the glucose graph (so the full text is "Glucose   Eventually 100 mg/dl".  The placement above the glucose graph is the same here, but it's complicated by having the IOB there, too.  So if we follow the app, you'd get:

 0.6 U IOB;  Eventually 100 mg/dl

which to me doesn't make as much sense at a glance as 

 0.6 U IOB;   Eventual BG 100 mg/dl

since we've reported IOB early in the string and then we're saying "eventually" another number.  No one would *really* confuse that for being the eventual IOB, I don't think, but it does seem incongruous to me.  

What do you think?   And while we're on formatting, I'll point out another small thing: originally I had the number and the "IOB" switched, like this:

 IOB 0.6 U;   Eventual BG 100 mg/dl

but I found that in my actual use, the glanceability was slightly worse; I spent just an extra fraction of a second mentally finding that value in the middle of the string.  With it all the way to the left, it's definitely easier to mentally process at a glance.  

Similarly, I initially had two decimal places, matching the app, but decided that again, the mental information processing was simpler with a cleaner number there, and that the IOB in the hundredths place wasn't meaningful (i.e. the rounding of 0.05 U of insulin won't matter for me, though I suppose it could for someone who is super sensitive).   That one I don't feel as strongly about; could go back to two places here as a default if you think that's better. 

Just FYI about some small design decisions in case you're wondering about that as we discuss format here.   Both of these were usability / legibility related for me.

elnjensen:
OK, thanks for the comments @bharat and @ps2.  I think I've addressed all of the issues you raised, but let me know if there are other suggestions or fixes needed.  I changed "eventual" to "eventually" (though kept the addition of "BG" in there).  And I decided that the whole struct for ActiveInsulinContext wasn't needed; since we're only using the Double value, I changed to that (akin to how batteryPercentage is handled).

bharat:
looks good to me. much simpler!

Loop Status Extension/StatusViewController.swift
```diff
@@ -205,11 +215,20 @@ class StatusViewController: UIViewController, NCWidgetProviding {
...
+                    } else {
+                        activeInsulinString = "--"
+                    }
+
                     if let eventualGlucoseNumberString = formatter.string(from: NSNumber(value: eventualGlucose.quantity.doubleValue(for: unit))) {
                         subtitleLabel.text = String(
                             format: NSLocalizedString(
-                                "Eventually %1$@ %2$@",
-                                comment: "The subtitle format describing eventual glucose. (1: localized glucose value description) (2: localized glucose units description)"),
+                                "%1$@ U IOB;   BG eventually %2$@ %3$@",
```
ps2: UI shouldn't be positioned using a semicolon and two space characters. The IOB value belongs in its own label. Perhaps in the lower left

elnjensen: OK, agreed about the separate label.  I thought of that at the start, but wimped out because it was easier to add it to the existing label.  But it's good motivation for me to keep learning more Swift. 

But if it's to remain visible in the collapsed view of the widget (see screenshot), I think it makes sense to keep it more or less where it is.   And I also think that from a usability standpoint it's useful to have both numbers relatively close to each other, to take them in in one glance. 

Perhaps both labels could be on that line, with one flush left and one flush right?  Are there potential issues with them overlapping? 

iob_widget

bharat: thanks for posting the screenshot. I agree that it's a little weird to put the two strings together. It's also a little weird that one of them is "<##> IOB" and the other is "BG <##>". I like the flush left, flush right option. So combining the two and normalizing the strings would yield:

IOB: 1.1U        BG: Eventually 66 mg/dL
|<-------        ---------------------->|

I don't think you'll have an issue with them overlapping.

ps2:
two_labels

elnjensen:
OK, IOB text is now in a separate label.  I'm not sure if I have set all the constraints on it in the most efficient way, so let me know what you think.  I also reverted the labeling to be more consistent with the rest of the app.  

I'll note that there's a small issue with clipping of the uppermost y-axis value when the widget is in expanded mode (see screenshot), but this isn't new; it was present before I started this branch.  I tried setting the label height (of both parts of the label) to be smaller, but it seems to want to stay at 50, I think because of a readability constraint that's set.  Any thoughts there are welcome, though as I noted this has been there for a while and doesn't have to be fixed here.

elnjensen:
There's also an issue with the x-axis range of the plot - I'm guessing it's because the glucose forecast extends longer than it used to.  But that's a separate issue; I'll try to tackle that next once this is wrapped up. 😄

ps2:
Thanks @elnjensen!

wkcwells:
@elnjensen Thanks so much for getting this done!  I haven’t updated in a year or so, and was so happy to see this in the current version!!  👍

PR #678, opened 2018-02-18
Pull temp targets from NS

Kdisimone:
This will enable Loop to pull temp targets, and cancel temp targets, from NS's careportal.  Requires internet connection.  Local alerts are provided via Loop notifications when a remote temp target is set or canceled.  The last enacted temp target is recorded in the Override targets area.  Blue BG targets displays, as with pre-meal and workout targets.

in the event of two simultaneous temp targets (local and remote), local will take precedence.  

Much thanks to @kenstack for the work on this!  We've been testing it for some time and have enjoyed the feature.  (i'm just the PR maker and a tester.)


Kdisimone:
@ps2 , I've left the cartfile unchanged, but submitted the LoopKit change as a PR against master.

PR #682, opened 2018-02-18
Use Loop Settings to update NS profile

Kdisimone:
I'm cherry picking from @trixing work to get this feature that I want...automatically update NS profile with Loop settings.  Credit goes to him for the work...i'm just eager to get it into use.  :)

will require iOS 11.0, or you could add a check.  But, it looks like LoopKit v2 will require iOS 11.0, so i don't think that will be a conflict?


Loop/View Controllers/SettingsTableViewController.swift
```diff
@@ -69,6 +69,10 @@ final class SettingsTableViewController: UITableViewController, DailyValueSchedu
         super.viewDidDisappear(animated)
 
         dataManager.rileyLinkManager.setDeviceScanningEnabled(false)
+        // TODO: Do not upload every time we exit settings.  Perhaps have a check for change?
```
trixing: This check exists:

https://github.com/LoopKit/Loop/pull/682/files#diff-5b9067796f682df0e6723dfbb8aa0e6bR320

Kdisimone:
Thanks @trixing fixed the note.

ps2:
Implemented in https://github.com/LoopKit/Loop/pull/1064

PR #683, opened 2018-02-20
Adding info to the Loop Pill (preamble)

Kdisimone:
Based on user requests to have some info available without needing to click on the Loop pill to see inside...this brings a couple information pieces as a preamble by using the Loop name to transmit the info.  Since not everyone would prefer this option, also left an easy way to undo the change in the code and could add it to the customization section quite easily.


Kdisimone:
img_9707

Kdisimone:
and thanks again goes to @kenstack...this is a feature we've been asking for in our house.

Loop/Managers/NightscoutDataManager.swift
```diff
@@ -138,7 +138,51 @@ final class NightscoutDataManager {
...
+        else
+        {
+            loopEventualBG = "N/A"
+        }
+        
+        loopParams = "BGTargets (" + loopBGTlow + ":" + loopBGThigh + ") | EvBG " + loopEventualBG + " | " + loopName
+        
+        //upload loopParams instead of just loopName
+        //that is the only pill that has the option to modify the text
+        let loopStatus = LoopStatus(name: loopParams, version: loopVersion, timestamp: statusTime, iob: iob, cob: cob, predicted: predicted, recommendedTempBasal: recommended, recommendedBolus: recommendedBolus, enacted: loopEnacted, failureReason: loopError)
```
ps2: Rather than overriding the "name" field, a new field/fields should be added to the upload structure in the rileylink_ios project.

kenstack: @ps2 I can do that no problem, but how will NS display it ?  I dont think NS will just create a new pill.  But I may be wrong on that!

ps2: Yes, supporting this in NS would require an NS PR.

francescaneri: @kenstack Is it a complicated thing?

francescaneri: @Kdisimone @ps2 it is still possible to pull to the dev?

PR #687, opened 2018-03-04
Add dashed line in graphs to mark the current time.

lgruen:
Example: img_1298

### DIFF
--- a/Loop Status Extension/StatusViewController.swift
+++ b/Loop Status Extension/StatusViewController.swift
@@ -35,6 +35,7 @@ class StatusViewController: UIViewController, NCWidgetProviding {
                 axisLine: .axisLineColor,
                 axisLabel: .axisLabelColor,
                 grid: .gridColor,
+                lastDate: .lastDateColor,
                 glucoseTint: .glucoseTintColor,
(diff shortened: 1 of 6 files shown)


dabear:
Great! This is something I've wanted for a long time

francescaneri:
@ ps2 what do you think?

ps2:
Looks nice! The bg graph might not need the dashed line, but it may make sense to keep it there for consistency.  On the other (lower) graphs, I often use the touch readout to see where the current time lines up.

Kdisimone:
@ps2 for old BGs, the line stays with the last BG reported, until a new BG comes in.

ps2:
The more I think about this and look at it the more the vertical line seems unnecessary.  The top graph is very obvious where the forecast begins. For the insuliin and carb graphs, the values are always predictions, and thus there isn't the same distinction between current and predicted values.  For insulin delivery the future part isn't really a prediction either, but a display of the duration of the currently set basal.

marcusfpereira:
I was about to open a feature request for this line, I miss it a lot. Mainly on the Insulin Delivery graph.

PR #698, opened 2018-04-05
Spike master

ronnieg123:


Kdisimone:
@ronnieg123 I didn't have any intention of merging this into Loop at this point.  I'm not sure if you've seen the previous conversations on Gitter, but Pete has concerns about an audio file that runs in the background of Spike app and doesn't want to have that merged into Loop app.  So, the support that I've done for Spike app intergrations for Loop will be as a side-stream more than likely.  Unless something changes, just wanted you to know why it has been separated at this point in time.

PR #709, opened 2018-04-26
Do not display stale glucose data

erikdi:
Currently the UI sometimes keeps display the last Glucose value, which can be in the best case confusing, in the worst case dangerous.  Blank it out if the data is older than 15 minutes.


diggabyte:
Could this perhaps be addressed through a new `sensorAlertState` which indicates the value is stale, like ⏱or ⎋ or something, rather than blanking it out completely?

PieterGit:
I would follow @diggabyte solution of adding an icon. But I would also let the loop implementation look a bit like Nightscout's implementation. Nightscout has a striketrhough and uses grey font color. For example it looks like:
image

ps2:
If you (or anyone else) rebases this change and submits before I get to it, I can take a look at it; this is a needed change.

diggabyte:
@ps2 are you fine with blanking it out? Or should we do an icon or text treatment to this?

PR #710, opened 2018-04-27
fix for face-id authentication on bolusing

Kdisimone:
adding to info.plist to allow for face ID authentication


ps2:
This issue and fix is pertinent if you are installing for the first time on a new phone, or a new instance of the app.  Installed Loop apps that already had FaceTime auth permissions aren't affected.

PR #712, opened 2018-04-30
Configurable bolus threshold

ddaniels1:
Pete this is my first pull request.  I am not sure I have configured this to work right with mmol though the code is essentially mimicked through all the files after suspend threshold. The basic premise here, and it works in testing for a month is that you can now set a separate bolus threshold that will allow a bolus between bolus and suspend threshold numbers where loop would otherwise still zero temp.  Any feedback appreciated. DM61 did the coding to set the feature up, my code was to make it configurable from loop instead of hardcoding a number as he had.


trixing:
It may actually be nicer to have a gradient instead of a threshold, ie.

do something like: units = units * min(currentGlucoseValue - suspendThreshold, 0) / (initialThreshold - suspendThreshold)

sethgagnon:
Any chance of this making its way into dev branch, or has the code changed so much since then that there are too many conflicts? Thanks

ps2:
I don't think this will be able to land in mainline Loop; having Loop suggest insulin when you're low is risky.  You can always override what Loop suggests.  In the future we may have a bolus interface that visually shows you what Loop thinks will happen with a specified dose amount, and you could use that as a guideline for how much to dose in these situations.

PR #715, opened 2018-05-04
Automatic Pump Time Synchronization

erikdi:
If an out of sync pump time is detected, automatically adjust the time and date of the pump, assuming that the mobile has a better idea of where we are. 

Potential corner case is traveling around the world, but for 99% of users this is probably what they want.  Given that the IOB calculation relies on accurate sync between the pump and mobile device, it seems prudent to make sure this is the case.

An alternative might be to refuse to loop if too big of a time difference is detected.


ps2:
Thanks @erikdi.  I do have a patch to do this in #721.  Traveling should not be an issue, as pump date is already normalized via the pump offset timezone object.

PR #730, opened 2018-05-27
Dynamic carb absorption once again... this time with graphs!

wannabeflatliner:
Hi,
I've been using loop for a week now. I've read a lot about the dynamic carb absorption, and what was available before the "dynamic" aspect was added to Loop.

I love the idea of dynamic carb absorption. It is very well thought.
What I would like to see added is a visual aid.

I don't have the right skills to do this by myself so I'm posting here to see what you guys think about it.
I've made some schemas so it's easier to understand. It would be a new way to input carbs (you could maybe toogle it on/off?)

[Schemas : https://www.dropbox.com/s/8p28gq9xz17uu4o/Absorption.pdf?dl=0](https://www.dropbox.com/s/8p28gq9xz17uu4o/Absorption.pdf?dl=0)

I think it could be a useful addition so it's easier for beginners to visualise what is happening so they can better adjust their bolusing methods.
Of course dynamic carb absorption would still be used, but based of the estimation of the graphs.

SH


ps2:
It sounds like this is a feature request, rather than a pull request.  Pull Requests are used for submitting actual code or documentation changes that you would like to get into the main release line. 

As for implementing this feature, I think it would fall lower on our priority list, as it requires a fair amount of user input, and carb input is already more work than we'd like.  However, if you want to keep  the idea out there in case someone else would like to implement it, please [open an Issue](https://github.com/LoopKit/Loop/issues/new) for it.

PR #738, opened 2018-06-14
Show carb absorption data for the past 24 hours

### LINKED ISSUE #681: Ability to see Glucose Change in Carbohydrates screen prior to current day.
Is there a way to adjust how far back the carbohydrates screen shows the carbs entered on the chart. When meals are still being absorbed overnight it would be nice to see how the carbs acted when I wake up in the morning. Currently, it looks like Loop resets the carb chart sometime in the middle of the night so I don't see how a meal I ate the prior evening was observed vs. predicted. This would be helpful to know in order to make adjustments to my meals. 
img_3047

bharat:
This resolves #681


ps2:
We still need to show a daily total.  Also, we only cache insulin data for 24 hours, so there can be inaccuracies in ICE data within DIA of the 24-hour limit.

bharat:
Good point. I can calculate and show the daily total, but before I do that- do you think that the inaccuracies are a blocker? Should I change it to 18 hours not 24 to avoid that problem?

PR #741, opened 2018-06-15
Add method to G5 CGMManager to make use of the backfilled glucose data.

erikdi:


ps2:
This is already in dev.

PR #743, opened 2018-06-16
Synthesize reservoir data on the simulator

bharat:
This enables the simulator to generated glucose predictions.

### DIFF
--- a/Loop/Managers/LoopDataManager.swift
+++ b/Loop/Managers/LoopDataManager.swift
@@ -854,10 +854,14 @@ final class LoopDataManager {
             throw LoopError.missingDataError(details: "Glucose", recovery: "Check your CGM data source")
         }
 
-        guard let pumpStatusDate = doseStore.lastReservoirValue?.startDate else {
-            self.predictedGlucose = nil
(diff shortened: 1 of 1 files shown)


ps2:
This does not actually synthesize reservoir data.  Are you using something else to get the data into HK on the simulator?

bharat:
The description of the commit itself could be better, sorry. I originally set out to synthesize whatever was necessary to get the simulator to do predictions. It turned out that all that was necessary was to have a pumpStatusDate so I left it at that. 

This is highly useful when you're working on the simulator.

ps2:
Any #if IOS_SIMULATOR defines mean that we have code that is untested when run on the simulator. In general testing should be done on a real device, at least for large features.

One thing you could do is take the database on disk from the LoopKit Example container and move it to Loop’s container, or it would be OK with a separate app target in the Loop project that generates data in various ways, since apps in the same app group container share a database anyway.

bharat:
ok, makes sense. let's close this out and I'll try a different approach. I've largely moved off of the simulator so my need for this is much less pressing.

PR #745, opened 2018-06-18
Update to match LoopKit-v2 api

bharat:


ps2:
Thanks @bharat, the breakage was due to expecting api changes in RLKit, which I decided to include in the LoopKit v2 PR.

PR #753, opened 2018-06-26
Implement PumpManager protocol

ps2:


Kdisimone:
There may have been some appreciative tears over here this weekend.  thank you for this.  Only thing I've noticed, for the basal rates and delivery rate screens, would be nice if any edits are saved when exiting the screen, even if "save to pump" is not selected.  Current operation is that making an edit and then backing out of the screen when done (similar to what current fingers would do by muscle memory) results in the edits not being saved (unless save to pump is used).  In situations where the "save to pump" fails because of comms issues, the edits also disappear (but that's probably an expected thing?  could it be prevented?).  thank you again!

ps2:
It's intentional that it doesn't save.  We don't want to get the two systems to get out of sync. A "track changes and warn before discard" would be a better experience.  Needs work.

Kdisimone:
Sounds good. I like that option too

PR #761, opened 2018-07-18
select the pickerValue based on 75% of recommendedBolus

jasoncalabrese:
previously was using 75% of the pickerValue
also remove the extra casting


jasoncalabrese:
good catch, cleaned it up
(I need to find the xcode settings to show whitespace)

ps2:
Editor -> Show Invisibles

PR #772, opened 2018-08-05
Display a glucose chart on the Watch

bharat:
This is a collaboration with Eric Jensen (@elnjensen). 

This change introduces a glucose chart in the Loop watch app. It preserves the existing interface but allows the user to swipe between two different experiences - the original one that is action-oriented and allows the user to add carbs and trigger a bolus, and a new one that shows the glucose chart.

The user can adjust the y-axis (blood glucose) using the crown, and the x-axis (hours of visible data) by using 3D touch on the chart. Here are a couple of videos to demo the experience:

Y-axis: https://www.youtube.com/watch?v=pD0UrJlqB5o 
X-axis: https://www.youtube.com/watch?v=mRudZmTTQyM&feature=youtu.be

Implementation notes:
- Load glucose data from GlucoseKit on the watch
- If there's a gap in data, backfill with data from the phone
- Refactor common interface support into HUDInterfaceController base class
- Pause the SpriteKit scene when there's no new data or running animations
- The size of the scene is hardcoded in the Storyboard and in the code

Open questions:
- See the note in ChartHUDController.awake - any thoughts on the root cause issue here?
- GlucoseChartScene relies on a bunch of new extensions to create more readable code. Most of these could be private to the file but would appreciate some guidance on whether these should move out into their own files
- Are the new image assets added correctly?
- Is there a better way to go about determining the size of the scene? It appears that you cannot get any details back from a WKInterfaceSKScene, would appreciate guidance here.

@elnjensen did I miss anything?


elnjensen:
One other small open question - [the code](https://github.com/bharat/Loop/blob/dev-watch-sk/WatchApp%20Extension/Controllers/ChartHUDController.swift#L20-L28) that implements the actions from the force-touch menu is quite redundant, but I can't see a way to make it simpler.  The menu items don't appear to be able to pass any information to the corresponding action function, e.g. there's no equivalent of the `tag` property of a `UIButton` that can let a single function know which button was pressed.  Not a big deal, but let us know if you see a better way to do this.

jlucasvt:
screen shot 2018-08-07 at 11 00 22 am

Not sure of the details just yet but I am getting this error when running the updated watch code and it kills the loop app when it's hit.  I am going to uninstall the watch app and reinstall to see if there is some bad previous data.

elnjensen:
@jlucasvt  You might be hitting that error when it's trying to convert a negative value to an unsigned integer.   Unlike real glucose values, predicted glucose can be negative...   Try changing both instances of `UInt16` in that file to `Int16` and see if that fixes the problem.

jlucasvt:
Makes sense Loop thinks Im on a crash course to -bg..  (Just had a Cannula Change and Needed to enter some fake carbs and a bolus to get the site wet..)  BG is dropping and loop predicts a fast and hard landing.. (i got glucose tabs if it gets rough)  

- That fixed it.  (good to check for those - values)

bharat:
ha! that never occurred to me. thank you for beta testing!! I just pushed Eric's suggested fix which should resolve this.

shanselman:
Is this in a good place where I could merge it into my own build? Is this getting merged anytime soon?

elnjensen:
@shanselman   You can merge from my branch [dev-watch-sk-ej](https://github.com/elnjensen/Loop/tree/dev-watch-sk-ej), which is based on Loop 1.9.1 but not yet updated for 1.9.2 changes.   No idea about timescale on this PR.

ps2:
Accepting this, and adding some tweaks as #824

Issue #705, opened 2018-04-25
Enhancement: interpret and plot reservoir changes above max basal rate as bolus triangles


elnjensen:
Under some circumstances, Loop will display basal rate spikes on the "Insulin Delivery" graph that have very large values, much greater than the max basal rate allowed by the pump.  Those of us who have been using Loop for some time know that these are actually boluses, and that Loop will update the graph eventually when it reads the full pump history.  

However, these graphs continue to confuse new users, and to cause some consternation, since they lead people to think that Loop has actually somehow delivered huge amounts of insulin somehow.  Thus, avoiding them if possible would be less confusing to users. 

Since we know the max basal rate, in theory it should be possible to identify such events and to plot them more or less correctly.  I say "more or less" because it's probably not completely possible to separate out what the actual basal rate was at the time, i.e. how much of the very large "basal" rate during that window was really basal, and how much was bolus.  However, we *can* say with confidence that any delivery above the allowed max basal rate surely included a bolus component. 

So my proposal would be to flag these events while parsing the dose entries, and plot them as boluses.  The parsing is [here](https://github.com/LoopKit/Loop/blob/master/Loop/Managers/StatusChartsManager%2BLoopKit.swift#L86), and right now the logic is "if entry is labeled as bolus, plot as bolus, otherwise plot as basal."  We could add another check there:  "if entry is labeled as bolus, plot as bolus; if entry has a delivery rate greater than max basal, plot as bolus (possibly with a different symbol to indicate it was estimated - open triangle?); otherwise, plot as basal." 

I'd be willing to tackle this, but I'm curious to hear comments from others.  Any downsides to this?  Or things to think about if I implement it?   One thing that would be useful to know is under what circumstances this effect appears.  I would think it's when Loop is relying on reservoir data that are newer than the latest event history data, but I don't know exactly when that happens.

diggabyte:
So this calculated "transient" bolus could have a margin of error ranging from zero to max basal units, correct? And therefore, it could be _slightly_ inaccurate until it has been reconciled from events. But as long as this transient bolus is only used for display purposes in the graph and never in any looping logic, that margin of error isn't much of a concern. 

The idea of a open / hollow triangle is an excellent way to communicate it has not been fully reconciled.

elnjensen:
Yes, that's correct - it may not be exactly right because we don't know the temp basal at the time.  But it would be closer the actual delivery situation than the current graph, and would only be temporary.

PR #777, opened 2018-08-11
Reconcile delivery above max basal rate as bolus on chart

elnjensen:
This addresses issue #705 by interpreting delivery rate above the max basal rate (if it is set) as a bolus, and graphing it as such.  One part of the rendering that isn't perfectly accurate is that a bolus that is in progress at the time of a reservoir reading (i.e. whose delivery is partly reflected in two consecutive reservoir readings since it spans them) will initially be registered as two smaller boluses (see first screenshot below).  Eventually this will be rendered correctly as a single, larger bolus once event history is read, as shown in the second screenshot below.  In my view this initial inaccuracy (two boluses instead of one) is less confusing / alarming for a user than the current initial inaccuracy of showing it as an extremely high basal rate.  

Initial rendering based on high delivery rate in reservoir data: 
img_2873

Later rendering after event history is read: 
img_2874

### DIFF
--- a/Loop/Managers/StatusChartsManager+LoopKit.swift
+++ b/Loop/Managers/StatusChartsManager+LoopKit.swift
@@ -72,7 +72,7 @@ extension StatusChartsManager {
         }
     }
 
-    func setDoseEntries(_ doseEntries: [DoseEntry]) {
+    func setDoseEntries(_ doseEntries: [DoseEntry], _ maxBasalRateUnitsPerHour: Double?) {
         let dateFormatter = self.dateFormatter
         let doseFormatter = self.doseFormatter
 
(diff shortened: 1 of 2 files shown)


elnjensen:
I should add that I’d be happy to add code to avoid the “double bolus” issue shown above. I opted for a pretty minimal approach here, but it wouldn’t be hard to address that issue with a bit more code, or also to go a step further and flag tentative boluses with a different symbol on the graph.

dm61:
@elnjensen I've merged this to my Loop, thanks for adding this feature. It's working well and is serving the purpose of eliminating confusion around temporary ultra-high basal rates. I do not think addressing the double-bolus situation would be necessary. Another "inaccuracy" is that a small-enough bolus still shows up temporarily as high-temp basal, but I do not think that would confuse or disturb anyone. OIn the other hand, it would probably be good to have a different symbol for tentative boluses (a hollow triangle perhaps?).

elnjensen:
@dm61   Adding a different symbol/color would definitely be possible, but it would add some additional complexity to the code.  Thought I'd wait and see if the overall change had any appeal first, and then I could implement that feature if people think it's worthwhile.

ps2:
The long term plan here is to switch away from having LoopKit track reservoir, and for the minimed pump manager to reconcile pending doses using history and reservoir as appropriate. It should be able to prevent this situation from happening, by canceling out reservoir doses that could be accounted for by in-progress boluses. In progress boluses that originated from the pump could show as high rate temps, or it could assume after a certain rate that they were boluses that hadn't shown up in history yet.  Either way, the code needs to be in minimed pump manager.

PR #778, opened 2018-08-11
From DM61's work on IRC, fixes a case where RC velocity can have an a…

ps2:
…rtificially large magnitude from short glucose change intervals.


ps2:
IRC is likely going to end up in Loop, but there are some structural changes to be made first, to better handle recomputation when needed.

diggabyte:
`s/descrepancy/discrepancy/` :)
Otherwise, this looks good 👍

PR #780, opened 2018-08-14
Fix x15 bolus 

noahlisser:


Kdisimone:
this looks like an errant PR.  If you really would like to submit a PR for an x15 fix, it would have to include LoopKit changes and be based on dev branch.  Thanks...i'll close and you can resubmit to the right branch/repo if that was your intent.

PR #787, opened 2018-08-24
Duplicate reference file reference deleted && Common folder path fixed

Kdisimone:


Kdisimone:
Looks like these were already addressed.  Thanks

PR #805, opened 2018-09-13
Improving german translations

InsulinJunkieDe:
I've improved couple of German translations


Loop/de.lproj/Localizable.strings
```diff
@@ -192,7 +192,7 @@
 "Fiasp" = "Fiasp";
 
 /* The format string used to describe a finite workout targets duration */
-"For %1$@" = "Für %1$@";
+"For %1$@" = "für %1$@";
```
ps2: Why is this not capitalized?

Issue #816, opened 2018-09-21
RFC: Apple Watch complication update algorithm


mpangburn:
## Background

One of the primary technical challenges in the development of an Apple Watch complication for a diabetes app is the timely transmission of data from the phone to the watch: an app has only 50 guaranteed immediate complication updates per day. With a new sensor glucose data point incoming every five minutes but only enough guaranteed updates for one every ~30 minutes, keeping the complication up-to-date presents a meaningful challenge.

## Present State

Currently, after a loop is completed, the decision to use one of the budgeted complication updates boils down to the fulfillment of one of the following requirements:

- Sufficient time (30 minutes) has elapsed since the last update.
- The sensor glucose value has changed significantly (20 mg/dL) since the last update.

If neither of these conditions are met, Loop requests an update to the watch via application context. The underlying framework determines an appropriate time to transfer the information based on factors such as communication strength and battery life; this transfer of data is opportunistic rather than deterministic.

Loop accommodates the restrictions on complication updates by including the time elapsed since the recorded glucose in complications providing enough space for the additional text. This information dwindles in utility over time nonetheless, often leading users to necessarily tap the complication and open the main app in order to view the most recent glucose value.

## Proposed Changes

While Loop remains restricted by the confines of the existing complication update system provided by Apple, we can improve the algorithm used to update the complication to enhance the timeliness of the information presented.

I propose the following:

**1.** The time threshold for using an immediate complication update, rather than being fixed at 30 minutes, should be determined dynamically based on the number of remaining complication updates for that day:

`(affordable update rate) = (remaining immediate updates) / (time until budget resets)`

This simple adjustment has the following effects:

- If the watch is disconnected from the phone for any period of time throughout the day, the update rate increases appropriately.
- If the second immediate update condition (sufficient change in glucose value) holds when sufficient time has not yet passed, the update rate will decrease to accommodate the remainder of the day accordingly.

**2.** The complication update algorithm should favor the user's waking hours.

People sleep. Even if a user elects to wear their watch while sleeping, expending immediate updates during sleeping hours is an inefficient use; the amount of time spent glancing at the watch face overnight is substantially lower than during the day. To make better use of the budget, we can bias the distribution of immediate updates in favor of the user's waking hours.

[...]

PR #825, opened 2018-10-05
Watch Feature Updates

ps2:
- Update IOB and COB immediately after carb/bolus send
- Adjust complication push budget to spend evenly before midnight
- Adjustments for Series 4, including rounded action buttons, correct text margin insets, and using system scroll haptics
- Improve efficiency of the chart redraw

### DIFF
--- a/Loop/Managers/WatchDataManager.swift
+++ b/Loop/Managers/WatchDataManager.swift
@@ -42,33 +42,18 @@ final class WatchDataManager: NSObject {
     @objc private func updateWatch(_ notification: Notification) {
         guard
             let rawUpdateContext = notification.userInfo?[LoopDataManager.LoopUpdateContextKey] as? LoopDataManager.LoopUpdateContext.RawValue,
-            let updateContext = LoopDataManager.LoopUpdateContext(rawValue: rawUpdateContext),
-            let session = watchSession
+            let updateContext = LoopDataManager.LoopUpdateContext(rawValue: rawUpdateContext)
         else {
             return
         }
 
         switch updateContext {
-        case .glucose:
-            break
-        case .tempBasal:
-            break
+        case .glucose, .tempBasal:
+            sendWatchContextIfNeeded()
         case .preferences:
             sendSettingsIfNeeded()
-            return
         default:
-            return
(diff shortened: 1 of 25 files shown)


Loop/Managers/WatchDataManager.swift
```diff
@@ -301,3 +306,57 @@ extension WatchDataManager: WCSessionDelegate {
...
+            "* outstandingFileTransfers: \(outstandingFileTransfers)",
+            "* outstandingUserInfoTransfers: \(outstandingUserInfoTransfers)",
+            "* receivedApplicationContext: \(receivedApplicationContext)",
+            "* remainingComplicationUserInfoTransfers: \(remainingComplicationUserInfoTransfers)",
+            "* complicationUserInfoTransferInterval: \(round(complicationUserInfoTransferInterval.minutes)) min",
+            "* watchDirectoryURL: \(watchDirectoryURL?.absoluteString ?? "nil")",
+        ].joined(separator: "\n")
+    }
+
+    fileprivate var complicationUserInfoTransferInterval: TimeInterval {
```
mpangburn: As I've been digging into this over the last week or two, I've found that midnight isn't the right estimate for the budget reset time—@elnjensen and I have been recording 5AM budget resets, but I don't think our two experiences are sufficient evidence to draw make a general conclusion.
I'm continuing to look into this, but we'd benefit from more logging here.

ps2: You have been recording additional resets at 5am (so a midnight *and* a 5am reset?) or no midnight reset, and just the 5am one?

elnjensen: It's 5:02 AM every day, but not midnight, at least for the two of us who have been tracking it - see #816 .   What we've observed is a daily reset at 5 AM, and then an additional reset (which doesn't change the daily schedule) whenever the Watch is fully charged.

ps2: Ok.  The midnight target isn't an estimate of when the counter will reset, but an estimate of when the user will be asleep.  Still open to better algorithms for divvying this resource up, though.

PR #828, opened 2018-10-08
Add Spike support to Loop

dabear:
This adds in my spike-client-swift which is a modified version of dexcom-share-client-swift. It allows loop to connect to the spike app running locally on port 1979 on the same device

image

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -6,3 +6,4 @@ github "LoopKit/dexcom-share-client-swift" == 1.0
 github "LoopKit/G4ShareSpy" == 1.0
 github "ps2/rileylink_ios" == 2.1.0
 github "LoopKit/Amplitude-iOS" "decreepify"
+github "dabear/spike-client-swift" "master"
--- a/Cartfile
+++ b/Cartfile
@@ -6,4 +6,4 @@ github "LoopKit/dexcom-share-client-swift" == 1.0
 github "LoopKit/G4ShareSpy" == 1.0
 github "ps2/rileylink_ios" == 2.1.0
 github "LoopKit/Amplitude-iOS" "decreepify"
-github "dabear/spike-client-swift" "master"
+github "dabear/spike-client-swift" == 0.9.1
--- a/Cartfile
+++ b/Cartfile
@@ -6,4 +6,4 @@ github "LoopKit/dexcom-share-client-swift" == 1.0
 github "LoopKit/G4ShareSpy" == 1.0
 github "ps2/rileylink_ios" == 2.1.0
 github "LoopKit/Amplitude-iOS" "decreepify"
-github "dabear/spike-client-swift" == 0.9.1
+github "dabear/spike-client-swift" == 0.9.2
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -3,6 +3,6 @@ github "LoopKit/CGMBLEKit" "v3.0"
 github "LoopKit/G4ShareSpy" "v1.0"
 github "LoopKit/LoopKit" "v2.2.1"
 github "LoopKit/dexcom-share-client-swift" "v1.0"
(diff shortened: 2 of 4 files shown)


francesc0-cgm:
Great! Thank you @dabear . Hope it will be merged very soon!

dabear:
@Kdisimone are you able to test this?

ps2:
@dabear I don't use Spike, so I am not able to test this. Integrating this into Loop requires that someone is committed to running this with Loop dev as it is updated, responding to users who file issues with the spike client manager, fixing any problems, and keeping the Cartfile of spike-client-swift pointed to a compatible version of LoopKit;

Specifically, for Loop to build with spike-client-swift, the Cartfiles of both need to resolve to the same rev of LoopKit.  If we can't find a way to keep this updated in a timely manner, then I would have to remove this integration.

I noticed that the spike-client-swift code is not setting an HKDevice for the stored CGM data. That would need to be set so we can identify glucose values stored in HealthKit that came from Spike. Ideally, Spike would know the device (Dexcom, Libre, etc.) that it is fetching data from, and would include that in the response, so you could create a more meaningful HKDevice.

dabear:
This is pending some extra metadata that needs to be added to Spike's endpoint and tested with spike-client.

dabear:
This is pending some extra metadata that needs to be added to Spike's endpoint and tested with spike-client.

dabear:
This is pending some extra metadata that needs to be added to Spike's endpoint and tested with spike-client.

dabear:
This is pending some extra metadata that needs to be added to Spike's endpoint and tested with spike-client.

Hedgey72:
Wich release are you using? I can't find the allCGMManagers file just the overall CGM file.

dabear:
@Hedgey72 this is for the dev branch. 

I am still waiting for the requested changes to be done in Spike. It seems it will take some time.

francescaneri:
@ps2 @Kdisimone what do you think about the integration of spike? It's possible?

sad1121:
Hi!
Can you provide step-by-step guide for non-programmers loop user for spike integration due to stopped spike-master branch.
It would be very helpful for community.
🙏

dabear:
1.  git clone on the spikeclient, which you can find here: https://github.com/dabear/spike-client-swift/tree/master/SpikeClient
2. Update cartfile dependencies to point to same dependecy versions as loop master uses.
3. Build the the spikeclient in xcode to make sure there aren't any code edits to the spikeclient necessary. Resolve any build failures.
3.  push those changes to a new github repo you create
4.  take a fresh copy of loop master, clone it and add a reference to the spikeclient you just pushed into the carthage file: github "yourusernameingithub/spike-client-swift" "master"
5.  do a carthage update
6. In xcode, add a reference to SpikeClient and SpikeClientUI.
7. In xcode, go to the build targets you want, for example Loop, and add the following input paths to the "Copy Frameworks with Carthage" section:
```
"$(SRCROOT)/Carthage/Build/iOS/SpikeClientUI.framework",
				"$(SRCROOT)/Carthage/Build/iOS/SpikeClient.framework",
```
8. Build Loop, and a "spike" should now appear in the dropdown list when choosing a cgmsource
9. Publish your fork of loop somewhere, for example on github, for others to use.

Kdisimone:
And to add to the instructions above, please make sure there's step 8b

8b. Update the info.plist files in Loop folders so that the version naming reflects which version of loop was used as the base and what edit has been done.  For example...

https://www.facebook.com/groups/TheLoopedGroup/permalink/2234943270055660/

dabear:
Due to the way spike's distribution model has changed the recent couple of days, I no longer feel it is safe to use spike for looping, and as a consequence I will close this pull request. The code is still available in my github repo for those who want to take the risk them selves.

PR #830, opened 2018-10-10
Tweak colors of ranges; add menu for set workout durations; increase font size of graph labels

elnjensen:
This PR includes a few small changes to the graph to make it a bit more legible.  The main code addition is a force-touch menu on the initial Watch screen (with the carbs, bolus, pre-meal, and workout buttons) to allow setting Workout mode for three finite intervals, complementing the indefinite Workout mode setting from the main button.


WatchApp Extension/Controllers/ActionHUDController.swift
```diff
@@ -81,6 +81,28 @@ final class ActionHUDController: HUDInterfaceController {
...
+        }
+        let endDate = Date().addingTimeInterval(TimeInterval(hours: duration))
+        guard glucoseTargetRangeSchedule.setOverride(.workout, until: endDate) else {
+            return
+        }
+        sendGlucoseRangeSchedule(glucoseTargetRangeSchedule)
+    }
+    
+    // Add force-touch menu items to set workout mode for certain fixed durations:
+    @IBAction func setWorkout30Min() {
```
ps2: Should use a a long-press gesture on the workout mode button to provide options for workout mode duration, that would present an action sheet (WKAlertControllerStyleActionSheet) with the time options.

WatchApp Extension/Scenes/GlucoseChartScene.swift
```diff
@@ -334,7 +339,7 @@ class GlucoseChartScene: SKScene {
 
         correctionRange?.quantityBetween(start: start, end: end).forEach({ (range) in
             let (sprite, created) = getSprite(forHash: range.chartHashValue)
-            sprite.color = UIColor.glucose.withAlphaComponent(activeOverride != nil ? 0.2 : 0.3)
+            sprite.color = UIColor.glucose.withAlphaComponent(activeOverride != nil ? 0.3 : 0.5)
```
ps2: These color ranges are too bright, and drown out the glucose value points.

WatchApp Extension/Scenes/GlucoseChartScene.swift
```diff
@@ -28,7 +28,7 @@ private enum NodePlane: Int {
 private extension SKLabelNode {
     static func basic(at position: CGPoint) -> SKLabelNode {
         let basic = SKLabelNode(text: "--")
-        basic.fontSize = 12
+        basic.fontSize = 15
```
ps2: This is too big for 38mm

ps2: instead of hard-coding the font size, use the system-provided size for .caption2
```
UIFont.preferredFont(forTextStyle: .caption2).pointSize
```

WatchApp Extension/Controllers/ActionHUDController.swift
elnjensen: Long-press (force touch) is sort of a one-trick pony in terms of interface.   It can only apply to the whole screen (can't be associated with a button), and the only thing it can do is to present a context menu with simple buttons.  So I think there are two choices:  keep the code as-is (with the admittedly odd combinations of two different ways to set workout mode), or change the function of the Workout button to present the action sheet, including both the finite time and indefinite time options.  The latter is definitely more consistent with the main Loop app interface.  The only thing it loses is the ability to set Workout mode indefinitely with one press - it would now require two presses.  That's probably a worthwhile tradeoff for a cleaner interface.  What do you think?

elnjensen: Ah, ok!  Thanks @mpangburn .  I'll take a look.

WatchApp Extension/Scenes/GlucoseChartScene.swift
elnjensen: I agree that the color of the target ranges starts to look pretty similar to the color of the glucose points as the alpha value gets larger.  However, I don't think the very low alphas here work well overall.  At 0.2 or 0.3, I find the target ranges very hard to see, especially for narrow ranges.  With the dark background, anything with low alpha looks pretty dark as well, and I think the model that works well on the iPhone app (same color with different alphas) works less well with the small screen and dark background here.  [Calculating contrast ratios](https://gist.github.com/elnjensen/e7b649e83b6213f9fbe8e503d59f98a8) of the blended color (taking the alpha into account and blending with the background), alphas of 0.2 and 0.3 give contrast ratios of 1.5 and 1.8.  Apple's [Human Interface Guidelines](https://developer.apple.com/design/human-interface-guidelines/watchos/visual-design/color/) recommend a minimum of 4.5.   Now, we don't have to hit that for every element, but it does suggest how far below that we are.

A different approach to increasing the contrast but still maintaining a visual distinction between elements is to use a slightly different color for the ranges, but keep the alpha up.  That way the contrast is there, and the visual connection between similar elements is maintained by not having the colors *too* different, but they compete with each other less.  In the original PR we used `UIColor(red: 158/255, green: 215/255, blue: 245/255)` with alphas of 0.4 and 0.6 for disabled and active ranges, respectively.  Combined with the background we're currently using, those have contrast of 2.5 and 3.2, still below 4.5 but noticeably more than the values above.  

This is what that color scheme looks like: 
img_3028

And this is the color scheme from the code that's currently in dev (glucose color with alpha=0.2/0.3): 
img_3032

I'm not wedded to that exact color scheme, but I do think more contrast would be helpful here.

ps2: The contrast that is important here is between the chart points and the chart ranges.  Both of your suggestions reduce the difference between them.

elnjensen: Clearly we don't agree on this, which is fine.  I'll revert the changes and we can see how those colors work in practice.

elnjensen:
@ps2  I've made the requested changes.  Let me know if you see anything else.  Thanks!

WatchApp Extension/Scenes/GlucoseChartScene.swift
```diff
@@ -235,10 +238,12 @@ class GlucoseChartScene: SKScene {
         addChild(now)
 
         hoursLabel = SKLabelNode.basic(at: CGPoint(x: textInsets.left, y: size.height - textInsets.top))
+        hoursLabel.verticalAlignmentMode = .top
```
ps2: This is already set to .top in the basic() initializer above.

elnjensen: Hmm, I see that now.  That label was failing to align correctly when I was trying alternate scene sizes, and this fixed it.  I'll see if I can reproduce that later, but probably can revert this.

elnjensen: Can't reproduce what I saw, so I've reverted those changes.

ps2:
This is obsolete with overrides.

PR #836, opened 2018-10-16
Move data display for IOB, COB, etc to a third page

bharat:
This allows us to have a larger, cleaner graph page. 
Additionally, remember and restore the graph visible duration.


bharat:
Here's what the 3 pages look like:
Action: img_1871
Chart: img_1872
Data: img_1873

WatchApp Extension/Scenes/GlucoseChartScene.swift
```diff
@@ -182,9 +182,10 @@ class GlucoseChartScene: SKScene {
         return lowerBound..<upperBound
     }
 
-    var visibleDuration = TimeInterval(hours: 6) {
+    var visibleDuration = UserDefaults.standard.visibleDuration {
         didSet {
             setNeedsUpdate()
+            UserDefaults.standard.visibleDuration = visibleDuration
```
elnjensen: Nice idea to save this and restore each time!

WatchApp/Base.lproj/Interface.storyboard
```diff
@@ -468,6 +440,88 @@
...
+                                        </group>
+                                    </items>
+                                    <variation key="device=watch40mm" widthAdjustment="-8.5"/>
+                                    <variation key="device=watch44mm" widthAdjustment="-9.5"/>
+                                </group>
+                            </items>
+                        </group>
+                        <group width="1" alignment="left" layout="vertical" spacing="0.0" id="seG-UI-feg">
+                            <items>
+                                <label alignment="left" text="INSULIN ON BOARD" id="kvt-sg-hNd">
```
elnjensen: Using "Active Insulin" (and below, "Active Carbs") would be more consistent with the iPhone app - plus "on board" is a bit of an odd English idiom that might be a little harder to translate English language learners if they're not running a localized version in another language.

elnjensen:
Just deployed to my 40mm watch.  The number of carbs is truncated a bit by the rounded corners of the display, so the overall height needs to be a bit shorter.   (Not sure why the image comes out rotated here - it's not that way on my computer - but you can still see what I mean.)  Interestingly, this truncation doesn't show up in a screenshot, so it may not in the simulator, either.   Keeping the bottom of the label above the page indicator dots would keep things in safe territory, I think. 

img_3040

ps2:
The main graph is looking good. For the additional info, I think we should look to first party apps as examples of good design. I've mentioned the activity app, but the weather app is also a good example. The current platters look like buttons, and should have a horizontal separator instead.

As for 3rd page instead of scrolling, scrolling is a better navigation method in this context. The problem with pages is that they get harder to use when you add more. We want the action screen to be quickly accessible.

ps2:
Closing in favor of #851

Issue #803, opened 2018-09-13
Apple Watch Series 4 complication support


mpangburn:
Apple Watch Series 4 comes with two exclusive new Watch faces, Infograph and Infograph Modular. Accompanying these Watch faces are four new complication families for Loop to support:
- Graphic Corner (`CLKComplicationFamily.graphicCorner`)
- Graphic Circular (`CLKComplicationFamily.graphicCircular`)
- Graphic Bezel (`CLKComplicationFamily.graphicBezel`)
- Graphic Rectangular (`CLKComplicationFamily.graphicRectangular`)

More information and sample complications [here](https://developer.apple.com/documentation/clockkit/graphic).

Without having yet experimented with the API, here are some initial thoughts, by complication:
**Graphic Corner**
- `CLKComplicationTemplateGraphicCornerStackText` with the two standard text providers, glucose + trend and time elapsed since update. There might be space for an additional piece of text here, too.

**Graphic Circular**
- `CLKComplicationTemplateGraphicCircularClosedGaugeText` with the closed gauge kept full to mimic the Loop icon, adjusting color to match the closed loop state, and the center text using the current glucose (and ideally trend, space permitting). 
- `CLKComplicationTemplateGraphicCircularOpenGaugeRangeText` with progress around the open ring relative to the current glucose, with the leading and trailing text providers denoting the range bounds (i.e. mimic the example weather complication, but using blood glucose concentration rather than temperature). This design has potential for diabetes apps but comes with a couple challenges for Loop in particular:
    - Loop has no color association for high or low glucose values; since the app uses solid blue for glucose points in all cases, using a gradient of new colors would introduce design complexity.
    - The appropriate bounds for the gauge are not immediately apparent. Making the bounds user-configurable introduces undesirable setup complexity, but making the bounds dynamic makes the complication more difficult to digest at a glance. The best we can do here is pick a reasonable range (say 50...300 mg/dL) and keep it fixed.

**Graphic Bezel**
- `CLKComplicationTemplateGraphicBezelCircularText` with the circular design from above and the additional text used for the time elapsed since the last update. It looks like the text can wrap pretty far around the bezel, in which case we can fit additional pieces of information here (IOB? COB?).

**Graphic Rectangular**
[...]

PR #848, opened 2018-11-03
Apple Watch Series 4 complications

mpangburn:
WIP, but could use some guidance.
Some relevant discussion in #803.

Summary:
- Abstraction of model layer used in existing chart.
- CoreGraphics rendering of chart for use in Graphic Rectangular complication.
- Simple complication implementations for the other Graphic complications.

My main challenge is with the Graphic Rectangular complication image rendering. I've gotten the glucose points to render to scale properly, but the prediction path and labels still have pixel alignment issues.


mpangburn:
To keep the graph updated, it looks like incoming glucose values should be cached. This change will probably involve queueing user info transfers to ensure all glucose values are delivered (rather than requesting a backfill when the in-app chart is displayed).

Still some work to do here, but I'll leave this open in case anyone wants to try it out or has design suggestions for the complications.

mpangburn:
Having trouble tracking down this build failure. No issue locally. Build script output in the Travis log is unhelpful; unclear where it's segfaulting.

WatchApp Extension/ComplicationController.swift
```diff
@@ -67,28 +103,107 @@ final class ComplicationController: NSObject, CLKComplicationDataSource {
...
-        } else {
-            entries = nil
+            handler(entries)
         }
+    }
+
+    // MARK: - Placeholder Templates
 
-        handler(entries)
+    func getLocalizableSampleTemplate(for complication: CLKComplication, withHandler handler: @escaping (CLKComplicationTemplate?) -> Void) {
```
ps2: ## Complication Placeholders
The header comment for `getLocalizableSampleTemplate(for family:)`:

> These methods will be called once per supported complication when your extension is installed, and the results cached.

Using `CLKRelativeDateTextProvider` in this method means the values get to be crazy, e.g. "-25 days".
You could use a simple text provider to fake a date string, or just not bother with this and let the Apple Watch app render a default (using app name/icon).

WatchApp Extension/Extensions/CLKTextProvider+Compound.m
```diff
@@ -0,0 +1,33 @@
...
+//  Copyright © 2018 LoopKit Authors. All rights reserved.
+//
+
+#import "CLKTextProvider+Compound.h"
+
+// CLKTextProvider.textProviderWithFormat (compound text provider creation) is unavailable in Swift.
+// c.f. https://crunchybagel.com/using-multicolour-clktextprovider-in-swift-in-watchos-5/
+@implementation CLKTextProvider (Compound)
+
++ (CLKTextProvider *)textProviderByJoiningTextProviders: (nonnull NSArray<CLKTextProvider *> *)textProviders separator:(nullable NSString *) separator {
```
ps2: Use NS_ASSUME_NONNULL_BEGIN / NS_ASSUME_NONNULL_END at the top and bottom of each file to streamline nullability annotations.

WatchApp Extension/Extensions/CLKTextProvider+Compound.h
```diff
@@ -0,0 +1,20 @@
+//
+//  CLKTextProvider+Compound.h
+//  Loop
+//
+//  Created by Michael Pangburn on 10/27/18.
+//  Copyright © 2018 LoopKit Authors. All rights reserved.
+//
+
+#ifndef CLKTextProvider_Compound_h
```
ps2: This can be removed

WatchApp Extension/Managers/ComplicationChartManager.swift
```diff
@@ -0,0 +1,130 @@
...
+//  Copyright © 2018 LoopKit Authors. All rights reserved.
+//
+
+import Foundation
+import UIKit
+import HealthKit
+import WatchKit
+
+
+private enum ComplicationChartConstants {
```
ps2: The empty enum namespace only makes the access longer. Prefer either private extensions on the type (so the type name can be omitted when used) or just use top-level private constants instead.

WatchApp Extension/Managers/ComplicationChartManager.swift
```diff
@@ -0,0 +1,130 @@
...
+        return data?.unit ?? .milligramsPerDeciliter
+    }
+
+    func renderChartImage(size: CGSize) -> UIImage? {
+        UIGraphicsBeginImageContextWithOptions(size, false, 0.0)
+        defer { UIGraphicsEndImageContext() }
+
+        let context = UIGraphicsGetCurrentContext()!
+        drawChart(in: context, size: size)
+        let image = context.makeImage().map(UIImage.init(cgImage:))
```
ps2: This is unreadable... it's OK to take extra lines of code if it means future readers can follow. Use a `guard` to unwrap the CGImage.

WatchApp Extension/Managers/ComplicationChartManager.swift
```diff
@@ -0,0 +1,130 @@
...
+        renderedChartImage = image
+        return image
+    }
+
+    private func drawChart(in context: CGContext, size: CGSize) {
+        guard let data = data else {
+            // TODO: handle empty data case
```
ps2: If data is nil, `renderChartImage` should return nil.

WatchApp Extension/Models/GlucoseChartData.swift
```diff
@@ -0,0 +1,101 @@
...
+    init(unit: HKUnit?, correctionRange: GlucoseRangeSchedule?, historicalGlucose: [SampleValue]?, predictedGlucose: [SampleValue]?) {
+        self.unit = unit
+        self.correctionRange = correctionRange
+        self.historicalGlucose = historicalGlucose
+        self.historicalGlucoseRange = historicalGlucose?.quantityRange
+        self.predictedGlucose = predictedGlucose
+        self.predictedGlucoseRange = predictedGlucose?.quantityRange
+    }
+
+    func chartableGlucoseRange(from interval: DateInterval) -> Range<HKQuantity> {
```
ps2: shouldn't go below 0 (this is currently in dev)

WatchApp Extension/ComplicationController.swift
```diff
@@ -40,25 +40,61 @@ final class ComplicationController: NSObject, CLKComplicationDataSource {
...
+            else {
+                completion()
+                return
+            }
+
+            ExtensionDelegate.shared().loopManager.generateChartData { chartData in
+                self.chartManager.data = chartData
+                completion()
+            }
         }
```
ps2: doesn't call the completion handler on watchOS 4

WatchApp Extension/Extensions/CLKTextProvider+Compound.m
```diff
@@ -0,0 +1,33 @@
...
+
+    NSString *formatString = @"%@%@";
+
+    if (separator.length > 0) {
+        formatString = [NSString stringWithFormat:@"%@%@%@", @"%@", separator, @"%@"];
+    }
+
+    CLKTextProvider *firstItem = textProviders.firstObject;
+
+    for (int index = 1; index < textProviders.count; index++) {
```
ps2: Use `NSInteger` instead of `int`

PR #849, opened 2018-11-04
Fpu dev

rsilvers129:
rsilvers FPU modifications.

Adds in support for fat and protein.

See Wiki: 

https://github.com/rsilvers129/Loop/wiki/Fat-Protein-Unit-Branch-(FPU-dev)-of-Loop!-With-IRC-and-Spike.

See video:

https://www.youtube.com/watch?v=OvpRUlD0acU


outerim:
@rsilvers129 have you been continuing to work on this? We’ve been using it for some months now with good success. Wanting to update to the latest dev but the merge conflicts here look pretty intimidating :)

mitchellhenke:
I’ve done my best to bring the changes here with FPU and IRC (#726) in line with dev on my branch here: https://github.com/mitchellhenke/Loop/tree/mitchell-rsilvers129-fpu-irc

ps2:
The direction Loop needs to go is towards requiring *less* of the user in terms of input. I think an updated carb absorption algorithm that accounts for fat & protein based on the three main choices (lollipop, taco, pizza) or something similar, would be acceptable. Adding something like this should *replace* duration. Finally, the algorithm updates should be integrated into the carb absorption model in a thorough way; not 'hacked' in, like using the FPU terms to modify number of carbs or duration terms.

dyjaks:
Hey Pete, in that line of thinking would we still be having users input "fake" carbs for protein (as carbs) in a more simple carbs + protein/2 or whatever ratio they determine?

ps2:
Yeah, I'm not sure if there is a way to have users enter a meal simply and still account for fat & protein separately. That's why this interface is tough.  Many users already just swag their carb estimates.  I would need to look more into the modeling that Robert's work is based on, but if, by selecting a meal size of some sort, and something that indicates roughly the proportion of fat, protein, and carbs (like lollipop, taco, pizza), it's possible that the model works out better, and is still convenient enough to use many times a day.

Jacob2877:
Interesting. I guess my point for the new carb model in future releases is to factor in meals w/low carbs but lots of protein and fat and there will be a spike hours later, not sure if high temping will always work. Or, if carb entries are required like fpu does right now? Do you see the new direction addressing this? 

Think ribeye steak, some pork belly, and avocado boat

dyjaks:
In my experience a high temp can deal with protein just fine. All FPU really does as far as Loop is concerned can be though of as:
Carbs with 3 hour absorption now  and then based on size do:
"small" extend bolus over 3 hours delayed 1 hours
"medium" 4 hours delayed 1 hours
"large" 5 hours delayed 1 hours
"extra large" 6 hours delayed 1 hours
  
https://youngandt1.com/how-to-bolus-for-fat-and-protein/  
  
Is the most friendly way to read it. So for a steak example (where we'd choose the steak icon) and maybe put in what we assume to be our fake carb amount that I eyeball as 60g protein or 30g carbs we would consider that as "large" and then absorb the 60g over the next 6 hours, assuming it starts in 1 hour.

The real trick here is fat is the bigger player in absorbtion time because FPU is based on calories and fat is 9cals a gram and protein is only 4cals a gram. I think, if we want user friendly, we magic away the fat and just have that known from icons or size of meal.

Jacob2877:
Ok but if you are only eating steak and Fat for your meal and you plug 30-35 grams of carbs Loop will still give you an upfront bolus. In my experimentation I do not need an upfront bolus, the rise is usually 3-4 hours after eating. This is where I have seen the fpu be more of an automatic, less thinking tool. I suppose I could push the 30 carbs to start 2 hours later but again just not sure. I guess we need some real world testing to really prove the point.

dyjaks:
Well how about this, three checkboxes (there's got to be a prettier way) for carb, protein, and fat. Check the ones that are relevant to your meal and that would dictate what FPU considers the delay factor. If you don't check carb you don't need anything up front. The actual math isn't something I can speak to how we make it work right but I do think the point that having people input three macros correct isn't user friendly when most already eyeball the first.

gwelda:
Agree with Jacob2877.
Having a teenager, and for better control, we try to eat more low carb. But protein and fat is high. I always need to put fake carbs and keep the bg in check for more than 4 hours after the meal, for the later spike. With FPU, it helps, its like planning ahead the spike.There is less effort.

Jacob2877:
That would a great idea actually. Love it. 

> Well how about this, three checkboxes (there's got to be a prettier way) for carb, protein, and fat. Check the ones that are relevant to your meal and that would dictate what FPU considers the delay factor. Each also has a slider for (1-5 rating) for how much of it you ate. If you don't check carb you don't need anything up front. The actual math isn't something I can speak to how we make it work right but I do think the point that having people input three macros correct isn't user friendly when most already eyeball the first.
> 
> Loop should be smart enough to figure it out, that's half the point of an APS. I've used loop with FPU shortcut and this branch and we can do something more friendly.

dyjaks:
@gwelda 
You could, if you wanted to, mimic that without FPU which is what a lot of people don't already know. You can forecast "carbs" one hour in the future and choose a longer absorption based on how much fat/protein there is.  
  
FPU abstracts that math away from you but it is still operating in the confines of what carb absorption you could, as a user, already do. What we can do is further hide it because the less user input, or the friendlier the input, the better.

mitchellhenke:
Loop’s FPU does not match the research exactly around the duration in particular, but the math isn’t too bad relatively speaking 🙂
The math is written in code in Loop here: https://github.com/mitchellhenke/LoopKit/blob/815654406e55f0a2ed5ac3e283052151049b10fa/LoopKitUI/CarbKit/CarbEntryEditViewController.swift#L108
I also wrote it in Elm here: https://github.com/mitchellhenke/ElmWarsawInsulin/blob/master/src/Main.elm#L86
The research itself: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2901033/

The algorithm is based on the total amount being eaten as well as the ratio between carbohydrates and fat/protein.

An interface that provides the “size” could be sufficient, though the sizes ultimately have to become a number in Loop. Getting that approximately correct may work though given the longer timeframes.

Happy to discuss with anyone that has ideas/feedback or wants to develop on Zulip. Have a thread going here: https://loop.zulipchat.com/#narrow/stream/144182-development/topic/FPU

gwelda:
> @gwelda
> You could, if you wanted to, mimic that without FPU which is what a lot of people don't already know. You can forecast "carbs" one hour in the future and choose a longer absorption based on how much fat/protein there is.
> 
> FPU abstracts that math away from you but it is still operating in the confines of what carb absorption you could, as a user, already do. What we can do is further hide it because the less user input, or the friendlier the input, the better.

@ dyaks
Thank you for the tip. Will try it out!

dyjaks:
@mitchellhenke @ps2 
What would be your thoughts on using a hacked Fibonacci sequence for amounts? The idea is similar to why it's used in agile: the larger the number the less likely it is to be accurate? Maybe 5, 10, 20, 40, 90? The 40 to 90 jump scares me. It also limits users who want to YOLO at the spaghetti factory unless they input a meal many times.

dyjaks:
@ps2 @Jacob2877 
The more I thought about it the more I don't like the slider idea OR the checkboxes. Checkboxes are ugly, users never get them right (8 years of working with users to back that up), and they don't match the Loop UI. Sliders introduce false precision. I then thought of this:

https://imgur.com/a/holMiPa

A food triangle. Not sure how many tiles but I think (hope) the idea is intuitive. If you eat a straight carb meal there is an icon there, the tip, that is only carbs. Same for fat (butter) and protein (egg whites) with example foods all throughout. Nobody wants to look up the macros for a doughnut to configure the slider correctly so let's just do that for them.

Then we have the meal size, XS, S, M, L, XL. This is where I'm stuck. Ideal would be user configured based on calories. That then makes the FPU algorithm easy to go off of. Problem is, people underestimate calories and there is some physiological issues even thinking about them. So how do we determine what is S and XL?

If I/we can answer that then users are just needing to pick a food that is close and a size, Loop does the rest.

Feel free to tear this apart. :)

rsilvers129:
"The direction Loop needs to go is towards requiring less of the user in terms of input."
This is simple. Just have a toggle switch in the settings. Turn on FPU mode, and the fat and protein fields appear. They can be off by default. Even with them on, you can just not put in values, and it behaves normally.

"I think an updated carb absorption algorithm that accounts for fat & protein based on the three main choices (lollipop, taco, pizza) or something similar, would be acceptable"
I know this will never be able to estimate the fat and protein in the meal, especially for people on low-carb or Keto diets. 

"Finally, the algorithm updates should be integrated into the carb absorption model in a thorough way; not 'hacked' in, like using the FPU terms to modify number of carbs or duration terms."
Most of the medical research (I provided citations for them all over the Wiki) converts the fat and protein to equivalent carbs, so by trying to do it another way, you would be discarding 15 years of medical research papers. I didn't invent this stuff - I coded up what has already been medically tested - and I was in tough with two of the prominent authors - one of which was my Endo. Also went over it with the Chief Medical Officer and Senior Vice President of Joslin clinic.

Please, add it with a toggle in settings or else it will take years to be done another way. Low-carb people need it. People who are precise with nutrition apps want it. I know you said that the beauty of open source is that people can use any branch they want, but having it not be in the main branch is basically death for this feature.

rsilvers129:
> Loop’s FPU does not match the research exactly around the duration in particular, but the math isn’t too bad relatively speaking

It doesn't exactly match because I had a conversation with the research paper's author, and she said that it was coded for a manual pump with programming limitations. She gave me guidelines about how the duration should be coded for a computer, and I implemented those. She couldn't tell people "In an hour start a square wave" but a computer can do that.

dyjaks:
MY $0.02. Feature toggles in production are a recipe to maintainable code. One toggle is two code paths, two toggles make four, three make eight, it escalates very quickly. If there were to be a toggle it be something to have manual input vs meal guesstimate input but both should execute the same math.

rsilvers129:
It could be the same code path, but the fat and protein fields would be hidden on the meal screen - leaving fat and protein filled in as 0.

dyjaks:
That doesn't get around the bigger issue outlined here:
https://loop.zulipchat.com/#narrow/stream/144182-development/topic/FPU/near/165470322

rsilvers129:
I made a poll to see if the actual users wanted more control or more simple:

https://www.facebook.com/groups/TheLoopedGroup/2262841400599180

dyjaks:
Two things on that:
1) In all my years of professional software development I can count on one hand how many times users knew what they wanted.  
2) It's on you then to address this scenario:
I enter 5g carbs, 60g protein, 30g fat. 
There are now 2 entries in Loop, 1 for 5g with 3 hour absorption and one with 51g and 10 hours absorption.  
I now realize I fucked up and I ate 40g protein and 20g fat and I realize that 2 hours later. 
The second bit of "carb" should actually be 34g with 8.5 hours absorption.  
Users know what they want, right? With your model they total know that they 51g of "carbs" should be 34g and that the absorption lost 1.5 hours.

rsilvers129:
The easy thing to do would be to just delete the equivalent carb entry and be no worse off than the current version of Loop where one goes high if they eat that much fat and protein. Or edit it to a smaller amount - about half would be good. While I use a gram scale and nutrition app, most people estimate these macros anyway - so they can estimate their edit as easily as they can estimate the initial entry.

I mean yes, coding in editable fat and protein is better, but not if no one does it. After all, I coded this to begin with because no one else would do it. But if you want to do it "right" and make fat and protein editable, that would be great - it would be better than what I did, and more perfect.

But that is not what Pete said. He said that he doesn't want people to be able to enter fat and protein in the form of grams. The poll shows that people do want to enter fat and protein in the form of grams. There are a ton of low-carb/keto eaters and I am very aware of them. And it has allowed me to eat half of a large pizza and stay in range. How great is that?

I made this because I thought it would help people and be popular, but it has no chance if it is only in my obscure branch. We have an opportunity here for another innovative feature in Loop that will keep it as being a forward-pushing leader, as this is not in other systems yet. Adding editable fat and protein would make it perfect, but it is beyond my ability.

dyjaks:
> The poll shows that people do want to enter fat and protein in the form of grams.   

“If I had asked people what they wanted, they would have said faster horses.”
- Henry Ford

This discussion is now on three fronts: here, zulip, and FB. I vote for keeping it here or zulip. Before any public discussion happens, because it gives false hope which is why FB was not the right place), on getting some kind of FPU feature into Loop it needs to be decided on how it is presented to users and how Loop will be modified for it from an algorithm standpoint.

Jacob2877:
I am not a developer although I am interested in DIY programming.  I guess for me, the most important point regarding FPU is that my diabetes care doesn't work the best with a carbohydrate only model.  I am guessing this is typical for the people who responded favorably to the pool, Robert posted.    Regardless of me being low carber or not, fat and protein are very important in metabolism and how type 1 diabetes needs to be managed. 


 I like the idea that Shawn wants to innovate and make FPU or the idea of including fat and protein easy and simple.  Shawn reminds me of the late Steve Jobs at Apple.  However, with this being a DIY open source software, I am a little confused as to the problem.  If there is a person within the community that knows how to integrate fat and protein into the carbohydrate model, why not allow that with innovative changes during refinement?  


I guess if the argument is to keep it on a dev branch and not on a major release, I can live with that.  But, using fat and protein in the metabolism model of food is so important that I think that is why Robert is bringing it up again.  I have seen much better results with incorporating Fat and Protein into my loop settings than without it.  I think one of the takeaways with all this happening on here, facebook, and zulip is that this is something important and many people agree.  How to get there or what it looks like?  Not sure.  FPU is a starting point. 

On May 13, 2019 at 10:28 AM, Shawn <notifications@github.com> wrote:



The poll shows that people do want to enter fat and protein in the form of grams.

“If I had asked people what they wanted, they would have said faster horses.”

Henry Ford

This discussion is now on three fronts: here, zulip, and FB. I vote for keeping it here or zulip. Before any public discussion happens, because it gives false hope which is why FB was not the right place), on getting some kind of FPU feature into Loop it needs to be decided on how it is presented to users and how Loop will be modified for it from an algorithm standpoint.

dm61:
I'd like to give credits and thanks to @rsilvers129 who brought up front the need to address protein+fat dominated meals in Loop, and also developed a working branch demonstrating one possible approach. The branch has been tried by a number of people, and anecdotal feedback has been positive, I as far as I've seen. How to properly account for protein+fat, while at the same time keep the user interface simple, clean, robust, and effortless is a problem worth discussing. 

Here, zulip or FB are all public forums, and there is nothing wrong with having this type of discussion in the open. FB is messy, so I suggest we keep it here or better yet on the original issue https://github.com/LoopKit/Loop/issues/702 opened by @rsilvers129, just so we try we have as much as possible in place.

Jacob2877:
Sounds good to me -

dyjaks:
I think maybe a starting point could work like this:
Incorporate fat and protein into a meal but users only I put carbs. Based on the icon we determine very roughly what the fat and protein should be based on ratios. Then loop always does the fpu model of absorption and we hide absorbing tome from the display. 

Then, as a toggle (even though I hate them) enable the ui as Robert already did. 

The big change would be on the code side where we have a single start point of absorption we’d have two but they’d tie to the same entry. When you modify that entry if you have the toggle on you can modify all macros. 

 Never do we show “carbs” to be edited, users change protein carbs and fat. The math for absorption time and carb equivalent are all dealt with behind the scenes.

rsilvers129:
True that Facebook is not useful for the implementation methods, but it was useful to see if people want to enter fat and protein exactly, with guesstimate buttons, or not at all. We learned that (so far) 13.5x as many people want direct fat and protein entry vs a pizza button (216 vs 16 people). We also learned that no one wants it the way it is now (with no consideration for fat and protein). I created this on the cusp of people being macro conscious, and the response this time was even a lot larger than when I first coded this.

Pete is the decider and he didn't want this, even with editable fat and protein, so I am hoping he changes his mind. Others seem to instead be debating only if this is good enough with FPU converted to carb, or if protein and fat need to be editable. I am going to agree that editable protein and fat is the "correct" way to do it. But is perfection the enemy of good enough if it means that the feature dies entirely and no one gets to use it just because 1% of the time someone will have to delete the entry and start over? I am not sure. If it were my call, I would put it into dev, and then let it evolve inside dev to eventually have editable fat and protein.

The second option, that most of you seem to insist on, is to get protein and fat editable before adding it to dev. That is the kind of thing that Pete and developers more involved that me get to decide. I am fine with anyone's code, but don't want it to die as a feature - and I can't code that myself. It's not that I didn't think of it. It was beyond my ability.

dyjaks:
The elephant in the room for this is if it gets into dev, which is main, what does that mean for Tidepool? As I understand it the base code will be the same which I’m sure is a heavy factor into not only what gets pushed but how the underlying code is built.

rwfnf:
Did this go anywhere?

While I understand the desire to keep things simple from a user's perspective, it really seems to me that, long-term, integrating fat/protein data into Loop's data input is a must-have.

Can you imagine a day when machine/deep learning is used to analyse local historical data and even crowd-sourced data, to extrapolate how to best behave in a particular situation, for the most likely optimal outcome? The more relevant data is fed into such algorithms, the better they usually perform.

And we don't necessarily have to assume that all such data is manually entered by the user every time. For example, Loop could:

- Catalog a user's particular meals and allow entry of fat/protein the first time, store it for future instances of the same meal. Proportionally alter the values if the size of the meal is different.

- Image recognition technologies are getting really quite good, meal components could be someday identified by taking a photo of the meal with Loop suggesting values as a result, and learning over time, through corrections and/or initial settings. This would be even more accurate if a user photographs a meal the first time they eat it, future instances would be more easily recognised.

- Fat/Protein/Carb/nutritional data could be pulled from barcode/image recognition of product via online database(s) of processed foods and simple foods, this could be automatically filled into Loop's values for a meal.

- Loop could use these learning technologies to look at how your blood sugar behaved in previous instances of the same or similar meal, and make adjustments to insulin output model/levels when you have the same/similar meal in the future.

- For situations where fat/protein or even carb is not known, or the user is feeling lax, the concept of tagging could be introduced. Imagine a set of available tags: "Fatty", "Carby", "Meaty", "Large", "Small", "Sugary", "Liquid", etc. - the user could tag a meal with as little or as many of these tags, and they could serve as a proxy for Loop to make a best guess at the various required values - or to alter existing meal values.

- Going on a tangent a little here, it would be great if Loop would guide you through basal, insulin sensitivity, and carb ratio testing. It could build its own models of these ratios and rates over time, determined by covering different time range and situations. So eventually the user doesn't need to think much about such things, they are simply taken through a process that allows Loop to "learn" their required rates/ratios. And Loop would suggest such tests at optimal times (e.g. basal test when no active insulin and haven't eaten for a while).

- At some point perhaps Loop would be smart enough to figure out what likely caused something unexpected to happen after a meal, was it because the carb/fat/protein was wrong? Was it because your insulin sensitivity needs adjustment? Was it because you exercised recently? etc.

So I hope that the fat/protein effect is not ignored in Loop, as they can really be significant, and ultimately users would benefit from better blood sugar control if it is implemented in a way that is user friendly.

shawn-dyjak-pp:
This still lives on here:
https://github.com/mitchellhenke/Loop/tree/mitchell-rsilvers129-fpu-irc

But I don't think there is any intention of it getting into mainline loop. Perhaps some kind of floating carb and UAM/IRC/something new kind of approach could do it since the deviation is small enough. I dunno.

ps2:
See [my comments](https://github.com/LoopKit/LoopKit/pull/246#issuecomment-548167071) on the LoopKit ticket for this for information on what needs to happen to make this land in Loop

fmb787:
Hi,

Can I use this changes in last version 2.2.4 !?

mitchellhenke:
@fmb787 It won’t work, no. I am thinking about taking another look after the next major release, which includes some bigger changes to meal entry.

PR #875, opened 2019-01-29
Corrected German localizable strings

UlricusR:
Some German strings in Loop were wrong or at least not a good translation - fixed this. Request to merge into master.


ps2:
Thanks!

PR #891, opened 2019-02-18
Success haptic in BolusInterfaceController#deliver

diggabyte:
Plays a success haptic when sending a bolus from the WatchApp. We already a play a success haptic when adding a carb entry, so it seems both logical and consistent to do the same for the bolus entry.

### DIFF
--- a/WatchApp Extension/Controllers/BolusInterfaceController.swift
+++ b/WatchApp Extension/Controllers/BolusInterfaceController.swift
@@ -148,6 +148,7 @@ final class BolusInterfaceController: WKInterfaceController, IdentifiableClass {
                         if let error = error {
                             ExtensionDelegate.shared().present(error)
                         } else {
+                            WKInterfaceDevice.current().play(.success)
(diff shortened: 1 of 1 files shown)


ps2:
This doesn't play a haptic when bolus is successfully issued to the pump.  It plays a haptic when the bolus has been successfully issued to Loop.

Because it can take a while for the command to get to the pump, and for Loop to confirm it, it's better to have the pattern be to notify the user on failure, so we can let the user get on with things rather than waiting.

diggabyte:
Yeah, never intended for the haptic to play on bolus delivery complete. I have noticed it is very easy to cancel / close the bolus screen in the watch app unintentionally (cancel is pretty tight to other controls), and there's no way to know if you accidentally closed the bolus screen or sent it without checking the pump. 

Perhaps cancelling / closing from the bolus screen could play the error haptic instead? Just some way to distinguish between the two outcomes other than being 100% certain you tapped the right area.

PR #919, opened 2019-04-18
Show 24h time format on charts for non-English devices.

reelman:
It's a small change.

If you use non-English language on iPhone it's most probably you have 24h time format instead of 12h am/pm. Loop is using everywhere correct time format except x-axises on the charts.

Before

My change instead of "1 PM" show "13:00" for an example.
English
Polish

It was tested on my instances and used for a couple of months. 
Now I wish to share this with other Loopers.


ps2:
Thanks for submitting this! It's a good fix. I had some changes to the charts that needed to land; now that they're in, could you reapply your changes to the current code?

reelman:
Sorry, I'm not sure. Would you like me to create new PR or just update my PR branch? I just did update but I can't see where is the conflict in StatusChartsManager.swift.

ps2:
If you're unfamiliar with merging or rebasing to get your changes applied to the latest version of the dev branch, then creating another PR is just fine. :)

ps2:
Also, one more change is needed; the date formatter should be created once instead of on each call.

ps2:
Updated version here: #952

vbaderks:
Hi Pete,

Could this commit also be merged to the omnipod testing branch? Or do you want me to create an new issue/PR for it?

PR #942, opened 2019-05-01
Adds a blur StatusBar

idoodler:
This PR adds a blured Statusbar to the main entrypoint of the app. This is just a small UI change, but it makes the app a better experience in my opinion.

I also had to add the `idea` folder to the `.gitignore` file which won't hurt either.
IMG_B059822CBA68-1


idoodler:
There has been an issue with portrait mode.

PR #947, opened 2019-05-04
include pump manufacturer and model in the pump status

jasoncalabrese:
depends on https://github.com/ps2/rileylink_ios/pull/522

this will let us display 50+ U on the pump pill in Nightscout

uploaded device status looks like this:

```
"pump": {
  "bolusing": false,
  "suspended": false,
  "secondsFromGMT": -25200,
  "model": "Eros",
  "pumpID": "1F01ACAB",
  "clock": "2019-05-04T19:19:35Z",
  "manufacturer": "Insulet"
}
```

```
"pump": {
    "suspended": false,
    "pumpID": "999999",
    "manufacturer": "Medtronic",
    "model": "522",
    "battery": {
      "percent": 100
    },
    "reservoir": 90.6,
    "bolusing": false,
    "clock": "2019-05-04T05:13:58Z",
    "secondsFromGMT": -25200
  }
```

### DIFF
--- a/Loop/Managers/NightscoutDataManager.swift
+++ b/Loop/Managers/NightscoutDataManager.swift
@@ -152,6 +152,13 @@ final class NightscoutDataManager {
                 bolusing = false
             }
             
+            let currentReservoirUnits: Double?
+            if let lastReservoirValue = deviceManager.loopManager.doseStore.lastReservoirValue, lastReservoirValue.startDate > Date().addingTimeInterval(.minutes(-15)) {
+                currentReservoirUnits = lastReservoirValue.unitVolume
+            } else {
+                currentReservoirUnits = nil
+            }
+
             pumpStatus = NightscoutUploadKit.PumpStatus(
                 clock: Date(),
                 pumpID: pumpManagerStatus.device.localIdentifier ?? "Unknown",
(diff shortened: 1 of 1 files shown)


Kdisimone:
works well for us, tested this on both mdt and pod loops to make sure there weren't any errant issues.

ps2:
If the reservoir data is older than 15 minutes, then the uploaded `reservoir` field will be nil, which NS would display as "50+ U", right?

ps2:
There probably needs to be some semantic support in Nightscout for reservoir values that are missing vs above reading threshold.

jasoncalabrese:
We could use a special reservoir value, in combination with manufacturer/model or maybe only send include the pump info when it's current?  The timestamp on the pump info in the device status will still be old in this case, so the user may start getting stale pump data alarms depending on their settings.

jasoncalabrese:
I was thinking the pump clock was the last time we read the state, but it's just the current time `clock: Date()`

ps2:
It sounds like the NS convention would be to not include the `pumpStatus` part if pump data was stale. On the Loop side, I think the PumpManager could be updated to vend reservoir status, so you don't have to fetch it from DoseStore, and we'd also include a date for staleness. I'll merge this PR with the known issue (shouldn't have big effects), and we can tackle fixing it with the backend upload generalization.

PR #952, opened 2019-05-11
Show 12h or 24h time on chart x-axis depending on locale

ps2:
Updated version of #919. Thanks @reelman!


reelman:
Thank you for doing it in right way 🙂👍

eszcloud:
Does "Show 12h or 24h time on chart x-axis depending on locale" mean that the x-axis format will change if we travel to a new locale (e.g. US to Europe) and update the timezone to a local one?

ps2:
No, only if you change your time format in settings or change the language (locale) on your phone.

eszcloud:
Got it. Thanks!

PR #969, opened 2019-05-26
Blured Statusbar

### LINKED ISSUE #968: Statusbar overlaps UI
**Describe the bug**
The iOS Statusbar overlaps UI elements when scrolling

**To Reproduce**
1. Open the Apps main Graph view
2. Scroll down

**Expected behavior**
The UI should not overlap

IMG_B5F8C39CB09E-1

idoodler:
This PullRequest adds a blured Statusbar to prevent any UI overlapping.

iVBORw0KGgoAAAANSUhEUgAABS0AAAo4CAYAAAC8JoK+AAABgmlDQ1BzUkdCIElFQzYxOTY2LTIu-2

This PullRequest fixes #968

### DIFF
--- a/Loop/View Controllers/StatusTableViewController.swift
+++ b/Loop/View Controllers/StatusTableViewController.swift
@@ -24,11 +24,16 @@ private extension RefreshContext {
 final class StatusTableViewController: ChartsTableViewController {
 
     private let log = OSLog(category: "StatusTableViewController")
+    private let visualEffectStatusBar = UIVisualEffectView(effect: UIBlurEffect(style: .light))
 
     lazy var quantityFormatter: QuantityFormatter = QuantityFormatter()
 
     override func viewDidLoad() {
         super.viewDidLoad()
+
+        // Adding a visualEffect view to prevent the StatusBar from overlaying the UITableView
+        visualEffectStatusBar.frame = UIApplication.shared.statusBarFrame
+        view.addSubview(visualEffectStatusBar)
 
         statusCharts.glucose.glucoseDisplayRange = HKQuantity(unit: .milligramsPerDeciliter, doubleValue: 100)...HKQuantity(unit: .milligramsPerDeciliter, doubleValue: 175)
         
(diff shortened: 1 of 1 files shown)


Loop/View Controllers/StatusTableViewController.swift
```diff
@@ -24,12 +24,17 @@ private extension RefreshContext {
...
+    private let visualEffectStatusBar = UIVisualEffectView(effect: UIBlurEffect(style: .light))
 
     lazy var quantityFormatter: QuantityFormatter = QuantityFormatter()
 
     override func viewDidLoad() {
         super.viewDidLoad()
 
+        // Adding a visualEffect view to prevent the StatusBar from overlaying the UITableView
+        visualEffectStatusBar.frame = UIApplication.shared.statusBarFrame
+        view.addSubview(visualEffectStatusBar)
```
ps2: It's not ok to add a subview to the view hierarchy here, as it is management responsibility of the UITableViewController superclass

idoodler: @ps2 I am sorry, why can't one add a subview to the view hierarchy in the `viewDidLoad` of the `UITableViewController`? The transformation when scrolling takes place in the `scrollViewDidScroll`.

ps2: By reaching into views that aren't owned by us, we risk messing with behavior that we don't know the details of. Even if it seems to work now, there's no guarantee it won't break in future versions. The proper thing to do is to change this controller to not be a UITableViewController, and to have it layout its own UITableView in its own view layout, along with a blur view.

idoodler: @ps2 I'm sorry, but I don't get your point. 
Why is it necessary to change the parent class of this ViewController to a `UIViewController`? 
The `UIViewController` would need to handle the `UITableView` behaviour which is already been done by the `UITableViewController`. Adding the `UIVisualEffectsView` is just a small modification using public Apple API and the frame from the Statusbar, so changes (if any) will be well known.

ps2: With a UITableViewController, the view hierarchy is owned by the UITableViewController, and we are not protected from changes to how it manages its internal view hierarchy. If we create a view controller and place a UITableView inside it, we do control the view hierarchy, and can do what we want with it.

ps2:
Closing due to stagnation.

PR #1023, opened 2019-07-21
Fix display of correction range override

ps2:
Fixes overrides being drawn incorrectly (wrong units) when schedule has different units than currently being displayed.


bartsopers:
I just rebuilt Loop and this problem still exists for me (for both custom and predefined presets)

PR #1025, opened 2019-07-21
Remove output file list for copy-frameworks script to ensure it runs …

ps2:
…every time.


ps2:
Fixed in a separate PR

PR #1026, opened 2019-07-22
Include OmniKit and MockKit in Loop app extension

ps2:
Fix issue with Loop Status Extension not showing pumpmanager hud views correctly.


ps2:
Fixed in #1033

PR #1052, opened 2019-08-13
Use default insets for labels below Watch glucose graph

elnjensen:
This prevents clipping of the text on the rounded corners of 40mm and 44mm Series 4 Watches.

### DIFF
--- a/WatchApp/Base.lproj/Interface.storyboard
+++ b/WatchApp/Base.lproj/Interface.storyboard
@@ -414,7 +414,6 @@
                                             </separator>
                                         </items>
                                         <color key="backgroundColor" white="0.0" alpha="1" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
-                                        <edgeInsets key="margins" left="0.0" right="0.0" top="0.0" bottom="0.0"/>
                                     </group>
                                     <connections>
                                         <outlet property="bottomSeparator" destination="VV1-lG-fC0" id="T5d-VF-GlT"/>


ps2:
I heard this was a platform bug, finally fixed in recent watch builds. Can you include screenshots from both pre-series 4 and series 4 devices, before and after?

I want to make sure this doesn’t regress the older models where text should be edge-to-edge.

This also needs to be updated, presumably:
```
ChartHUDController.update():
            if #available(watchOSApplicationExtension 5.0, *) {
                cell.setContentInset(systemMinimumLayoutMargins)
            }
```

elnjensen:
Interestingly the screenshot from the Watch itself doesn't show the rounded corners - it appears to use a square display canvas. 

Here's a "before" shot, both real-world wrist photo and screenshot: 

wrist_before
screenshot_before

and here's the "after" view, both ways: 
wrist_after
screenshot_after

My watch isn't running the very latest WatchOS so I'll update and see if that matters.  But looking at the code, it makes sense to me that explicitly setting the left offset to 0 as we were doing before is necessarily going to clip - letting the system determine the defaults seems like the right approach. 

I'm traveling now but when I'm back home could test this on an older (pre-Series-4) watch.

ps2:
In those screenshots there is also a difference in the chart label placement, which you didn't change. I think there is already code in there to fix the insets. Can you make sure your "before" shots are running the latest code?

elnjensen:
Good eye - I had initially mistakenly changed the label-inset code on the graph when I was trying to figure out how to fix this.  I'll install on a clean dev and do before and after shots.  I'm updating my watch right now to the latest WatchOS (5.3) and will test after that. 

Regarding the update code you posted in the initial comment, my understanding is that these insets don't change based on content, but are static on a per-model basis, in order to keep content within [the content-safe area](https://developer.apple.com/documentation/watchkit/storyboard_elements/building_watchos_app_interfaces_using_the_storyboard/supporting_multiple_watch_sizes).  If that's true, then they shouldn't need to be updated.

elnjensen:
OK, after updating my watch to WatchOS 5.3, the display is now correct with the dev code as-is, so I'll close this PR.   Apparently it was indeed a bug, as you say.  It's a small nit, but I'll note that by putting the edge offsets into ChartHUDController rather than just allowing the system to set them as default for the container in the layout, the initial draw of the status info clips the label, and it isn't fixed until the first successful update of that info.  Not a big deal at all, but the other way (container with system default insets) seems simpler - though it would have to be tested on the pre-Series-4 watches to make sure it works. 

Thanks for your help with this, and sorry for the false alarm.

Issue #1039, opened 2019-07-28
Overrides correction range units issues


ps2:
**Describe the bug**
Overrides store correction ranges without an associated unit. If the user switches their preferred blood glucose unit, the unit will be displayed wrongly.  Also, in dev there is this line: https://github.com/LoopKit/Loop/blob/dev/LoopUI/Extensions/ChartPoint.swift#L63, which assumes the unit is mg/dL, which is wrong for users with mmol/L.

The unit of the defined correction range should be stored in the override settings, or enforce a fixed storage unit, like mg/dL.

Kdisimone:
closed with fix

PR #1058, opened 2019-08-16
Track units for overrides target ranges

ps2:


Kdisimone:
Addresses https://github.com/LoopKit/Loop/issues/1039

PR #1038, opened 2019-07-28
Override ns upload

Kdisimone:

PR #1064, opened 2019-08-17
Upload Loop settings to NS when they change

ps2:
Similar functionality to what is in#682 and #1038, but NS models are in rileylink_ios, uploads are triggered by settings changes, and Loop settings model is decoupled from the model that represents them in NS.


Kdisimone:
Thank you!!

PR #1065, opened 2019-08-18
100% autofill recommend on watch app bolus

Kdisimone:
Hoping to get a common customization in to help end jojo branching...

### DIFF
--- a/WatchApp Extension/Controllers/BolusInterfaceController.swift
+++ b/WatchApp Extension/Controllers/BolusInterfaceController.swift
@@ -86,7 +86,7 @@ final class BolusInterfaceController: WKInterfaceController, IdentifiableClass {
         var pickerValue = 0
 
         if let context = context as? WatchContext, let recommendedBolus = context.recommendedBolusDose {
-            pickerValue = pickerValueFromBolusValue(recommendedBolus * 0.75)
+            pickerValue = pickerValueFromBolusValue(recommendedBolus * 1.0)
 
             if let valueString = formatter.string(from: recommendedBolus) {
                 recommendedValueLabel.setText(String(format: NSLocalizedString("Rec: %@ U", comment: "The label and value showing the recommended bolus"), valueString).localizedUppercase)


elnjensen:
I think this is a really good idea.  Even for people who don't take the time to customize, I suspect the 75% factor on the watch is more confusing than helpful - the idea that the two should behave differently is not at all intuitive.

ps2:
The recommendation is the same on both. The input mechanisms vary. On the watch, we use a scroll wheel, so we need to pick somewhere to start (prefill 75%). On the phone, the input is empty until the user enters a value. I.e. no prefill. There is a shortcut they can take to accept the full recommendation. 

Just saying this to counter the common misperception that watch and phone make “different” recommendations.

However I do think that it is most common that people accept the full recommended bolus, so I’m leaning towards accepting this change.

WatchApp Extension/Controllers/BolusInterfaceController.swift
```diff
@@ -86,7 +86,7 @@ final class BolusInterfaceController: WKInterfaceController, IdentifiableClass {
         var pickerValue = 0
 
         if let context = context as? WatchContext, let recommendedBolus = context.recommendedBolusDose {
-            pickerValue = pickerValueFromBolusValue(recommendedBolus * 0.75)
+            pickerValue = pickerValueFromBolusValue(recommendedBolus * 1.0)
```
sascha-egerer: Why not remove the `* 1.0`?

Kdisimone: The red line is the line that will be edited out, the green line is the line that will be included after the PR is accepted.  The PR is to include the 1.0 multiplier in place of the 0.75 multiplier.

ps2:
Looking into providing a little extra step as confirmation, if we're going to go this route of starting at the full recommendation.

bartsluimer:
Would it be possible to have the 100% option if you have a Apple Watch with code protection. So it will only work when the owner is wearing the watch.

Kdisimone:
@bartsluimer the watch doesn’t know who’s wrist it is on...owner vs stranger. So, no...not possible in that respect regardless. Sounds like you have a use case worry in mind?

bartsluimer:
> @bartsluimer the watch doesn’t know who’s wrist it is on...owner vs stranger. So, no...not possible in that respect regardless. Sounds like you have a use case worry in mind?

Not directly. Just an idea for a bit more safety. We now just change code. For us 100% as standard would be fine.

scottleibrand:
@bartsluimer are you talking about the Apple Watch feature where it remains unlocked (and doesn't require a PIN) as long as it stays on the same wrist it was on when the PIN was entered?

Kdisimone:
@scottleibrand technically the code on watch isn't needed so long as (1) the phone (that is paired to the watch) is unlocked and (2) the watch is on a wrist. So, I don't think its about being on the same wrist afaik.

ps2:
@Kdisimone the watch can be locked, so that when it is removed from a wrist, it requires a code to unlock it.

The confirmation step I mention above is not about authorization, but just about adding a confirmation, to help reduce accidental boluses, and to ensure that the user is aware of the full bolus.

scottleibrand:
Are you describing the current Loop behavior? I'm describing the Apple Watch feature (independent of Loop) called Wrist Detection: https://www.payetteforward.com/how-to-turn-off-wrist-detection-on-apple-watch-fix/

I don't know if that feature is accessible to apps like Loop to allow them to lock when the phone is removed from a wrist, but if so it sounds like Bart may be suggesting using such a feature?

Kdisimone:
@scottleibrand @ps2 I haven't played much with code-locked watches...only a bit. As I had read from apple's screen prompts yesterday while setting up someone on Loop with new apple watch, the watch described the code lock as "when watch is on a wrist and the phone is unlocked, the code isn't needed"...didn't mention that taking it off a wrist and replacing it on a wrist would trigger a code need again, but that makes sense if that is the case I suppose. Just wasn't explicit in Apple's onscreen prompts.

Kdisimone:
Implemented better with a different PR

ps2:
https://github.com/LoopKit/Loop/pull/1132

PR #1071, opened 2019-08-20
Store min and max for overrides in mg/dl on Watch

elnjensen:
This aligns the override data structure with how target ranges are stored on the Watch, so both can be graphed with the same code assumptions. This fixes plotting of override ranges in mmol/L on the Watch graph.


ps2:
Thanks Eric, for sleuthing this out. Fixed in #1072

PR #1079, opened 2019-08-25
Use Minimed and Omnipod PumpManagers as Plugins

ps2:
Code, assets, and dependencies for PumpManagers can now be bundled in a plugin, that is installed into the Loop app, but not linked directly.  Each of the plugins has a metadata in its plist file that lets Loop identify it as a Loop Plugin, and show it as an option to the user.


francescaneri:
Schermata 2019-08-27 alle 14 59 04
@ps2 compilation error. am i wrong?

francescaneri:
Schermata 2019-08-27 alle 18 03 48
it still gives me error both with workspace and by downloading the zip

francescaneri:
Version 11.0 beta 5 (11M382q)


Il giorno mar 27 ago 2019 alle ore 18:11 Pete Schwamb <
notifications@github.com> ha scritto:

PR #1101, opened 2019-09-12
watch app install in build settings turned on

Kdisimone:
The watch app not installing seems to be a common issue. Users have installed Loop, paired watch keeps failing to install despite being listed in the available apps for installation. Some work-arounds have been changing the deployment targets, but that is not a fix that works for everyone (nor gets at whatever the underlying issue is apparently). This fix has worked for myself and another as reported on Zulipchat.

My previous attempts (and @elnjensen too) to get around the install failure have been pretty labor intensive and involve repairing the phone from scratch. That seems excessive now if this solution works.


ps2:
This does not change the IPA, so I do not think it is affecting user's ability to install. We need to see if we can capture the cause of the install failures; perhaps via a sysdiagnose.

Issue #660, opened 2018-01-06
[Feature] Set Temporary BG Correction Overrides Remotely via Nightscout


kenstack:
I've created some code (links below) to allow Loop to receive temporary BG correction targets remotely (from nightscout, though using IFTTT etc is fine) using Loop's override functionality for glucose targets.  Instead of writing over user created preMeal and workout ranges, I created a new override called remoteTempTarget.  It functions in the same way.  I am very new to Swift, Xcode and Loop and Im far from a professional programmer so I'm sure this will need quite a lot of code review from the experts here, if the community wants such a feature in the first place.  This  opens up the usual security issues, but as far as I know no one has had any negative situations on the openAPS side which has this same feature.  I think this feature will be particularly useful for those with small kids in a variety of situations, and for those of us who simply want to set some new targets remotely (ie not getting out of bed) if our PWD is trending high or low and we want to increase or decrease insulin for a period of time via changing the target.  Again, the community obviously needs to decide if we want to open the door to allowing remote commands to be sent.  So Ill put the idea out there for debate and discussion. 

I've been running it for a few days now and @Kdisimone has tried it as well.  Ive been setting temp targets via NightScout and so far so good.  Im 100% sure the code could be cleaner and Im wide open to advice and criticism as Im using this to learn Loop.  This same idea of course can be applied to other remote commands, but I think this one is the most basic.  

Right now the code ignores the "type" of override sent from NS - eating soon etc - any Temporary Target events that are sent to Treatments are applied to remoteTempTarget and do not overwrite preMeal or workout, but that was just my thought.  The requirements on this also need a lot of input from the community. This is a first cut.

https://github.com/kenstack/Loop/tree/nstemptarget
https://github.com/kenstack/LoopKit/tree/nstemptarget

if you want to try it you need to update carthage as there are a few small changes to loopkit which I couldn't figure out how to avoid - see #3 below - Im hopeful someone can point me in the right direction on how to keep all the changes in loop.  The cartfile  should point correctly to my github branch but let me know if there are issues.  

There are some issues Id like to discuss if people are interested in the feature 

1. I could not find a clean way to set the ranges of an override - I can set duration, etc but I could not set the range without editing the "rawValues" directly.  Im guessing either I missed how to do this with the available methods or its probably better to create an extension to do this cleanly.  I tried the latter but failed... 

[...]

PR #1108, opened 2019-09-15
old jojo code for NS remotes as a starting point

Kdisimone:
Just as a starting point...here's the old code from jojo for NS remote overrides (need fix for units still since dev fixed that). LoopKit also has a very small change for handling notification cases (case remoteTempSet & case remoteTempCancel) in LoopKit/LoopNotificationCategory.swift...I don't have a PR in for that currently. If it is helpful, I'd be happy to get a workspace branch going to help collaborate? Or ???

### DIFF
--- a/Loop/Managers/DeviceDataManager.swift
+++ b/Loop/Managers/DeviceDataManager.swift
@@ -451,6 +451,187 @@ extension DeviceDataManager: PumpManagerDelegate {
         }
     }
 
+    //////////////////////////////////////////
+    // MARK: - Set Temp Targets From NS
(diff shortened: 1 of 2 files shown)


Kdisimone:
@jasoncalabrese @kenstack

jasoncalabrese:
I was thinking that it was about time to start working on a new branch, I’d need the insulin delay and some other stuff like the bolus guard change

ps2:
I added some notes to https://github.com/LoopKit/Loop/issues/660 earlier today. I'd love to get this going.

PR #1127, opened 2019-09-27
Added checks for bash profile / bashrc to build scripts

JamieKeene:
For policy reasons people may not have brew installed in the default location, and use bash profile to manage their paths.

This PR:
- Adds checks for non-empty .bash_profile and .bashrc in the user's $HOME folder,
- Works around the hard-coded carthage location where carthage is detected


ps2:
Sourcing the user's bashrc or profile opens up a large can of worms of possible ways to break the build. I would recommend adding a new User-Defined Build Setting to your workspace (something like `CARTHAGE_BINARY_PATH`, and updating the build scripts to use that if it was set.

JamieKeene:
Closing to retry with a new fork.

PR #1137, opened 2019-10-08
Adding checks for Bash profile & rc

JamieKeene:
For policy reasons users may not be able to install Homebrew in the default location, so could use a profile to add this location into the user's path.

The current project assumes the default location. This PR removes this hardcoding of location.

### DIFF
--- a/Loop.xcodeproj/project.pbxproj
+++ b/Loop.xcodeproj/project.pbxproj
@@ -2350,7 +2350,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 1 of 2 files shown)


ps2:
Sourcing the user's bashrc or profile opens up a large can of worms of possible ways to break the build. I would recommend adding a new User-Defined Build Setting to your workspace (something like CARTHAGE_BINARY_PATH, and updating the build scripts to use that if it was set.

JamieKeene:
Thanks @ps2 – will try this instead and update the PR.

JamieKeene:
@ps2 I tried to do this today but realise that i would also need to submit a PR to the rileylink-ios repo to look for the same user-defined variable. If I use a profile to source the locations this cascades down, however because the copy-frameworks.sh file is baked into the rileylink repo sending the carthage path directly once causes failures down the line.

I'm happy to do this if you'd support it, otherwise I can abandon this change and modify the source each time for my own setup's weirdness.

Thanks in advance for your advice!

PR #1153, opened 2019-10-17
Fixed building from directories containing spaces

ktomy:
SRCROOT environment variable is not escaped, when using it as part of a path it should be escaped or enclosed in quotes

Tested on Workspace build


ktomy:
@ps2 could you please review?

PR #1157, opened 2019-10-20
Spike master

ramzxxx:
spike-loop crashes everytime I make a call or send a text


ramzxxx:
> spike-loop crashes everytime I make a call or send a text

i apology for not understanding, but i'm very new i don't know what to do  do have to change the code in loop or something

Issue #864, opened 2019-01-11
Feature Suggestion: Set Fast, Medium, Slow carb absorption times to 30 minutes, 3 hours, 5 hours.


brandonarbiter:
**Feature suggestion**

Currently, Loop carb absorption times are 2, 3, and 4 hours for fast, medium, and slow carbs, respectively. 

I find it really helpful to use the fast carbs for glucose rescue (e.g., eating glucose tabs for a low BG). But that's more like a 30 minute carbs absorption time, not 2 hours. So I customize the setting in the code.

On the other side, the difference between 3 and 4 hours to indicate a slow meal isn't very much. 

I suggest we set the carb absorption times for fast, medium, and slow to 30 minutes, 3 hours, and 5 hours, respectively.

Kdisimone:
The lollipop at 2 hours is too long...but 30 min might be a too fast, but we don’t use that entry often. I’d think something like 60 min might be closer to the wide variety people use for lollipop coverage. Like many fruit are longer than 30min for us. Closer to 1 hour overall for apples and bananas. We use lollipop for fruit and do wish it were quicker than 2 hours, but more like 60min. 

And 5 hours for pizza icon would be too long for us for most of our pizza entries. Currently, 4 hours starts at 5.5 hours for DCA. So if pizza was 5 hours it would start at 7.5 hours. That’s a 2 hour difference. Pretty significant DCA change compared to what people are currently used to using. They would be getting A LOT less up front bolus. So if the idea is to cover later carbs better, that’s going to come at a significant cost of upfront bolus. If you’re getting early lows and late highs, than that may work well...but otherwise you’ll more highs early from smaller upfront bolus. 

Just my two cents.

Kdisimone:
👍👍

beached:
One thing I have noticed is that if one puts the correction carb count in is that loop is going to potential treat for it.  If the reason for the low is incorrect basal, Boris, ISF numbers it may be more likely to bring one low again

beached:
Maybe a way of marking carbs as tracking only.  Such that they may show up in the graphs but will not be acted upon with insulin.  Assuming one isn't rage carbing the low and it shouldn't even cause a high, but even then the system will correct for it.

Is using time as that label the correct analogy?  If I don't want to track correction carbs, why not label them.  If it needs to be stored like that, so be it, but that would be hidden from the interface.

Kdisimone:
[...]

PR #1162, opened 2019-10-26

ps2:
#864


MoreSolutions:
Are there plans to merge this into `dev` soon? I've been using this customization and noticed significant improvements, especially when using fast-absorption carb entries.

ps2:
There are a few more things that need to change for this; I don't think watch input is using the right values , and CarbStore might need some more updates.  After 2.0, I'll review (or if anyone else wants to take a swing at implementing this in a more complete way), and then we'll get a little more testing on it.

PR #1175, opened 2019-11-08
iOS 13 Dark Mode UI Reset

ryandorn:
There is a bug with the UI where a user setting dark mode in iOS 13 will cause the background to go black and cause grey text to become illegible and UI to generally not be good. This change is to add a key to info.plist to force light mode within the app until dark mode can be supported.


Kdisimone:
Closing as this is master to master. Also closing because dark mode is in dev branch. When doing PRs please use dev branch to base on. And check dev to see if the issue you have might already have in master might already been resolved in dev. Dark mode came out after the last master release, so any dark mode work is in dev currently.

PR #1185, opened 2019-11-23
Current time lines on charts

ivalkou:
Added current time lines to charts for convenience.
Screen Shot 2019-11-23 at 15 07 24


Kdisimone:
Just so you're aware, this feature had been submitted once before and was closed. https://github.com/LoopKit/Loop/pull/687  Leaving the comment here as I suspect @ps2 would have referenced this in reply so I'm saving him some look-up time.

ps2:
Please describe what use case you're trying to solve with this PR. As in, without this feature, what problem do you have?

Thanks for the reference tag, @Kdisimone .

Issue #1170, opened 2019-11-04
Loop Settings not loading


ps2:
If Loop is launched in the background (via bluetooth, for example) before the first unlock of the phone, UserDefaults will be unavailable, due to it being stored with a file access protection of NSFileProtectionCompleteUntilFirstUserAuthentication. 

This will then appear to the user as if settings have been deleted. Restarting Loop restores access to the settings, unless something triggers a write to settings before Loop is restarted.

skimaniac:
I tried : ''This will then appear to the user as if settings have been deleted. Restarting Loop restores access to the settings.''

-my override setting do not comme back.
-still do not see basal on loop and no basal in apple Health

![IMG_0886](https://user-images.githubusercontent.com/44727542/68793100-9a8bf000-061a-11ea-8fef-07a12b96cc3f.PNG)
![IMG_0885](https://user-images.githubusercontent.com/44727542/68793101-9b248680-061a-11ea-9b8b-e739fc126713.PNG)
![IMG_8B2C4E784706-1](https://user-images.githubusercontent.com/44727542/68793153-b2637400-061a-11ea-92e3-0db8eefe99b6.jpeg)

skimaniac:
i also notice that when loop begin to communicate with the pump the basal appears (and the prediction line drop very low), and after few second the basal disappear again and the prediction line comme to a better view

![IMG_0888](https://user-images.githubusercontent.com/44727542/68793713-7e3c8300-061b-11ea-9dcf-e94b8248c2aa.PNG)

![IMG_0889](https://user-images.githubusercontent.com/44727542/68793711-7e3c8300-061b-11ea-9774-e603875e950e.PNG)

detroiterinnyc:
I’m having this problem too. My full settings don’t restore. I’m guessing a rebuild is in my future.

elnjensen:
Got hit by this tonight and can confirm that not all settings come back.  Insulin model, carb ratios, sensitivities, and basal rates were correctly restored after restarting the app, but delivery limits, correction ranges, suspend threshold, and override presets were lost and had to be re-entered.

ps2:
Those will only be lost permanently if something is done to trigger a settings store before restarting Loop. Unfortunately, that’s pretty easy to do. Lots of actions, like enacting an override, can trigger that.

loyalchallenger:
[...]

PR #1186, opened 2019-11-23
Fixed loss of settings after reboot.

ivalkou:
Bug: When iPhone is rebooted (for example, if battery is low) but not yet unlocked, the application may be launched by background fetch, but `UserDefaults` will be unavailable, and users lose the settings.

The bug has been confirmed by 3 users, but I can’t reproduce it.

Found a possible solution on [StackOverflow](https://stackoverflow.com/questions/20269116/nsuserdefaults-losing-its-keys-values-when-phone-is-rebooted-but-not-unlocked)


ps2:
This is documented in #1170.  This PR will not *fix* the issue. See the apple documentation for resetStandardUserDefaults. A method to reload settings on protectedDataDidBecomeAvailableNotification might be a short term stopgap, but would still not let loop loop until the phone was unlocked. The long term solution is a bigger work to move the items stored in UserDefaults to a different storage backend.

PR #1191, opened 2019-11-27
Easier code customize for Watch App bolus confirm

diggabyte:
Allows for the rotation threshold of bolus confirmation in the Watch App to be easily code customized


ps2:
If this is still desired, please add a new PR against current code. Thanks!

PR #1194, opened 2019-12-02
UPDATE

escottdexcom:


Kdisimone:
Closing. Looks like you were trying to update your dev branch? In which case do the compare again, but switch your name and LoopKit name in the URL to get the compare going in the right direction.

PR #1198, opened 2019-12-14
Settings loss patch

ps2:
Pause launch if launching before first unlock. Workaround for #1170

### DIFF
--- a/Loop/AppDelegate.swift
+++ b/Loop/AppDelegate.swift
@@ -18,49 +18,92 @@ final class AppDelegate: UIResponder, UIApplicationDelegate {
 
     var window: UIWindow?
 
-    private(set) lazy var deviceManager = DeviceDataManager()
+    private var deviceManager: DeviceDataManager?
 
     private var rootViewController: RootNavigationController! {
         return window?.rootViewController as? RootNavigationController
     }
-
-    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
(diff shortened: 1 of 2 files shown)


marionbarker:
I just tested this on my phone. I used Katie D's low power trick to make my phone mess up, then built the new version of the code and repeated the trick and Loop worked nicely.  Dexcom has added a squawk that alerts you their app isn't working properly.  Had to force quit Dexcom app both reboots.  Loop opened nicely with all settings after the fix.
iPhone 11 pro
iOS 13.2.3
gitBranch: settings-loss-patch
Omnipod

Thanks Pete.

ps2:
Thanks Marion.  Looking for other testers to try this out. Not as much to check that it fixes the issue, as I believe that is working, but to watch out for regressions.

skimaniac:
I tested it and everything looks ok now with the fix.
I tried with the cell unlocked and forced the start (power button and low volume) and also with the cell locked and forced also the start, if I remember correctly this is how I had lost my settings at least once.

My only concerns is that it takes 2 to 3 mins before a normal graph is displayed, after starting I had a prediction of 0 mmol/L in 1 hour and -16 mmol/L in about 4 h , whereas it is 6.3 in 1 hour and 7 in 4 hours.

Thank you!


Iphone 7 plus
IOS 13.3

rachel1220:
I have had loop delete my settings 2 days in a row when I was skiing. The second time it disconnected my omnipod and dexcom. I never turned my phone off and it was not on low power mode. I have an iPhone 11pro with iOS 13.2.2. Do I need to rebuild? Thanks in advance!

PR #1208, opened 2019-12-23
Feature/carb ui refinements

PorkShoulderHolder:
First time loop contributor here!

Entering carbs on my Apple Watch 40mm, I noticed that the input mode (carb grams or time) is often toggled by mistake when using the small circular increment/decrement buttons to adjust the value.  While I think the crown is probably the preferred carb data entry method, this issue should be addressed either way.  

The issue was occurring because a tap gesture recognizer was attached to the entire group containing carbs, date, and the buttons - so if you missed the tiny buttons a little bit you would spuriously toggle from carbs to time input or vice versa.  

My proposed solution makes the buttons larger and removes the toggle functionality for determining the input type.  Now input type is determined by tapping either the carb label or the date label (each now has its own tap gesture recognizer).

**Visual changes**
Before:
old
After:
new

**Tested on** 
40mm series 5: watchOS 6.1.1
40mm series 4: watchOS 6.0 
In simulator on 44mm watchos 6.1


ps2:
Making the tappable area for the + and - bigger might be an acceptable change here, but the toggle was put into place because it's too hard to tap on the time; a lot of people use the time input to pre-bolus.

PorkShoulderHolder:
Ok I will switch back to the toggle interaction.  Do we think we should fix this issue by making the buttons larger or just by reducing the size of tappable area for the tap recognizer.  For now I will leave buttons large.

PR #1217, opened 2019-12-29
Incorporate sleep data into complication user info transfer calculations 

novalegra:
As discussed in #816 and #832, complication transfer intervals could be optimized if bedtime was incorporated into the calculations, as this would make the complication updates more likely to occur during waking hours. 

This implementation uses HealthKit sleep data in order to determine "bedtime" without adding additional settings to Loop. This adds an extra HealthKit permission, for read access to sleep data. A HealthKit query for this data is made every 24 hours and uses a maximum of the most recent 30 samples from the past 6 months. The start times of these sleep samples are then averaged to arrive at the final "bedtime".  `complicationUserInfoTransferInterval` was changed from a variable to a function to allow for bedtime to be incorporated. The time until complication allowance refresh is calculated with `bedtime - now` if bedtime is in the future, or `midnight - now` if the bedtime is in the past. If there is no bedtime (bedtime is nil), the calculations are the same as in the current implementation.


scottleibrand:
Does the Apple Watch or iOS passively track sleep times and record them to Health if you don’t use an active sleep tracker? A quick check of a non-health-tracking phone shows no sleep data, so I assume not. If not, does this PR default to the previous behavior, or to a default bedtime, or something else?

Loop/Managers/WatchDataManager.swift
```diff
@@ -334,6 +372,52 @@ extension WatchDataManager {
...
+        let sleepType = HKObjectType.categoryType(forIdentifier: HKCategoryTypeIdentifier.sleepAnalysis)!
+        // get more-recent values first
+        let sortDescriptor = NSSortDescriptor(key: HKSampleSortIdentifierEndDate, ascending: false)
+        
+        let query = HKSampleQuery(sampleType: sleepType, predicate: predicate, limit: sampleLimit, sortDescriptors: [sortDescriptor]) { (query, samples, error) in
+
+            if let error = error {
+                completion(.failure(error))
+            } else if let samples = samples as? [HKCategorySample] {
+                // find the average hour and minute components from the sleep start times
```
mpangburn: Each sample's `value` property will correspond to a value of HKCategoryValueSleepAnalysis. If a user has a sleep tracking app, they might have primarily `.asleep` values; if the data comes from Bedtime in the Clock app, they'll have primarily `.inBed` values. It's possible that a user has a value of each type each night, in which case they'll both be averaged.

Consider performing an additional step here which filters samples down to those with `.inBed` values to use in the computation. If there aren't "enough", try `.asleep` instead. In any case, we'll want to make sure we don't include any `.awake` values.

Loop/Managers/WatchDataManager.swift
```diff
@@ -334,6 +372,52 @@ extension WatchDataManager {
...
+                let average = samples.reduce(0, {$0 + $1.startDate.secondsFromMidnight()}) / samples.count
+                let averageHour = average / 3600
+                let averageMinute = average % 3600 / 60
+                
+                // find the next time that the user will go to bed, based on the averages we've computed
+                if let time = Calendar.current.nextDate(after: Date(), matching: DateComponents(hour: averageHour, minute: averageMinute), matchingPolicy: .nextTime) {
+                    completion(.success(time))
```
mpangburn: There's a tricky case here where the time might not exist within the next 24 hours, like in the hour jump that occurs for DST. It's worth doing a check that `time.timeIntervalSinceNow <= .hours(24)` before proceeding.

Loop/Managers/WatchDataManager.swift
```diff
@@ -334,6 +372,52 @@ extension WatchDataManager {
...
+                } else {
+                    completion(.failure(NSError()))
```
mpangburn: You might consider defining a SleepStoreError enum to use in SleepStoreResult, with a `noMatchingBedtime` case.

Loop/Managers/WatchDataManager.swift
```diff
@@ -334,6 +372,52 @@ extension WatchDataManager {
...
+                }
+            } else {
+                assertionFailure("Unknown return configuration from query \(query)")
```
mpangburn: Since assertion failures trigger only in debug modes, it's best practice to still call the completion handler on this path—not that I'd expect it to happen!

Loop/Managers/WatchDataManager.swift
```diff
@@ -334,6 +372,52 @@ extension WatchDataManager {
...
+            }
+        }
+
+        healthStore.execute(query)
+    }
+
+}
+
+extension Date {
+    fileprivate func secondsFromMidnight() -> Int {
```
mpangburn: This function's body does what you want: produce a scalar value that encodes the time of day to be averaged. However, this scalar isn't always the number of seconds from midnight; consider DST. Something like timeOfDayInSeconds would communicate the intention more accurately.

Loop/Managers/WatchDataManager.swift
```diff
@@ -40,6 +44,37 @@ final class WatchDataManager: NSObject {
...
+        else {
+            // only look at samples within the past 6 months
+            let monthsToGoBack = -6
+            let start = Calendar.current.date(byAdding: .month, value: monthsToGoBack, to: Date())!
+            
+            getSleepStartTime(start: start) {
+                (result) in
+
+                // update when we last checked the bedtime
+                self.lastBedtimeUpdate = Date()
```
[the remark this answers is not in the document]
novalegra: This variable should be renamed, as it's intended to be the last time that we've attempted to find the bedtime. I had it update even in a failure case to ensure that where there's no sleep data, the query doesn't get attempted every time the function is called. Would this be the desired behavior?

Loop/Managers/WatchDataManager.swift
```diff
@@ -334,6 +372,52 @@ extension WatchDataManager {
 
         return items.joined(separator: "\n")
     }
+    
+    private func getSleepStartTime(start: Date, end: Date? = nil, sampleLimit: Int = 30, _ completion: @escaping (_ result: SleepStoreResult<Date>) -> Void) {
```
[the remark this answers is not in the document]
novalegra: For SleepStore, would it be more appropriate to have it in LoopKit? I can put it into the Loop/Managers file then it could be moved as needed

novalegra:
> Does the Apple Watch or iOS passively track sleep times and record them to Health if you don’t use an active sleep tracker? A quick check of a non-health-tracking phone shows no sleep data, so I assume not. If not, does this PR default to the previous behavior, or to a default bedtime, or something else?

If Bedtime, an iOS feature, is set up, then it'll track the sleep times; there are also other iOS and/or Watch apps that can track sleep. If there is no data, this PR would default to the previous behavior, which uses `midnight - now` as the interval until the complication allowance is refreshed (or a duration of 24 hours if midnight can't be found)

Loop/Managers/WatchDataManager.swift
```diff
@@ -334,6 +372,52 @@ extension WatchDataManager {
...
+        // get more-recent values first
+        let sortDescriptor = NSSortDescriptor(key: HKSampleSortIdentifierEndDate, ascending: false)
+        
+        let query = HKSampleQuery(sampleType: sleepType, predicate: predicate, limit: sampleLimit, sortDescriptors: [sortDescriptor]) { (query, samples, error) in
+
+            if let error = error {
+                completion(.failure(error))
+            } else if let samples = samples as? [HKCategorySample] {
+                // find the average hour and minute components from the sleep start times
+                let average = samples.reduce(0, {$0 + $1.startDate.secondsFromMidnight()}) / samples.count
```
mpangburn: We'll want a `guard` to check that `!samples.isEmpty` to avoid trapping here if no sleep data is available. SleepStoreError can include a `noSleepDataAvailable` case to represent the failure.

Loop/Managers/WatchDataManager.swift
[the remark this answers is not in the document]
mpangburn: That's reasonable—and it prompted me to look at the behavior of the bedtime computation function when no sleep data exists. Added an additional comment down there.

Loop/Managers/WatchDataManager.swift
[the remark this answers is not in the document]
mpangburn: Loop/Managers sounds good.

Loop/Managers/WatchDataManager.swift
novalegra: Done. For the purposes of "enough," what could be a decent threshold? The current code just needs one value for it to use the .inBed values instead of the .asleep values

Loop/Managers/SleepStore.swift
```diff
@@ -0,0 +1,121 @@
...
+//  Created by Anna Quinlan on 12/28/19.
+//  Copyright © 2019 LoopKit Authors. All rights reserved.
+//
+
+import Foundation
+import HealthKit
+
+enum SleepStoreResult<T> {
+    case success(T)
+    case failure(Error)
```
mpangburn: You can use SleepStoreError as the type contained by the `failure` case.

Loop/Managers/SleepStore.swift
```diff
@@ -0,0 +1,121 @@
...
+}
+
+enum SleepStoreError: Error {
+    case noMatchingBedtime
+    case unknownReturnConfiguration
+    case noSleepDataAvailable
+}
+
+extension SleepStoreError: LocalizedError {
+    public var localizedDescription: String {
```
mpangburn: It doesn't look like we'll be surfacing these errors to the user, so you can remove the localized descriptions.

Loop/Managers/SleepStore.swift
```diff
@@ -0,0 +1,121 @@
...
+            return NSLocalizedString("Unknown return configuration from query", comment: "")
+        case .noSleepDataAvailable:
+            return NSLocalizedString("No sleep data available", comment: "")
+        }
+    }
+}
+
+class SleepStore {
+    var healthStore: HKHealthStore
+    var sampleLimit: Int
```
mpangburn: `sampleLimit` strikes me more as a parameter to the query, rather than as a property of the SleepStore itself. Consider making this a parameter to `getAverageSleepStartTime` with a default value.

Loop/Managers/SleepStore.swift
```diff
@@ -0,0 +1,121 @@
...
+                completion(.failure(SleepStoreError.unknownReturnConfiguration))
+            }
+        }
+        healthStore.execute(query)
+    }
+}
+
+extension Date {
+    fileprivate func timeOfDayInSeconds() -> Int {
+        let calendar = Calendar.current
```
mpangburn: One detail here I forgot to mention—the time of day for a particular sleep sample is dependent upon the time zone in which the sample was recorded. For example, if a user went on vacation recently, using their home time zone to determine the time of day in seconds could yield a result that doesn't accurately reflect when they went to sleep in their _vacation location's_ time zone.

Add a parameter to this function of type TimeZone, and assign it to the `timeZone` property of `calendar`. You can retrieve the TimeZone for each sleep sample by retrieving the name string from the sample's metadata via [HKMetadataKeyTimeZone](https://developer.apple.com/documentation/healthkit/hkmetadatakeytimezone).

Loop/Managers/WatchDataManager.swift
```diff
@@ -40,6 +43,52 @@ final class WatchDataManager: NSObject {
     }()
 
     private var lastSentSettings: LoopSettings?
+    
+    let healthStore: HKHealthStore
```
mpangburn: I believe we no longer need a HKHealthStore instance directly inside WatchDataManager, since the access is managed through SleepStore; is this property used anywhere?

Loop/Managers/WatchDataManager.swift
```diff
@@ -40,6 +43,52 @@ final class WatchDataManager: NSObject {
...
+        let lastUpdateInterval = now.timeIntervalSince(lastBedtimeQuery)
+        let calendar = Calendar.current
+        
+        guard lastUpdateInterval >= TimeInterval(hours: 24) else {
+            // increment the bedtime by 1 day if it's before the current time, but we don't need to make another HealthKit query yet
+            if let bedtime = bedtime, bedtime < now {
+                let hourComponent = calendar.component(.hour, from: bedtime)
+                let minuteComponent = calendar.component(.minute, from: bedtime)
+                
+                if let newBedtime = calendar.nextDate(after: now, matching: DateComponents(hour: hourComponent, minute: minuteComponent), matchingPolicy: .nextTime) {
```
mpangburn: Here's another case where `newBedtime` could be more than a day in the future in case of DST. We could add a `newBedtime.timeIntervalSince(bedtime) <= .hours(24)` clause similarly to before; alternatively, see my comment in `complicationUserInfoTransferInterval` below.

Loop/Managers/WatchDataManager.swift
```diff
@@ -350,21 +403,28 @@ extension WCSession {
...
         let now = Date()
-        let timeUntilMidnight: TimeInterval
+        let timeUntilRefresh: TimeInterval
 
         if let midnight = Calendar.current.nextDate(after: now, matching: DateComponents(hour: 0), matchingPolicy: .nextTime) {
-            timeUntilMidnight = midnight.timeIntervalSince(now)
+            // we can have a more frequent refresh rate if we only refresh when it's likely the user is awake (based on HealthKit sleep data)
+            if let nextBedtime = bedtime {
+                let timeUntilBedtime = nextBedtime.timeIntervalSince(now)
+                timeUntilRefresh = timeUntilBedtime > TimeInterval(0) ? timeUntilBedtime : midnight.timeIntervalSince(now)
```
mpangburn: We might defensively make the condition in this ternary also check an upper bound on `timeUntilBedtime`, i.e. `(0..<TimeInterval(hours: 24)).contains(timeUntilBedtime)`, to ensure that complication transfers aren't slowed down if `bedtime` somehow ended up being a Date further in the future.

Loop/Managers/WatchDataManager.swift
novalegra: Nope, removed

Loop/Managers/SleepStore.swift
novalegra: For unwrapping the timezone, what could an appropriate default be? We could use the current time zone, or exclude the sample from the analysis (though this would probably mean it wouldn't be possible to use a reduce function to average the samples)

ps2: I just checked my sleep data, from SleepTracker, and it has timezone. Whether or not this field is present will depend on the app, and since it's metadata, the app isn't forced to store it. I think assuming local timezone in that case is probably the best option, as the user probably will have no samples with timezone.

novalegra: Sounds good. Just checked my sleep data (some from Bedtime, some from SleepCycle) and it all contains timezones

PR #1219, opened 2019-12-29
Automatic bolus dosing

ps2:
An implementation of using boluses for dosing. Independent of https://github.com/LoopKit/Loop/pull/1205. 

This feature is enabled by a new switch in settings that switches between the existing temp basal strategy and this new bolus strategy.

Boluses are issued for high corrections, when Loop thinks additional insulin is required beyond your basal needs. At each interval, up to 40% of the required insulin will be dosed via bolus. When insulin requirements are below basal, Loop will use temp basals, as before.

Automatic bolus sizes are limited to not exceed 40% of your max bolus setting.


francescaneri:
awesome wauu!
[Immagine del 30-12-19 alle 15 32](https://user-images.githubusercontent.com/9915825/71596199-f06f1580-2b3e-11ea-8454-1b9970038f83.jpeg)

erikdi:
Can we revisit the discussion from #707 around implementing a maximum Insulin-On-Board limit now?  

Having run more or less the same automatic bolus code for 2 years I still believe it is a necessary safe guard.
 
IIUC this implementation correctly it will issue an automatic bolus on every run of the loop, so every 5 minutes with a Dexcom cgm.

Given that most people run this unattended the 40% of max bolus per Loop will not limit the amount of insulin delivered in any meaningful way in practice and makes the use of it unsafe.

Let's say you have a max-bolus set to 5.  It would deliver 2 units in each iteration, ie. 4U after 5 minutes, 6U after 10 minutes, or 10U in 20 minutes when you realistically start to feel the effects of the wrong dosing. At that point it requires quite a bit of fast carbs to get out of it.

Because the carb amounts are not limited either, it is pretty easy to get into this state by entering the wrong carb amount (let's say an inexperienced caregiver mixes up total weight of food and carbs).

ps2:
@erikdi Yes, I do think a max iob guard might be necessary here. Not as a therapy setting that people use to often limit dosing, but as a protection for the kinds of scenarios you describe.

live4sw:
I agree it makes sense as a safety feature. Very similar issue as with OpenAPS with SMB enabled (OA has a max iob setting, though it is user configurable). Of course this also gets tricky with mealtimes, particularly for those with high carb diets.

dyjaks:
Since there are two approaches for this, is the intent that they go head to head and validate outcomes or will the two at some point merge?

alevar1975:
Definitely merge👍💕

dyjaks:
@alevar1975 The question would be how. If branch A does X and branch B does Y the union of the branches A&B doesn't mean the union of results X&Y. It in fact makes a whole new thing.

alevar1975:
Then AB do XY💕

alevar1975:
Then A & B do the unknown💕💕💕

dyjaks:
@ps2 Potential safety issue, would like your thoughts on. I just started a new G6 sensors so have no readings. I put in a manual BG entry at 9:12 PST. I have gotten three automatic boluses in that time at:

9:13 for .3U (good)
9:18 for .2U (probably fine)
9:23 for .1U (maybe not fine)

Because automatic boluses can stack really fast should we be more demanding of recent readings? Loop making decisions 10 minutes out from last reading after already giving two boluses might cause issue.

ps2:
That is expected behavior at this time. Loop will continue to dose as long as BG is < 15m old.  It's similar in behavior to how it works with Temp Basals, too.  This is actually more conservative than a temp basal would end up delivering.  A temp basal over 30m is 100% of Loop's calculated dose.  In your case, assuming Loop called for .75 U (based on your initial 0.3 dose being 40% of total), using temp basals Loop would deliver 0.75 U.  Using boluses with partial application, you got 0.5 U delivered, or 67% of the total dose.

dyjaks:
I thought so but always want to make sure.

dyjaks:
An update now that I've been running this almost a week. No real bad episodes, or any episodes really, have happened since that first initial scare. Seems to handle long meals better (great for low carb). I think pods like it a lot better than temp basals.

jbwittmer:
We are likewise quite pleased after a week.  I'd be curious to test with slightly more aggressive application of the calculated need.  Am I reading the code correctly that by changing     "bolusPartialApplicationFactor = 0.4 // %" to 0.5 it would apply half of the calculated corrective bolus each dosing interval rather than 40%?  Nothing else would need to be changed?

ps2:
@jbwittmer, that’s all you need to change. But remember that you need to leave room for loop to react to bg going lower than expected, like overestimating carbs, exercise, etc. 40% is already a huge jump from Loop’s previous effective partial application factor of 17%.

francesc0-cgm:
@ps2 so to mitigate the effect of automatic bolus that for my 6 years old son is way too aggressive, i can set it at 0.3-0.25 to put 30-25% ? Thank you Pete.

Inviato da iPhone

loyalchallenger:
Just to verify again, this feature will be able to be turned off/on once implemented? I have not needed a correction bolus in a very long time and sometimes my cgm predicts that I will be much higher than I ever actually get. But I’m on such low doses and low carb, that I don’t need correction boluses.

ps2:
@francesc0-cgm I think it's best to look at increasing partial application as not increasing Loop's "aggressiveness", but rather increasing Loop's confidence in dosing. The top variables you should play with if Loop is being too aggressive (i.e. calling for too much insulin) are ISF, basal, and target range. So I would recommend you try adjusting those first.  Of course all the variables play out together in treatment, and cannot be entirely separated, so moving to a bolus dosing strategy might mean that what was once an effective ISF value under temp basal strategy might be different now.

tkk41:
I've been running `automatic-bolus` since a few days, looks good to me, no issues so far. I'll carefully run this dosing strategy and in case any issues were coming up, I'll post here.

ps2:
At some point the various parts of the Loop forecast will likely have confidence terms associated with them, which will give us a better way to adjust partial application automatically. 

For now, tuning Loop is already fairly complicated, and reducing the number of free variables helps most people find their best settings quicker. Adjusting the partial application factor may make sense to vary based on the hour to hour variance of your individual response to insulin, but it is *not* a good way to adjust the average amount of insulin you need for a given situation.

If you're someone with *very* reliable response to insulin, carbs, etc, then maybe you could consider bumping up PA factor. If you're someone with a very unreliable response to those things (caution! bad settings can look like this, but is a different situation), then maybe 40% is too high.

francesc0-cgm:
@ps2 i know. I am talking about using same settings with and without automatic bolus on my son. It is completely different. With automatic bolus i find loop quite aggressive. In this way implementing automatic bolus as it is right now in dev and eventually in master in children and young kids should implement severe changes in every parameter. For adults could be different looking at other comments

Inviato da iPhone

denis-gozdzik:
@ps2 Pete, kind of newbie question - is this automatic bolus dosing implemented in dev already? Does it mean that I can download dev and test it?

ps2:
@denis-gozdzik No, it's not in dev. You should bring yourself up to speed on using git branches and following development on zulip, if you want to test this.

hjortling:
Hi
@ps How is this feature different from Microboluses #1205 feature branch?. 
Do they do the same thing or do they complement each other?
Will you need to choose one over the other?
I am currently testing the #1205. branch, which works fine, but this also sounds promising, and I will  try it out also.

ps:
Don't know, was not involved with this.

Best,
@ps

jbwittmer:
@ps2 Thanks for the answer to my question.  Hope I didn't open up a can of worms by asking.  With respect to our particular situation, I think our settings are pretty rock solid and we have seen increased TIR without any increase in lows.  Our particular circumstance pertains to a teenage boy who not infrequently "forgets" to treat/account for carbs.  In that setting would be nice to get the corrective insulin on board a little faster and so long as our other settings are accurate I'm hopeful that we won't see an increase in later lows.  I'll watch closely for sure after any changes.  This ability to make individual customization to meet the needs of an individual person with diabetes is a huge benefit of the DIY solutions in my opinion, but comes with the potential to really cause adverse effects if not applied carefully and with proper monitoring.  So to others I would echo the cautions Pete has voiced above.  Thanks again.

scottleibrand:
@hjortling's question was for @ps2 not @ps

Nightfoxy:
I think I'll up the partial application to 0.5. I am seeing a number periods where the bolus recommendation for my 6 year old is 0.1u but nothing is given since 0.05 (smallest omnipod dose) is greater than 40%.

I am seeing tighter SD and highs being cut down without lows increasing. Loving it!

dyjaks:
@ps2 Was this UI bug fixed in `e5dbed1`? I just updated today and maybe it didn't push to my phone? I swear I would have already had this in my last pull anyways.

unnamed

ps2:
@dyjaks yes, that was fixed in https://github.com/LoopKit/Loop/pull/1219/commits/e5dbed130ae5ed9eaf5b08b448399854efceb401

sthrthxs5:
I realize this is more of a NightScout issue but we are going to need a different way to visualize the micro boluses in NS.  Right now, if there are boluses every loop due to a high BG and incorrect CoB, they all sit on top of each other and it becomes unreadable very quickly.
Annotation 2020-01-10 093850

francesc0-cgm:
This issue is the same in active insulin HUD in Loop app too. Any chance to get automated bolus tagged differently in both Loop and NS?

Inviato da iPhone

ps2:
As discussed in Zulip, we are going to be tracking a couple new fields on boluses throughout Loop; whether they were issued via Loop or via pump, and whether it was an automatic dose, or a (loop) manual dose.

francesc0-cgm:
Great, thank you!

Inviato da iPhone

straykatz:
This is interesting. And yes, please add Max IOB. 
Not having max IOB is a reason I don’t feel as comfortable with LOOP as I do with OpenAPS. I set Max IOB to 6, and I don’t have many situations where I hit it. This without entering carbs and also a random diet that does include high carb meals.

sthrthxs5:
Yep, I agree on the MaxIoB. Been running this branch with my teenager for the last few days. It’s done much better at keeping her in range but we have had two significant, below 50, lows since switching which we normally only see once every few months. Being a typical teenager she is bad about counting carbs let alone entering and bolusing for them much of the time.  We know some of her ratios are off and we are addressing that but the IoB in both lows were above 7 and her BG was below 225 which would normally not happen without knowing for sure she has enough CoB to match.   I’m fairly confident our problem is three things. 1) her avg is down from 200 to around 150 which is affecting her ISF. 2) the microboluses are much more effective than the slow drip of a temp basal which combined with FIASP is causing the sudden drops. 3) her carb ratios are off because she was bad at entering accurate carb counts/types in the past. Don’t get me wrong, this has moved the needle the most since switching to loop, we just need some guardrails.

erikdi:
In terms of battery life and Nightscout UI:

I have been running with a maximum frequency of every 15 minutes for automatic bolus, see the logic here (my own code, not related to this pull request):
https://github.com/erikdi/Loop/blob/dev/Loop/Managers/LoopDataManager.swift#L1334

The other thing the code makes sure is that there is a blackout period of 2 minutes after a new carb entry. Previously we had ran into situations were an accidental carb entry was immediately bolus'ed due to the automation.  This way you can also add multiple carb entries and get only one bolus issued. Useful if you have a drink, something to eat, dessert, and don't want your kid add everything together in their head.  And you don't have to wait for the previous bolus to finish. Minimizes interaction time with the app.

antonio:
Something I've noticed (apart from delivery being resumed while I was in the shower and the pump was suspended, cc https://github.com/LoopKit/Loop/issues/1236) is that bolus are dosed even when the reservoir is empty: I was walking back home after breakfast when I ran out of insulin, and in the time it took me to get home Loop had already issued 3 bolus.

scottleibrand:
On Medtronic pumps, an “empty” reservoir isn’t really empty (much like your car’s gas gauge reading empty). There are about 8U of volume left in the reservoir before the pump gives a No Delivery alarm. As you get past about 5U of delivery on empty, the plunger starts to hit the angled sides of the end of the reservoir, and the volume delivered starts to be less and less relative to the amount the pump thinks it’s delivering. Also, air bubbles in the reservoir are most likely to be pumped into the tubing after the reservoir reads empty. For both these reasons, you can expect that IOB based on pump history will read higher than what was actually delivered as you get past about 5U of delivery on empty.

The above conclusions were informed by sitting at the sink and counting drops of saline coming out of a disconnected reservoir when repeatedly issuing equal-sized microboluses, and subsequently by using OpenAPS SMB with reservoirs reading empty.  

I haven’t used or closely reviewed Loop’s microbolus code, and am not sure what if anything it does with reservoir level checks, but based on the described behavior I believe it is doing the correct thing by continuing to microbolus after the reservoir reads empty.

antonio:
> but based on the described behavior I believe it is doing the correct thing by continuing to microbolus after the reservoir reads empty.

Sounds like it indeed! Thanks for the informative answer @scottleibrand

Loop/Models/LoopError.swift
```diff
@@ -130,7 +133,8 @@ extension LoopError: LocalizedError {
             return String(format: NSLocalizedString("Recommendation expired: %1$@ old", comment: "The error message when a recommendation has expired. (1: age of recommendation in minutes)"), minutes)
         case .invalidData(let details):
             return String(format: NSLocalizedString("Invalid data: %1$@", comment: "The error message when invalid data was encountered. (1: details of invalid data)"), details)
-
+        case .pumpSuspended:
```
sraka1: Typo: encountered -> suspended

straykatz:
So far, so good. I have been running this branch for about a week now, without problems. I changed partial application factor to 0.5, and am slowly moving toward "normal" targets.  I have seen good results without announcing carbs, playing around with changing post-meal sensitivities for up to 4 hours. 

Running this with Omnipod and a 1:1 dilution of FIASP (100% of that right now), first using LR solution, now Ely Lilly's Diluent. Normally I mix FIAPS and Novolog 80:20 for pump use, but I am trying to NOT do that, since I am diluting with diluent, and so far I am not seeing the local resistance issues I have seen before. 
Using a new Nightscout setup for this as a test, but I might move it into my regular Nightscout, as I am going to keep diluting insulin, even when using the Medtronic/OpenAPS setup. I am down to using about 30 - 35 units/day (with the Medtronic pump). WIth the pod I used up the 200 units in 3 days - which is about 33/day undiluted.

Thank you for this branch, I like having the option using Omnipod, and this branch appears to require less interactions for meals and acceptable BGs than I have experienced when I ran LOOP previously. 

I would very much like to see Max IOB settings.

Nightfoxy:
The rounding to the nearest deliverable pulse is smoothing out nights for my 6 year old. Thanks!

The main issue we ran into is my daughter used the watch to enter carbs, waited a little to verify the bolus amount with one of us, then delivered. In between the carb and bolus confirmation, Loop auto bolused 40% of the insulin and finished. Then my daughter's 100% bolus from the watch went through, delivering 140% of the recommended bolus. 

No harm since we noticed when she wanted more food and it did not suggest a bolus. 

The biggest setting change I made is to bump ISF closer to tested levels. I had it lower to get Loop acting like I wanted with TBRs.

ps2:
@straykatz Keeping the partial application at 40% will help contribute to testing for this branch. If you test at 40% and are finding issues with it, after you've adjusted all the other things you think pertain, then report them here, or in zulip, and that will help this branch get closer to acceptance.

ps2:
@Nightfoxy This is actually a long-standing issue in Loop, that auto-bolusing makes a little more visible. Loop would enact temp basals that should invalidate any outstanding bolus recommendations, but wasn't doing that. Now, we have the same problem, but with boluses.  Micheal's carb input branch fixes this problem for carb entry and bolusing on the phone, but we need to extend it to the watch as well. Thanks for the feedback.

Nightfoxy:
> @Nightfoxy This is actually a long-standing issue in Loop, that auto-bolusing makes a little more visible. 
I should have added that I was aware of this risk already. Just calling it out for others as it sort of mattered this time.
Happy to see the double bolus issue is fixed too (again, not auto bolus specific). I figured out how to get a sysdiagnose log because of it, but you had fixed it by the time I realized I didn't need it.

This a pretty awesome change. Its also very intuitive to understand how loop adds more insulin and takes it away, especially with overrides. You can see the higher basal rate being enacted in NS along with boluses. And, I love that a high temp basal can't get stuck running anymore.

andrew-dixon:
@sthrthxs5 I have raised https://github.com/nightscout/cgm-remote-monitor/issues/5514 on the NS project and will see if I can find some time to look at making this change and submitting a pull request to them for it.

PieterGit:
@vtourret You can use GitHub's compare function to find this out. 
This PR is based on the `dev` branch and only missing 
- Support enacting manual loop if loop error (PR 1252) and 
- Denote stale glucose as three dashes (PR 1275)

You can see the differences on https://github.com/LoopKit/Loop/compare/automatic-bolus...dev
These changes will be available on this brnach if @ps2 updates this branch.

The current `dev` is quite similar to the `master` branch (2.0) but it's a development branch, see 
https://github.com/LoopKit/Loop/compare/dev for differences.

PieterGit:
We use this branch for several weeks now. It works good. Automatic bolus dosing works very well if there is not much activity (walking, doing stuff, etc.) in future, but causes some lows in cases of higher IOB and future activity. Without automatic bolus dosing and future activity these lows do not occur. So our current strategy is to enable automatic bolus dosing when there is not physical activity scheduled for the next say 1,5 hour (e.g. in the evening after walking the dog or before going to sleep when there is a high BG). So perhaps these settings must be a bit more conservative, or we should implement some Dose Strategy Automatic Switching as is suggested in: https://github.com/LoopKit/Loop/pull/1283

mlaatu:
We have been using this branch for some weeks now (the user is my 12 year old). Switched from Ivan's branch, which was more aggressive and quite good with those stubborn highs. But I feel this branch is safer so I'm happy with the switch, although we now need to use fake carbs every now and then. What is a bit annoying is the recurrent beeping. There is an option to switch them off, but when we tried that, there were some failed boluses because she didn't realize the entered bolus didn't go through. So we turned the beeps back on, but then the beeping can be quite disturbing, at least when she returns to school. There could be different ways to decrease the number of beeping: 1) by making a feature that silences the beeps if the bolus size is very small (like they often are in automatic boluses) 2) by making a feature that silences all the automatic bolus beeps 3) by making a feature the stops the automatic boluses that are smaller than maybe 0,2U, and only gives the bolus (and the beep) when the insulin need has accumulated to be bigger than this.
Thank you for your work with this branch!

Nightfoxy:
The major aggressiveness difference is from the adaptive nonlinear carb
model as the default in Ivan's. You should troubleshoot your carb ratios
and absorption times you are using and understand how that model is
different than the nonlinear model that's default in Loop.

Pete has said there is work pending to add some low level data elements
that let tag boluses as automatically or manually delivered. When that is
possible, it will be possible to have beeps for manual boluses and not
automatic ones.

jmphillips33:
We have been using this branch since the beginning of January. The only issue that we have seen is autoboluses being given following a "compression low." Sometimes because the swing appears drastic the boluses are quite aggressive so they lead to actual lows. Not sure if this can be mitigated. Thankfully this doesn't come up often, but would love to hear other's thoughts on this.
Thank you for this branch Pete!

macbenoit:
Hello 
Is there a possibility to implement autobolus with spike ?

macbenoit:
Hello 
Is there a possibility to implement autobolus with spike ?

Nightfoxy:
Search the Looped group for "autobolus spike" and you'll see someone has
put that together.
https://github.com/Andrei0105/Loop/tree/spike-autobolus

andrew-dixon:
@ps2 How does an override affect autobolus, specifically if the override is turned on after carbs are entered. For example, if I enter 50g carbs with 4 hour absorption time and then after 1-hour turn on an override to increase by 20% for 1 hour are the auto boluses given during that 1 hour to cover the carbs increased by 20%? If so, what then happens in hours 3 and 4 as it will see IOB higher than it would have seen without the increase. Hope that makes sense, thanks.

PieterGit:
Pete, can you explain what need's to be done/tested/fixed before this branch can be merged to dev?

ps2:
@PieterGit A few things: 
1) The changes here need to be merged with all the changes that happened on dev recently, mainly the Tidepool merges.
2) There are some model changes needed to track whether a bolus was delivered automatically or manually. 
3) The model changes are slated to be done after the Dose logging changes are landed, as those involve model changes as well, and dealing with conflicts in model changes is messy.

PieterGit:
Sinc I overlooked the Zulip message of @ps2, I'll repeat it for others that are mainly reading Github and not the Zulip channel:

Automatic Bolus dosing strategy is now an option in `dev`, alongside the traditional temp basal strategy (which is still default). Appreciate any testing. Please remember this is a fairly unstable branch, and you should probably only be using it on test/development setups. Any issues, please report here on zulip or file github issues. 

Source: https://loop.zulipchat.com/#narrow/stream/144182-development/topic/Automatic.20Bolus.20on.20dev 

So I assume the `automatic-bolus` branch from this PR is now obsolete and you should test `dev` branch for testing Automatic bolus dosing.  

@ps2: perhaps you can close these branches/PR's to be clear about the state of these branches/PR's?

* track-automatic-bolus Updated 2 days ago by ps2 #1480 merged
* automatic-bolus-merge Updated 3 days ago by ps2 #1478 merged
* automatic-bolus Updated 14 days ago by ps2  #1219 (not merged directly, but merged with #1480 and #1478 I suppose)

I will do some testing and report with Github issues or in Zulip chat.

ps2:
The automatic-bolus branch this PR is tracking has de-facto become the main branch for anyone wanting to actually run automatic-bolus in Loop despite being what I would call an "experimental branch".  It is still an experimental branch, but it's fairly stable, and is very in sync with the current Loop release, 2.2.4.   The dev branch is still unstable so this branch will stay until there is an actual new release. The two other dev branches, `track-automatic-bolus` and `automatic-bolus-merge` will be deleted soon.

juan-restrepo:
Are there plans to sync this branch with 2.2.5 or 2.2.6?

jonfawcett:
@juan-restrepo I’m planning to submit as a PR to Dev. If you want it on master branch, see my loop patches repo

juan-restrepo:
Thanks @jonfawcett. I'll check that out next time if my wife or kid are very adamant about having the latest version + autobolus (although my kid is <3, so I guess he won't have an opinion for a few years).

andrew-dixon:
@juan-restrepo @jonfawcett Not sure if this is the same thing, but Automatic Bolus was added to master in release 2.2.5 - https://github.com/LoopKit/Loop/releases/tag/v2.2.5

jonfawcett:
> @juan-restrepo @jonfawcett Not sure if this is the same thing, but Automatic Bolus was added to master in release 2.2.5 - https://github.com/LoopKit/Loop/releases/tag/v2.2.5

My mistake. I have been getting so many tags about my dose strategy switcher that I mistakenly thought this was another one.

marionbarker:
I believe this PR should be closed as overtaken by events.  This dosing strategy was added to master with release Loop v2.2.5.

juan-restrepo:
I ended up having some issues and upon converging to master 2.2.6 I had the great surprise of having autobolus there. Thanks!

PR #1232, opened 2020-01-12
upload symbols to nightscout

CodeByMini:
This pull request allows the Lollipop, Taco and Pizza symbols to be automatically uploaded to Nightscout.
The reason for this is to have a quick view of what kind of carbs that were entered instead of having to hover over the treatment in Nightscout.

The code is kind of a work around and could definitely be improved, but it does the job.
A known bug is that if the user selects different food icons they will not be uploaded if the meal time is 2, 3 or 4 hours. It also has to be edited if the user has choosen different carb absoprtion times for .fast .medium. slow

I would love to see a more experienced programmer commit to this


loyalchallenger:
This already happens on my nightscout. I can send you a screenshot.

CaspianGeben:
> This already happens on my nightscout. I can send you a screenshot.

Show a screenshot on how you add carbs
This take the three symbols from the first screen and uploads that symbol. I am not using the symbols you can choose after selecting the plate.

loyalchallenger:
There is no option for me to upload screenshots here. 

I’m not sure what symbols you’re using then. I’m using the symbols provided in loop. I choose them when I enter carbs and they show up in my nightscout.

CaspianGeben:
> There is no option for me to upload screenshots here.
> 
> I’m not sure what symbols you’re using then. I’m using the symbols provided in loop. I choose them when I enter carbs and they show up in my nightscout.

If you use an apple watch do you see the symbols in nightscout then?

CaspianGeben:
if you choose one of these three symbols (circled in red) without clicking the plate (crossed over) and choosing the same symbols from that meny, does that still show up in nightscout?
IMG_9891

loyalchallenger:
I see what you’re saying. You want a default option to upload to nightscout. I do not use the Apple Watch.

rolandet:
The symbols automatically upload for me with Loop v2.0.

Nightfoxy:
Watch entries do not add the symbols.

CaspianGeben:
> The symbols automatically upload for me with Loop v2.0.
> 
> A5898ADC-C505-49F3-8666-CC05C1574D27

Yes. because you choose them under the plate. if you just choose the lollipop, taco or pizza on the screenshot I posted you will not get the symbols to Nightscout

ps2:
I believe this functionality is working in current Loop.

PR #1234, opened 2020-01-12
Set carb cache length to 24 hours.

ps2:
Investigations into delayed display and dosing availability have shown significant cache misses in CarbStore, which results in a HealthKit query.  When the phone has been locked for a significant amount of time, HealthKit queries can perform very slowly, as HealthKit is processing journaled data and doing other housekeeping. This fix increases the cache length to a longer value that should eliminate most cache misses.

### DIFF
--- a/Loop/Managers/LoopDataManager.swift
+++ b/Loop/Managers/LoopDataManager.swift
@@ -65,6 +65,7 @@ final class LoopDataManager {
         carbStore = CarbStore(
             healthStore: healthStore,
             cacheStore: cacheStore,
+            cacheLength: .hours(24),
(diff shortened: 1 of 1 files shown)


marionbarker:
I have run for 2 days with the PR 1234 (24 hours carb store) added to PR 1219 (automated-bolus branch).  I believe that I am getting prompt bolus suggestions now when phone has been locked and I wake it up and enter carbs right away.
https://loop.zulipchat.com/#narrow/stream/144182-development/topic/Automatic.20Bolus.20Experimental.20Branch.20%28PR.201219%29/near/185453886

Issue #653, opened 2017-12-29
Suspend threshold uses predicted BGs but not current BG


elnjensen:
When determining whether or not to recommend a bolus for carbs entered, Loop appears to be ignoring the current BG value and only using the array of predicted BG values.  This can lead to unexpected behavior when the current BG is slightly below the suspend threshold and carbs are entered.  If all of the predicted BG values are above the threshold, Loop will recommend a bolus even if the current BG is below the threshold. 

Steps to reproduce: 
1. Wait until BG is flat or rising. 
2. Set suspend threshold to a value just slightly above the current BG (e.g. just 1 mg/dl or 0.1 mmol/L larger). 
3. Enter a large number of carbs.  Typically this will result in a bolus recommendation even though current BG is below threshold.  (If it doesn't, scroll over the glucose prediction graph and you'll see that the first predicted value is above the threshold you set.) 
4.  Delete the fake carbs and set your suspend threshold back to a normal value!  :smile:

This is admittedly an edge case, but nonetheless doesn't match the documented function of the suspend threshold.  I tried to follow the logic through the code, but I had a hard time figuring out where the current BG value is stored.  There are some comments (e.g. [this](https://github.com/LoopKit/Loop/blob/master/Loop/Managers/DoseMath.swift#L440)) that imply that it is included as the first element in the array of predicted BG values, but if that were the case, then I think the code should work as-is. 

[This](https://github.com/LoopKit/Loop/blob/master/Loop/Managers/DoseMath.swift#L250) seems like possibly where the current BG needs to be considered if it is not in fact part of the `prediction` array.

kenstack:
@elnjensen I was looking at the same line but for other reasons.  I am completely new to loop but in discussions with @Kdisimone my understanding is that the line in question is serving more to prevent excessive lows from interim BGs from bolus / temp basals.  Ie if the math says that based on your required BG correction give say 3 units to end up at target in 5-6 hours, but 3 units in 2 hours based on older IOB would put you below that threshold, then it keeps reducing the bolus or basal until that threshold isnt violated.  It is not a true threshold BG suspend (based on actual), but that is only my understanding and those closer to it should correct me.  

[...]

PR #1235, opened 2020-01-12
Include current bg for suspend threshold and dosing thresholds

ps2:
Suspend threshold evaluation (when basal should be cut off), and dosing threshold (when any dosing above basal should be prevented), were not including current bg in the evaluation. This moves the window of consideration (DIA duration) back to the point of the most recent bg, rather than the current time.


ps2:
fixes https://github.com/LoopKit/Loop/issues/653

PR #1238, opened 2020-01-14
Include insulin model delay in computation of dose

dm61:
A 10-min delay is included in calculation of effected insulin so that the insulin model used in dose calculations becomes consistent with the insulin model used for prediction calculations. 

This addresses an issue brought up in the [Partial Bolus Recommendation](https://loop.zulipchat.com/#narrow/stream/148543-Loop/topic/Partial.20Bolus.20Recommendations) topic on Loop Zulip where a nonzero bolus recommendation is observed immediately after a bolus recommendation has been accepted.  DoseMathTests have been updated accordingly.

### DIFF
--- a/Loop/Managers/DoseMath.swift
+++ b/Loop/Managers/DoseMath.swift
@@ -236,6 +236,7 @@ extension Collection where Element: GlucoseValue {
         let unit = correctionRange.unit
         let sensitivityValue = sensitivity.doubleValue(for: unit)
         let suspendThresholdValue = suspendThreshold.doubleValue(for: unit)
+        let delay = TimeInterval(minutes: 10)
 
         // For each prediction above target, determine the amount of insulin necessary to correct glucose based on the modeled effectiveness of the insulin at that time
         for prediction in self {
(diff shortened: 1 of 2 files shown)


Loop/Managers/DoseMath.swift
```diff
@@ -236,6 +236,7 @@ extension Collection where Element: GlucoseValue {
         let unit = correctionRange.unit
         let sensitivityValue = sensitivity.doubleValue(for: unit)
         let suspendThresholdValue = suspendThreshold.doubleValue(for: unit)
+        let delay = TimeInterval(minutes: 10)
```
ps2: This value should not be hard-coded. We are also using a default value (of 10m) in LoopKit; if one changes, so should the other, so they should be controlled by a settings value, or at least a single value specified in one place.

dm61: Yup, I was going to look into how to get that value from LoopKit, but then decided to just put this out so you can quickly take a look. 

I'd prefer including the delay in the insulin model, where it really belongs (but I'm not sure I know how to do that correctly - will take a look and get back to you in the next few days).

PR #1242, opened 2020-01-18
Insulin model delay

dm61:
Adds delay to the insulin model presets, reverts hard coded delay and subtraction of delay in DoseMath. This is a companion PR to LoopKit PR https://github.com/LoopKit/LoopKit/pull/315. Upon review, and any further updates, the two PR's should be merged concurrently. This Loop PR and the LoopKit PR 315 serve the same main purpose of fixing the dose math as https://github.com/LoopKit/Loop/pull/1238, which should be closed once these two PR's are merged.

### DIFF
--- a/LoopCore/Insulin/ExponentialInsulinModelPreset.swift
+++ b/LoopCore/Insulin/ExponentialInsulinModelPreset.swift
@@ -38,9 +38,20 @@ extension ExponentialInsulinModelPreset {
             return .minutes(55)
         }
     }
+    
+    var effectDelay: TimeInterval {
+        switch self {
+        case .humalogNovologAdult:
+            return .minutes(10)
+        case .humalogNovologChild:
+            return .minutes(10)
+        case .fiasp:
+            return .minutes(10)
+        }
+    }
 
     var model: InsulinModel {
-        return ExponentialInsulinModel(actionDuration: actionDuration, peakActivityTime: peakActivity)
+        return ExponentialInsulinModel(actionDuration: actionDuration, peakActivityTime: peakActivity, delay: effectDelay)
     }
 }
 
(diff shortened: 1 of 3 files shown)


LoopCore/Insulin/ExponentialInsulinModelPreset.swift
```diff
@@ -38,9 +38,20 @@ extension ExponentialInsulinModelPreset {
             return .minutes(55)
         }
     }
+    
+    var effectDelay: TimeInterval {
```
sraka1: @dm61 ideally, this should be user configurable (due to YDMV) IMHO. In many cases 15 or even 20 minutes might be more adequate (without the user needing to make a code change).

dm61: Thanks for the comment. The main purpose of this PR is to address a slight inconsistency in dosing versus prediction algorithms, as discussed in  https://github.com/LoopKit/Loop/pull/1238. UI-related decisions and design are separate issues, beyond the scope of this particular PR.

Trpl7ca:
is it possible to merge this into automatic-bolus branch?

ps2:
Thanks @dm61. Looks great!

sarisarisari:
Can I please clarify that this means that the algorhythm will now assume that the insulin doesn't begin it's function until 10 min after it's been programmed? Where previously the algorhythm assumed immediate action?

ps2:
The main algorithm has always assumed a 10 minute delay, and you would see this in the forecast, IOB charts, etc.  The fix here is how dosing is calculated.  Loop iterates to see what size dose will correct any highs without bringing you below range. That calculation wasn't using the 10 minute delay, so doses could be smaller, depending on the timing of carbs and your current bg.

sarisarisari:
Oh interesting.  Ok. Thank you for the clarification.

Issue #1243, opened 2020-01-19
Basal Line Offset (& Negative)


kparisella:
**Describe the bug***
I’ve had the scheduled-basal baseline in the Insulin Delivery graph temporarily get shifted downward, so that scheduled basals (not temp basals) appear negative (which is, of course, impossible).

This happened this morning for at least an hour, while using Automatic-Bolus branch, but Pete said this has been an issue in other branches as well.

**Attach an Issue Report**
Attached Issue Report 

[Issue report ](https://documentcloud.adobe.com/link/track?uri=urn%3Aaaid%3Ascds%3AUS%3A394d2ab4-7c93-4111-996c-78a293f4af1c) 

**To Reproduce**
Unfortunately, I have no idea how it happened, but I’ve seen it a few times that I can remember, with different branches.

This morning, I had activated a 200% override before I noticed it.  I subsequently de-activated the override, but that didn’t make the basal offset go away, even after a few cycles.  It eventually resolved itself, but I’m not sure when/why.

**Expected behavior**
I expected the thin line on across the Insulin Delivery graph to represent my regularly scheduled basal rate, which was (positive) 0.6 u/h at the time.  

**Screenshots**
Attached a few screenshots while it was happening, plus screenshot of basal rates.

![AF5AAD56-C53E-41A8-9587-DDB7186FE6B2](https://user-images.githubusercontent.com/38840696/72689840-d55f4800-3ae3-11ea-8773-e59d5914ea32.png)
![38E6DC5D-D3CC-441E-B459-03F94D16E2D7](https://user-images.githubusercontent.com/38840696/72689841-d55f4800-3ae3-11ea-91c3-57f735a92e0c.png)
![358C9974-1224-4780-AD50-9D32BDB4F814](https://user-images.githubusercontent.com/38840696/72689842-d55f4800-3ae3-11ea-9d10-0abce40301d2.png)
![D7D6F930-6DAF-4A61-B3B3-EBA798851C1C](https://user-images.githubusercontent.com/38840696/72689843-d5f7de80-3ae3-11ea-915e-4fdacbd4014c.png)

**Phone**
 - Hardware: iPhone 8 plus
 - OS Version: iOS 13.3

**Loop Version**
 - Version Number: 1.10.3
 - Repo: Automatic-Bolus from Workspace

**CGM**
 - Device: Dexcom G6
 - Dexcom App

**Pump**
 - Medtronic
 - Model: 522
 - Firmware version: 2.4A

[...]

Issue #1113, opened 2019-09-17
Insulin Delivery Graph Columns not aligned with Graph


smartdigger:
**Describe the bug**
Insulin Delivery Columns display to the left Y Axis of the Graph. This is on the latest dev release (commited 2 days ago)
Device rebooted and app closed and reopened.

See attached screenshot.

**To Reproduce**
1. No Steps - Just a display issue 

**Expected behavior**
Insulin Delivery Columns to the right of the Y Axis

**Screenshots**

**Phone**
 - Hardware: iPhone XS]
 - OS Version: iOS 12.4.1]

**Loop Version**
 - Version Number: Dev v1.10.1
 - Repo: [Loop/Dev Branch]

**CGM**
 - Device: Dexcom G6
 - Manager app: Dexcom App

**Pump**
 - Omnipod

[Loop Report 2019-09-17 202327+0100.md.zip](https://github.com/LoopKit/Loop/files/3623100/Loop.Report.2019-09-17.202327%2B0100.md.zip)
![UNADJUSTEDNONRAW_thumb_26e](https://user-images.githubusercontent.com/47868089/65073346-9a5cd480-d98a-11e9-88d3-40d14bbfc9ed.jpg)

smartdigger:
In Landscape View it works as expected. Issue only occurs in Protrait Mode
![IMG_0628](https://user-images.githubusercontent.com/47868089/65144368-231d5400-da0f-11e9-82cf-ba17383d9998.PNG)

PR #1245, opened 2020-01-20
Fix interval multiple for dose chart y axis

ps2:
Fix for #1243 

Chart lines for dose chart were being created a non-integer intervals, but the labels were being rounded to integers.


ps2:
Adding fix for https://github.com/LoopKit/Loop/issues/1113 as well

PR #1273, opened 2020-02-23
Fix bolus screen memory leak

nathangiovannini3:
The old 'Swift doesn't warn you about implicit self when passing a member function point-free' gotcha.

https://github.com/tidepool-org/Loop/pull/57


ps2:
Thanks for submitting. This is covered by https://github.com/LoopKit/Loop/pull/1277

PR #1274, opened 2020-02-24
Fix nightscout uploading trend arrow inconsistency

dabear:
Issue reported on zulipchat: https://loop.zulipchat.com/#narrow/stream/144182-development/topic/trend.20arrow.20inconsistency

The code for uploading to nightscout has been inside Loop for a very long time, but has up to now only been used by the medtronic enlite cgmmanager. Now that the Dexcom cgmmanager also can upload to loop, more users have started using this feature. 

I've detected that the loop trend arrow doesn't always match the trend arrow displayed in nightscout. This fixes that by uploading the correct trend arrow names, adding those that were lacking from the original implementation. 

(My guess is that Loop's code for uploading glucose trend arrows predates nightscout's support for granular trend arrows. This code should be safe because Nightscout users generally update fast and because nightscout has supported granular trend arrows for a long time now, meaning the risk here is quite low.)

### DIFF
--- a/Loop/Managers/NightscoutDataManager.swift
+++ b/Loop/Managers/NightscoutDataManager.swift
@@ -417,15 +417,20 @@ final class NightscoutDataManager {
         } else {
             deviceStr = "loop://unknowndevice"
         }
+
         let direction: String? = {
             switch sensorState?.trendType {
             case .up?:
+                return "FortyFiveUp"
+            case .upUp?:
                 return "SingleUp"
-            case .upUp?, .upUpUp?:
+            case .upUpUp?:
                 return "DoubleUp"
             case .down?:
+                return "FortyFiveDown"
+            case .downDown?:
                 return "SingleDown"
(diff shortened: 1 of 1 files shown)


dabear:
On the nightscout repo, see which trend arrows are supported here: https://github.com/nightscout/cgm-remote-monitor/blob/7c272ec67a7795e0f4edf7ff9abdfac1bd1308a0/tests/direction.test.js#L69

dabear:
@ps2 this is a small change. Tested successfully. Should be ready to merge

pakritz:
We have Loop Master Branch 2.0.  We started looping in March and have not added any features or changed branches.  We have Loop Master 2.0/Ominpod/Dexcom G6/iPhone 11 Pro iOS 13.4.1.   I am not a technical person and make a very honest effort to watch Kate's Youtube videos and search groups to find information as well as other FB groups that post helpful videos.  The learning curve is steep but I am determined.  I would greatly appreciate input or direction!

PR #1283, opened 2020-03-06
Dose Strategy Automatic Switching

jonfawcett:
This adds settings to have Loop automatically switch to AB when glucose is over a specific threshold and switch to TB when under the threshold.


Trpl7ca:
I think this is what the FreeAPS branch does, when rising BG cannot be controlled by scheduled basal or a selectable factor of basil then MB will take over.

miki3421:
I tried on my lab and when I have in both section the Dosing Strategy ( Configuration and Dosing Strategy ) and when I'm selecting and I'm back to menu the app is Crashing.

jonfawcett:
I see that issue and will get it resolved. Hadn’t encountered the crash because we never went back into that screen after setting this up...had just let Loop do the switching. It should be an easy fix in the loop settings view controller and I’ll find it to resolve.

miki3421:
Amazing! Let me know when you fixed. Anyway from my experience I saw that
the best way to switch is not based on Glucose Value but on the trends. For
example I'm keeping Auto Bolus when the glucose is raising during the first
phase of the digestion and I'm switching to Temp Basal when the glucose is
start going down in that case you can control in better way the long phase
of the digestion. What do you think?

Il giorno ven 6 mar 2020 alle ore 21:10 jonfawcett <notifications@github.com>
ha scritto:

jonfawcett:
@miki3421 I think that issue should be solved now. I had some screwy issues with git the last 2 days which caused me to move from one local repo to another and I discovered that not all of my changes made it over. I have been thinking of various times when you would want to switch not just based on BG alone but wanted to make it simple to start. I think detecting a noisy sensor is another example...you'd want to switch to TB dependent on how large the deltas are combined with the bg range it is happening in.

Trpl7ca:
Sorry i see this is different

PieterGit:
@jonfawcett Thanks for this effort. Currently I think for us (my T1D and me) we should find a way to use TB in time, to prevent high IOB in combination of future activity in future. Perhaps switching to TB if eventual BG is below a certain threshold and switching to AB if current glucose is above a threshold AND eventual BG is expected to be well above target BG. See https://github.com/LoopKit/Loop/pull/1219#issuecomment-596192170

I noticed @stevebell117 started a similar PR for OpenAPS / oref0, see https://github.com/openaps/oref0/pull/1365 . Perhaps we can work together to find the best SMB/AB strategy.

jonfawcett:
@PieterGit I am not sure that I understand this "AND eventual BG is expected to be well above target BG". The inverse of this maybe explains what I don't understand...If Bg is higher but expected to be at or below target, it will not be issuing automatic boluses or high temp basal. So the last part of your comment seems like it wouldn't be needed in code because that is the only time Loop will enact them anyway.

miki3421:
@Jonfawcett, It's been two days that i'm using the Auto switching feature
and it's working perfectly. The only suggestion if is possibile showing
some where with mode is on.
thanks for your effort.
Michele


Il giorno lun 9 mar 2020 alle ore 16:02 jonfawcett <notifications@github.com>
ha scritto:

jonfawcett:
@miki3421 Glad it's working well for you. FYI I wrote it to toggle the underlying Loop strategy setting, so the settings screen will always reflect which method is currently running.

jonfawcett:
I just pushed a fix that should solve this working for MMOL. Can someone on MMOL confirm that it is automatically switching now?

culdeus:
I would look for potential interactions where
- BG crosses threshold while AB is in progress.  (and reverse, although less priority)
- BG crosses threshold while red loop is in progress, then returns prior to green loop.
- Too frequent switching where right at the threshold level, could there be "missed clicks".  

Safety could be maintained by only allowing a switch every so often (30 min) and switching only when green loop is established at that moment.

jonfawcett:
The switching code is in the same function of the Loop cycle that calculates the dose to apply for AB or the new TBR to apply. It takes place immediately before the dose is calculated. So the first 2 should not be an issue with the switching mechanism because of this.

For missed clicks, I had thought about this also. I don’t think it’s an issue, but want to investigate a little more.

robsyl:
Should I be updating to this build? Are people seeing improvements in blood sugar stability?

jonfawcett:
@robsyl it is up to you if you want to try my code. I would only encourage you to try it if you’re following along the main AB branch in Zulip and read through my commits to understand what it’s doing. I plan to keep it updated when AB has updates but can’t guarantee it will be immediate unless it’s a commit that I feel we need to build immediately. I built it to solve an issue we specifically had with AB. When lower in normal range, it could be too aggressive causing more lows. We originally bumped up the target by 10mg/dL to account for this but that meant a little higher the rest of the time. With this code, we’ve gone back down with the target range and aren’t having the extra lows from AB.

Trpl7ca:
So Jon, the FreeAPS / old micro bolus firm Ivan uses your scheduled basil until it cannot control the rise of BG and then will switch to MB. You can increase that factor so for example at night I will select a factor of 1.5 of my scheduled basal That way it is less aggressive in the night and a better chance of not going low.  In the day time I select a factor of 1 and then MB work immediately.  Also you can select the partial application factor so if forty is not working you can lower or raise with a sliding scale.  You can also set a TBR rate so if away from phone it will just continue like if you were swimming. That only works with open loop  Are any of these things what you are trying to correct in Auto bolus branch? What is the differences from your branch?

jonfawcett:
@Trpl7ca I don't know anything about FreeAPS. This is a branch of Loop's Automatic-Bolus early dev branch. For my daughter, TB is better than AB in lower Bg ranges but AB is better than TB when higher. So this is only adding an automated way to switch between the 2 to achieve better results.

Trpl7ca:
I know i have been following in Zulia, and really just wanted to mention it. This is tho old micro bolus branch by Ivan. Unfortunately he and Pete have disagreements and he went his way. But it usees Petes;s Dev a s the base and changes some of the UI, like you are doing. So as i mentioned it can treat rising BG with basal mostly with TBR or MB your choice with simple sliding scale and also change partial factor with scale and even adaptive rate option if you like. You can export your settings basil, cr,ISF, and save them and import others it your needs change, you can also just set a TBR for a duration of time with LOOP open. Not trying to sell you on it just would like someone’s like you who is smart and a coder to see if this is a viable option for folks. It has I believe too many options for Pete’s perspective for most people to grasp.

jonfawcett:
I’m somewhere between Ivan and Pete on my viewpoint about this. I do think that it is too many options for the average person which can lead to potentially dangerous combinations of settings without realizing it. But on the flip side, I think Loop needs to include more ways for altering Loop’s action regardless of the prediction. T1 is too unpredictable at times to fall back on a strategy of just figuring out why Loop’s predictions are wrong for every circumstance...there are times when you need to force the results to be different imho.

With that said...those options need to be simple for everyone to understand. My edits are along those lines. It’s not complicated to understand how it changes. The underlying math is that it changes the high temp doses between about 8% per 5 minutes with TB to 40% with AB. A simple reduction in aggressiveness. I agree with Pete that a better strategy is probably either bg-based ISF or a sliding scale of the 40% application factor. But both of those are more difficult changes than this. I am probably going to look at the difficulty of doing a sliding scale to consider that for my daughter ultimately.

sspade:
I get this impression often. OpenAPS seemed to have this worked out a bit more, but was less convenient of a system. Frankly. I would love to see some of the advances in OpenAPS adopted into loop, but feel observations like this are getting us closer to parity. 
 >  I think Loop needs to include more ways for altering Loop’s action regardless of the prediction. T1 is too unpredictable at times to fall back on a strategy of just figuring out why Loop’s predictions are wrong for every circumstance...
...
>  I am probably going to look at the difficulty of doing a sliding scale to consider that for my daughter ultimately.
Appreciate the work and perspective on this. Look forward to your commits as they become available.

Rdeboer180:
along the lines of Trpl7ca's comment. I would love the ability to only allow the switch to automatic bolus when carbs are present. I often find that i am switching to auto bolus after a meal and switching back to temp basal to preserve my phones battery and avoid low's.

jonfawcett:
> along the lines of Trpl7ca's comment. I would love the ability to only allow the switch to automatic bolus when carbs are present. I often find that i am switching to auto bolus after a meal and switching back to temp basal to preserve my phones battery and avoid low's.

You could probably easily add this. Look at line 1202 of LoopDataManager.swift. Right after the “if(“ you basically could add something like “cob > 0 &&”, you would need to figure out what the variable name is that stores carbs on board is all. The && simply means both have to be true...this new cob greater than and the bg greater than.

ps2:
Closing this. I do think we should continue the discussion about giving users a way to make loop less aggressive on the low end and more aggressive on the high end. I think there are a lot of ways to do this, but I don't feel like switching dosing strategy is the right approach.

PR #1292, opened 2020-04-01
Add build script to remove x86_64 and i386 archs from frameworks

josep1972:
This update addresses a failure when creating IPA for distribution. Specifically, when Frameworks contain sim archs (x86_64 and i386), the IPA generation process fails. This update removes the archs when Archiving and IPA generation.

There are two ways to test if the issue has been addressed:

Archive and then Validate App or
Archive and then create app for Ad-Hoc distribution.


josep1972:
New pull request is not compatible with latest LoopKit dev

ps2:
I believe with the workspace build, the architectures are being stripped correctly.

PR #1293, opened 2020-04-02
add remote bolus/carbs + OTP

josep1972:
Implementation of remote carbs and remote bolus. The two new remote commands are protected with a One Time Password generator. The remote provider must have the shared secret key to generate the same OTP as what Loop will be expecting. This also requires my changes to josep1972/cgm-remote-manager, branch: dev_otp to be deployed in Heroku.


francescaneri:
@ps2 what do you think about it?

geefrizspeng:
these added features are highly desirable in our situation, and we have used multiple branches of Loop for over 2 years now and are having great results with AB. I however am not great at troubleshooting, only good at following directions. Since we are all home I will likely give this a try soon. Thank you for all of your work!

Nightfoxy:
Would it make more sense to leverage the Apple Push Notification infrastructure and security being used with Remote Overrides? Then we wouldn't have to do 2 setups in NS?

josep1972:
@Nightfoxy: This update does exactly that. This update enhances the use of the Apple Push Notification by adding the two new commands. In order for the commands to be sent from NS, you have to update NS. In order for Loop to listen for the new commands, you have to update Loop. It's the same as when APN was first introduced for remote access. Because Bolus and Carb entries are very serious commands, this version of loop implements an additional safeguard: the one time password. 
Check out this thread: https://loop.zulipchat.com/#narrow/stream/144182-development/topic/Remote.20bolus.20.2F.20remote.20carbs

Schneiderc008:
You can find the instructions for installing and operating remote carbs/bolus here: https://docs.google.com/document/d/1_fl-fHdXd3WE8cGwmnPcd_CS-4KOSU8j1wbz4ndbTpc/edit?usp=sharing

ps2:
Remote bolus/carbs has been merged into dev.

PR #1294, opened 2020-04-02
Update HealthKit usage description

novalegra:
The HealthKit share description currently doesn't have information about why Loop is requesting permission to access for sleep data, so I added it.


ps2:
How about "Sleep data from the Health database is used to optimize delivery of Apple Watch complication updates during the time you are awake."

novalegra:
> How about "Sleep data from the Health database is used to optimize delivery of Apple Watch complication updates during the time you are awake."

Done

PR #1296, opened 2020-04-04
Add support for dose logging

novalegra:
This PR adds the ability to log outside insulin doses in Loop through the app itself, in addition to support for having multiple insulin types be active at the same time. Currently, when an outside dose is logged in HealthKit, Loop assumes that it is the same type of insulin as is currently present in the pump, which is not always the case. This change allows the user to select the insulin model if the dose is being logged; otherwise, if a bolus, basal, or temp basal is given via the pump, the current dose model is saved as a part of the dose entry. Even if the user doesn't log doses, adding insulin model information allows for correct accounting of IOB and insulin effects if the insulin type in the pump is changed (for example, from Humalog to Fiasp). The insulin model of the dose is persisted in the Loop's data store and in HealthKit, and is non-user-editable.

Loop/LoopKit tests have been updated accordingly based on the changes made, and additional tests were added with multiple insulin types active at once. Happy to add more tests if there's anything I missed.

While this PR does *not* add additional insulin curves, the underlying changes to the dose model could support new insulin types if they were added into Loop (such as Afrezza). A proposed Afrezza model is included in the code, but is disabled from a UI standpoint, pending additional testing; I can remove the Afrezza model if that would avoid confusion!

Current Testing Status:
- Has been tested with Dexcom G5 CGM, the mock pump, and Omnipod.
- Needs to be tested with Medtronic pumps.

This PR involves changes to both Loop & LoopKit; the LoopKit PR (https://github.com/LoopKit/LoopKit/pull/323) has the same title.

Screen Shot 2020-04-17 at 8 24 02 PM


Loop/Managers/LoopDataManager.swift
```diff
@@ -628,6 +630,66 @@ extension LoopDataManager {
...
+            }
+        }
+    }
+    
+    /// Logs a new external bolus insulin dose in the DoseStore and HealthKit
+    ///
+    /// - Parameters:
+    ///   - dose: The dose to be added.
+    func logOutsideInsulinDose(dose: DoseEntry, completion: @escaping (_ error: DoseStore.DoseStoreError?) -> Void) {
+        let rawData = Data(UUID().uuidString.utf8)
```
novalegra, line 671: Not sure where to get the raw data from

Loop/View Controllers/StatusTableViewController.swift
```diff
@@ -1075,12 +1075,13 @@ final class StatusTableViewController: ChartsTableViewController {
                 vc.restoreUserActivityState(activity)
             }
         case let vc as InsulinDeliveryTableViewController:
-            vc.doseStore = deviceManager.loopManager.doseStore
+            vc.deviceManager = deviceManager
```
novalegra, line 1078: This change was made so InsulinDeliveryTableViewController can create an instance of BolusViewController

Loop/View Controllers/BolusViewController.swift
```diff
@@ -572,8 +572,23 @@ final class BolusViewController: ChartsTableViewController, IdentifiableClass, U
...
         }
+        
+        // ANNA TODO: improve ~style~
+        var deliveredUnits: Double
+        
+        if let model = enteredBolusInsulinModel as? ExponentialInsulinModelPreset {
+            switch model {
+            case .afrezza:
+                // The manufacturer has said that 4 units of Afrezza is roughly the same as 3 units of rapid-acting
+                deliveredUnits = 0.75 * amount
```
sraka1, line 567: I believe this varies greatly on a per-person basis (much more than for example deviations from typical/pre-defined rapid-acting SC curve). If this model is to be included, perhaps a configurable option for potency should be included (same as the Walsh model is configurable). Just my 2 cents.

novalegra: Yes, this could vary. I think at this point, it makes the most sense to move all of the Afrezza work to a separate branch to keep the discussion focused on the underlying architecture changes. Thanks for the feedback!

Loop/Managers/LoopDataManager.swift
```diff
@@ -617,9 +617,11 @@ extension LoopDataManager {
     /// - Parameters:
     ///   - events: The pump events to add
     ///   - completion: A closure called once upon completion
+    ///   - lastReconciliation: The date that pump events were most recently reconciled against recorded pump history. Pump events are assumed to be reflective of delivery up until this point in time. If reservoir values are recorded after this time, they may be used to supplement event based delivery.
+    ///   - parameter updateLastReconciliation: Whether or not the last reconciliation time should be updated with the new value.
```
mpangburn, line 621: Same comment here as on the LoopKit PR; the reason this parameter is necessary isn't clear to me.

Loop/Managers/LoopDataManager.swift
```diff
@@ -628,6 +630,55 @@ extension LoopDataManager {
...
+            // The raw data can be empty, as the event won't be persisted
+            let pendingBasal = NewPumpEvent(date: dose.startDate, dose: dose, isMutable: true, raw: Data(), title: String(describing: dose), type: .tempBasal)
+            events.append(pendingBasal)
+        }
+
+        if let model = dose.insulinModel {
+            doseStore.longestEffectDuration = max(doseStore.longestEffectDuration, model.effectDuration)
+        }
+
+        addPumpEvents(events, lastReconciliation: nil, updateLastReconciliation: false) { (error) in
```
mpangburn, line 675: Ah, I see now. I don't know that we want to shoehorn a manual bolus in via pump event history.
It's worth creating a code path separate from `addPumpEvents`, factoring out any shared logic. That refactoring may require LoopKit changes as well, but should allow you to get rid of the `updateLastReconciliation` flag.

Loop/View Controllers/BolusViewController.swift
```diff
@@ -319,7 +367,12 @@ final class BolusViewController: ChartsTableViewController, IdentifiableClass, U
 
     private func updateDeliverButtonState() {
         let deliverText = NSLocalizedString("Deliver", comment: "The button text to initiate a bolus")
-        if potentialCarbEntry == nil {
+
+        if isLoggingDose {
+            footerView.primaryButton.setTitle(NSLocalizedString("Log Dose", comment: "The button text to log an insulin dose not given by the pump"), for: .normal)
+            footerView.primaryButton.isEnabled = enteredBolusAmount != nil && enteredBolusAmount! > 0
+            footerView.primaryButton.tintColor = enteredBolusAmount != nil && enteredBolusAmount! > 0 ? .alternateBlue : .systemBlue
```
mpangburn, line 374: Why prefer `alternateBlue` as the action button color?

Loop/View Controllers/BolusViewController.swift
```diff
@@ -24,15 +24,23 @@ final class BolusViewController: ChartsTableViewController, IdentifiableClass, U
         case chart = 0
         case carbEntry
         case notice
+        case date
+        case model
         case recommended
         case entry
     }
 
+    private var maximumDateFutureInterval = TimeInterval(hours: 48)
```
mpangburn, line 33: Can be a `let`; I also think we can shorten this interval. Maybe use the future interval for carb entry as a guide.

Loop/View Controllers/BolusViewController.swift
```diff
@@ -526,15 +620,27 @@ final class BolusViewController: ChartsTableViewController, IdentifiableClass, U
...
+                context.evaluatePolicy(.deviceOwnerAuthentication,
+                                       localizedReason: String(format: NSLocalizedString("Authenticate to Bolus %@ Units", comment: "The message displayed during a device authentication prompt for bolus specification"), amountString),
+                                       reply: { (success, error) in
+                    if success {
+                        DispatchQueue.main.async {
+                            self.setBolusAndClose(bolus)
+                        }
+                    }
+                })
+            }
```
mpangburn, line 643: The only difference between these branches is the reason string, right? You can use the branch to assign to some `localizedReason` constant and keep the authentication code shared.

Loop/View Controllers/InsulinDeliveryTableViewController.swift
```diff
@@ -8,6 +8,520 @@
...
+            style: .destructive,
+            handler: { (_) in handler() }
+        ))
+
+        addAction(UIAlertAction(
+            title: NSLocalizedString("Cancel", comment: "The title of the cancel action in an action sheet"),
+            style: .cancel
+        ))
+    }
+}
```
mpangburn, line 524: Didn't review this file extensively; I'm assuming it's copied from LoopKit aside from the instantiation of the bolus view controller near the top of the file.

Loop/View Controllers/BolusViewController.swift
novalegra: No good reason, removed.

Loop/View Controllers/BolusViewController.swift
novalegra: Makes sense. This PR was motivated by a desire to allow for long-acting insulin to be logged (still a WIP), which is the reason for the long interval, but in this context the carb entry duration would be better.

Loop/View Controllers/InsulinDeliveryTableViewController.swift
novalegra: Yes, there were only minor changes (adding the "+" symbol, instantiating the view controller, and adding the doseStore variable)

Nightfoxy:
Is this being added somewhere else? I really like this idea.

novalegra:
> Is this being added somewhere else? I really like this idea.

I closed this PR because I migrated this work to SwiftUI so it could be used with the Tidepool merge. If you're curious, you could definitely build these `multiple-insulin-curves` Loop & LoopKit branches with LoopWorkspace - my new PR only changes the UI for this feature to be in SwiftUI (and is created from the `tidepool-merge` branch). If you do build it, make sure to watch for unintended behavior, and please let me know if you see anything that looks weird.

PR #1297, opened 2020-04-07
Fix glucose unit on bolus VC when editing carb entry

sraka1:
When tapping to edit a carb entry, `glucoseUnit` was not set when preparing to present `CarbEntryViewController`. `CarbEntryViewController` in turn calls `BolusViewController` (`bolusVC`) with the default value of its property `glucoseUnit`, which is `milligramsPerDeciliter`.

Consequentially, on the Bolus screen when editing a carb entry, notifications regarding glucose predictions were shown in mg/dl even to mmol/l users. Charts were not affected as (as far as I was able to determine), `ChartsTableViewController` (which `BolusViewController` extends) sets the glucoseUnit for charts directly.

Screenshot is attached.
IMG_3730


mpangburn:
Nice catch, thanks!

ps2:
I believe this is fixed. Please reopen if that is not the case.

PR #1325, opened 2020-05-08
Upload cgmmanager's batteryLevel to nightscout if present

dabear:
Depends on https://github.com/LoopKit/LoopKit/pull/329

This allows CGMManagers to optionally specify their associated batteryLevel. It's a feature that I miss from when using xdrip+ on the android platform. I started out with a lot more complicated implementation touching RileylinkKit (nightscoutuploadkit), but decided to make the change as small as possible. 

Because the batteryLevel has a default implementation returning nil, this change would be optional for CGMManagers to implement. If CGMManagers don't implement this new property, or if they set the property to nil, there will be no change in behavior.

```
public final class SomeClientManager: CGMManager {
   public var batteryLevel: Int? {
        //return nil here if unknown, otherwise an int between 1 and 100
        return 66 
    }
    // (...)
}
```
An example implementation on a cgmmanger can be found here: https://github.com/dabear/miaomiao-client-swift/commit/c9f7e5ffb592b9c66564025798c2fd78d47e50da

It will be displayed like this in nightscout
image


dabear:
@ps2 This should be ready for review. The code is tested and working. I can share nightscouturl privately if you want to see how it looks

dabear:
Travis build failure is due LoopKit/LoopKit#329 having to be merged first.

dabear:
@ps2 any chance on getting this merged?

ps2:
@dabear This is going to conflict with the items being merged from Tidepool. I *will* take changes for this, but I think it will have to be updated once the Tidepool merge is complete.

francescaneri:
@dabear should work with dexcom g6 only if you are using xdrip + right?

dabear:
Will rework this and possibly resubmit

PR #1332, opened 2020-05-30
Adding Arabic translation 

abutmim:


abutmim:
after updating the translation
Do I need to do Pull request again ?

PR #1334, opened 2020-06-07
Dev-Complete ِArabic translations

abutmim:


abutmim:
Conflicting files
WatchApp/ar.lproj/Interface.strings



Is it Ok?

PR #1345, opened 2020-07-10
LOOP-1486: add suspend threshold information screen + editor

novalegra:
https://tidepool.atlassian.net/browse/LOOP-1486


novalegra:
Accidentally made to LoopKit instead of tidepool-org's fork

PR #1357, opened 2020-08-16
Add dose logging infrastructure + UI

novalegra:
## Description
This PR adds the ability to log outside insulin doses in Loop through the app itself, in addition to support for having multiple insulin types be active at the same time. Currently, when an outside dose is logged in HealthKit, Loop assumes that it is the same type of insulin as is currently present in the pump, which is not always the case. This change allows the user to select the insulin model if the dose is being logged; otherwise, if a bolus, basal, or temp basal is given via the pump, the *current insulin model* is saved as a part of the dose entry. Even if the user doesn't log doses, adding insulin model information allows for correct accounting of IOB and insulin effects if the insulin type in the pump is changed (for example, from Humalog to Fiasp). The insulin model of the dose is persisted in the Loop's data store and in HealthKit, and is not editable.

Tests have been updated accordingly based on the changes made, and additional tests were added with multiple insulin types active at once.

While this PR does not add additional insulin curves, the underlying changes to the dose model could support new insulin types if they were added into Loop (such as Afrezza).

## Current Testing Status:
- Has been tested with Dexcom G5 CGM, the mock pump, and Omnipod.
- Needs to be tested with Medtronic pumps.

## Other Relevant Info
- This PR also involves changes to LoopKit, see https://github.com/LoopKit/LoopKit/pull/335
- The infrastructure changes are the same as in https://github.com/LoopKit/Loop/pull/1296 (which was reviewed by Michael), though some of the UIs have changed so they're in SwiftUI
- This branch includes work from the Tidepool merge, so it will probably have to wait a bit to land


Cartfile
```diff
@@ -1,4 +1,4 @@
-github "LoopKit/LoopKit" "merge-from-tidepool"
+github "novalegra/LoopKit" "novalegra/multiple-insulin-swiftui"
```
novalegra, line 1: Carthage still needs to be updated once everything is ready to go in

DoseMathTests/DoseMathTests.swift
```diff
@@ -1092,4 +1092,35 @@ class RecommendBolusTests: XCTestCase {
         XCTAssertEqual(0, dose.amount)
     }
 
+    func testDoseWithFiaspCurve() {
```
novalegra, line 1095: Add tests with different insulin models

Loop/Managers/LoopDataManager.swift
```diff
@@ -657,6 +654,40 @@ extension LoopDataManager {
...
         }
     }
+    
+    /// Logs a new external bolus insulin dose in the DoseStore and HealthKit
+    ///
+    /// - Parameters:
+    ///   - startDate: The date the dose was started at.
+    ///   - value: The number of Units in the dose.
+    ///   - insulinModel: The type of insulin model that should be used for the dose.
+    func logOutsideInsulinDose(startDate: Date, units: Double, insulinModelSetting: InsulinModelSettings? = nil) {
```
novalegra, line 664: Created this function so there's a separate pathway to log doses vs administer a bolus dose

Loop/Views/BolusEntryView.swift
```diff
@@ -477,3 +518,14 @@ struct LabelBackground: ViewModifier {
             )
     }
 }
+
+extension Binding {
+    func onChange(_ handler: @escaping (Value) -> Void) -> Binding<Value> {
```
novalegra, line 523: This function will be build-in in iOS 14, but isn't currently available in iOS 13

Loop/Views/BolusEntryView.swift
```diff
@@ -293,13 +332,13 @@ struct BolusEntryView: View, HorizontalSizeClassOverride {
...
         VStack(spacing: 0) {
-            if isNoticeVisible {
+            if isNoticeVisible && !viewModel.isLoggingDose {
                 warning(for: viewModel.activeNotice!)
                     .padding([.top, .horizontal])
                     .transition(AnyTransition.opacity.combined(with: .move(edge: .bottom)))
             }
 
-            if isManualGlucosePromptVisible {
+            if isManualGlucosePromptVisible && !viewModel.isLoggingDose {
```
novalegra, line 341: I made it so logging a dose doesn't prompt for a manual BG, but I don't feel strongly about this

Loop/View Models/BolusEntryViewModel.swift
```diff
@@ -409,8 +486,13 @@ final class BolusEntryViewModel: ObservableObject {
     private func updatePredictedGlucoseValues(from state: LoopState, completion: @escaping () -> Void = {}) {
         dispatchPrecondition(condition: .notOnQueue(.main))
 
-        let (manualGlucoseSample, enteredBolus) = DispatchQueue.main.sync { (self.manualGlucoseSample, self.enteredBolus) }
-        let enteredBolusDose = DoseEntry(type: .bolus, startDate: Date(), value: enteredBolus.doubleValue(for: .internationalUnit()), unit: .units)
+        let (manualGlucoseSample, enteredBolus, doseDate, insulinModel) = DispatchQueue.main.sync { (self.manualGlucoseSample, self.enteredBolus, self.selectedDoseDate, self.selectedInsulinModel) }
+        var insulinModelSettings: InsulinModelSettings?
+        if let model = insulinModel {
+            insulinModelSettings = InsulinModelSettings(model: model)
+        }
```
novalegra, line 493: We only add the insulin model if it's actually set (which is not the case for manual boluses)

Loop/View Models/BolusEntryViewModel.swift
```diff
@@ -141,6 +182,30 @@ final class BolusEntryViewModel: ObservableObject {
...
+        $selectedDoseDate
+            .removeDuplicates()
+            .debounce(for: .milliseconds(400), scheduler: RunLoop.main)
+            .sink { [weak self] _ in
+                self?.dataManager.loopManager.getLoopState { manager, state in
+                    self?.updatePredictedGlucoseValues(from: state)
+                }
+            }
+            .store(in: &cancellables)
+    }
```
novalegra, line 208: Update the prediction if dose date or insulin model changes

Loop/View Models/BolusEntryViewModel.swift
```diff
@@ -73,6 +74,25 @@ final class BolusEntryViewModel: ObservableObject {
         predictedGlucoseChart.glucoseDisplayRange = BolusEntryViewModel.defaultGlucoseDisplayRange
         return ChartsManager(colors: .default, settings: .default, charts: [predictedGlucoseChart], traitCollection: .current)
     }()
+    
+    // MARK: - External Insulin
+    @Published var selectedInsulinModelIndex: Int = 0
```
novalegra, line 79: ANNA TODO: we should start the picker at the default insulin model setting

Loop/View Controllers/InsulinDeliveryTableViewController.swift
```diff
@@ -2,13 +2,571 @@
 //  InsulinDeliveryTableViewController.swift
 //  Naterade
 //
-//  Created by Nathan Racklyeft on 3/11/16.
+//  Created by Nathan Racklyeft on 1/30/16.
```
novalegra, line 5: Moved this from LoopKit because we need to add the DeviceManager into it. I basically added a section to view logged insulin doses, and a "+" button in the top right to open the dose logging screen

novalegra:
@ps2 I think this is ready for review now.

Loop.xcodeproj/xcshareddata/xcschemes/Complication - WatchApp.xcscheme
```diff
@@ -90,8 +90,8 @@
       launchAutomaticallySubstyle = "32">
       <RemoteRunnable
          runnableDebuggingMode = "2"
-         BundleIdentifier = "com.apple.carousel"
-         RemotePath = "/Naterade">
+         BundleIdentifier = "com.apple.Carousel"
+         RemotePath = "/$(MAIN_APP_DISPLAY_NAME)">
```
novalegra, line 94: Not sure which of these is the desired state

PR #1363, opened 2020-08-22
Add way to view override history in Loop

novalegra:
## Description
Currently, there's no great way to view the override history in Loop; the issue report does contain the history, but it's not in a very human-readable format. This PR contains a few small changes needed to support the associated LoopKit changes (see https://github.com/LoopKit/LoopKit/pull/336)

Since I started this before the Tidepool merge (which also bumps the target to iOS 13), I left the targets in the project file in this PR; they can easily be removed if the Tidepool stuff lands first.


novalegra:
Closing because I incorporated this into my Siri overrides PR

PR #1413, opened 2020-10-12
Automatic bolus with Lyumjev absorption model

alevar1975:
Dear Loop team,

Since the new insulin Lyumjev is spreading, at least in Germany, I added the absorption model of Lyumjev to the automatic bolus branch which I'm using right now. I've read the study PRONTO-T1D and T2D. 
Therefore, I set the peakActivity at 45 minutes, the actionDuration at 300 minutes, and the effectDelay at 8 minutes. I used to have the effectDelay at 10 minutes, but this caused some low sugars. That's why I added the available loc files as well. Please decide whether you want to use my changes. If yes, please be so kind to review all files.

 I added a current time layer line to the 4 home screen charts as well. I copied this code from Ivan Ivalkou's code because I like the vertical line indicating the current time. Take a look at it and decide whether you want to add this feature to Loop or not.

Since this is my first real pull request, I hope I've nearly met your expectations. If not, please let me know what could be improved to create a better pull request.

Kind Regals,
Alen


ps2:
Lyumjev is now in Loop dev.

PR #1429, opened 2020-10-22
Add Siri support for overrides

novalegra:
Currently, overrides aren't visible to Siri (meaning they can't be set with Shortcuts or via voice). There also isn't a great way to view the override history in Loop; the issue report does contain the history, but it's not in a very human-readable format. 

This PR:
1) adds a Siri extension & the ability to use overrides with Siri/Shortcuts
3) adds an override history screen
4) makes some under-the-hood improvements to override accounting in order to correctly determine the end date of an override

Some notes about the overrides with Siri implementation:
- I used `UserDefaults` so the extension could communicate with the main Loop app; the main app has a key-value observer that will watch for changes to set the appropriate override. One issue with this is that the KVO doesn't happen when Loop is in the background, so I changed the intent handler to always `.continueInApp` so the override will be set. Open to other ways to do this if there's something that could set the override in the background
- I chose to use an extension so it'd be easier to expand Loop's Siri capabilities in the future

[](http://www.youtube.com/watch?v=ufLSOx6KbX8 "Demo video")

LoopKit PR: https://github.com/LoopKit/LoopKit/pull/343
NightscoutService PR: https://github.com/ps2/NightscoutService/pull/4


Common/Extensions/UserDefaults+LoopIntents.swift
```diff
@@ -0,0 +1,42 @@
...
+extension UserDefaults {
+    
+    private enum Key: String {
+        case IntentExtensionContext = "com.loopkit.Loop.IntentExtensionContext"
+        // This key needs to be EXACTLY the same string as the objc dynamic var for the KVO to work correctly
+        case IntentExtensionOverrideToSet = "intentExtensionOverrideToSet"
+    }
+    
+    // Information for the extension from Loop
+    var intentExtensionInfo: IntentExtensionInfo? {
```
novalegra, line 20: I created `IntentExtensionInfo` so it would be easier to pass data between the intent extension and the main app in the event someone wanted to add other things to Loop's Siri capabilities

Common/Models/IntentExtensionInfo.swift
```diff
@@ -0,0 +1,33 @@
...
+//  IntentExtensionInfo.swift
+//  Loop Intent Extension
+//
+//  Created by Anna Quinlan on 10/17/20.
+//  Copyright © 2020 LoopKit Authors. All rights reserved.
+//
+
+import Foundation
+
+struct IntentExtensionInfo: RawRepresentable {
```
novalegra, line 11: I created IntentExtensionInfo so it would be easier to pass data between the intent extension and the main app in the event someone wanted to add other things to Loop's Siri capabilities

Loop Intent Extension/OverrideIntentHandler.swift
```diff
@@ -0,0 +1,48 @@
...
+        let lowercasedName = name.lowercased()
+        
+        if let defaults = self.defaults, defaults.intentExtensionInfo?.overridePresetNames?.first(where: {$0.lowercased() == lowercasedName}) != nil {
+            return true
+        }
+        return false
+    }
+    
+    func handle(intent: EnableOverridePresetIntent, completion: @escaping (EnableOverridePresetIntentResponse) -> Void) {
+        guard let defaults = self.defaults, let overrideName = intent.overrideName?.lowercased(), containsOverrideName(name: overrideName) else {
```
novalegra, line 25: My (pretty simple) implementation to set overrides just looks for the preset's name (and lowercases the names when comparing to avoid capitalization-related difficulties). This will fail to distinguish between overrides if they have the same name but different capitalization.

LoopCore/LoopSettings.swift
```diff
@@ -135,7 +135,7 @@ extension LoopSettings {
         case (nil, let scheduleOverride?):
             currentEffectiveOverride = scheduleOverride
         case (let preMealOverride?, let scheduleOverride?):
-            currentEffectiveOverride = preMealOverride.endDate > Date()
+            currentEffectiveOverride = preMealOverride.scheduledEndDate > Date()
```
novalegra, line 138: This is a rename that I made on the LoopKit level

ps2:
In the project file, `PRODUCT_BUNDLE_IDENTIFIER = "com..loopkit.Loop.Loop-Intent-Extension";` needs to change to use the `MAIN_APP_BUNDLE_IDENTIFIER`. Something like this: `PRODUCT_BUNDLE_IDENTIFIER = "$(MAIN_APP_BUNDLE_IDENTIFIER).Loop-Intent-Extension";`

PR #1437, opened 2020-11-20
Add github access token for Carthage

justinr1234:
You can be rate-limited when trying to build. Github allows more requests if you are authenticated.


ps2:
Loop is dropping support for Carthage based builds in favor of Workspace builds.

PR #1324, opened 2020-05-08
Add patch to handle invalid archs better (fixes #1117)

dekimsey:
This is an updated apply of PR #1126 which addresses the issue reported
in #1117

Note this has been an original issue was reported Sept 19. I myself, applied it locally with a minor tweak to support updated rileylink version for v2.0 in January and now again on v2.2.

Hopefully this helps others trying to sign the package.

Issue #1117, opened 2019-09-20
Can't build IPA with latest Dev with Xcode 11-2.


rsilvers129:
I am using the latest Dev and Xcode 11-2. I can build Loop for my phone, but when I try to Archive to make an Ad-Hoc IPA, it says IPA Processing Failed. It seems like part of MiniMed Plugin is having an issue. I see this in the logs:

2019-09-20 02:21:44 +0000 Scanning IPA...
2019-09-20 02:21:46 +0000 Assertion failed: Expected 2 archs in otool output:
/var/folders/8k/7c4406k90ds6vhz585ljmx140000gp/T/IDEDistributionOptionThinning.~~~kZyTmZ/Payload/Loop.app/Frameworks/MinimedKitPlugin.framework/MinimedKitPlugin:
Mach header
magic cputype cpusubtype caps filetype ncmds sizeofcmds flags
MH_MAGIC_64 X86_64 ALL 0x00 DYLIB 24 3136 NOUNDEFS DYLDLINK TWOLEVEL NO_REEXPORTED_DYLIBS
Load command 0

FlorinBv:
I have the same issue since few weeks. The problem is not with the Xcode 11, but with the DEV version. I have no issues on creating the IPA with Omnipod-testing version.

ludoded:
I'm having the same issue, any known workaround?

ps2:
You can try this PR: https://github.com/LoopKit/Loop/pull/1126

ludoded:
> You can try this PR: #1126

Doesn't work for me

josep1972:
@ludoded : Can you be more specific as to what you are seeing?

ludoded:
While trying to upload build to AppStore I got this error:

![Скриншот 2019-10-11 00 39 07](https://user-images.githubusercontent.com/4401289/66608589-9570f680-ebbf-11e9-917a-fb71330ca233.png)

And this is the log file while I got the error:
[IDEDistributionPipeline.log](https://github.com/LoopKit/Loop/files/3714725/IDEDistributionPipeline.log)

josep1972:
@ludoded Is hard to understand what you did from the log provided, but it does not seem like you re-built before distributing. In order you make sure this PR fixes the distribution problem, you have to re-Archive and then distribute. You can also run the Validate app step first, which I find to be easier to debug.

akashnsmahaswami:
I have the same issue with DEV version. Omnipod-testing version worked fine for me.
[...]

PR #1489, opened 2021-03-06
Removes x86_64 and i386 architectures from plugin frameworks

gestrich:
This is based off the work of PR [1324](https://github.com/LoopKit/Loop/pull/1324) which addresses the issue reported in [#1117](https://github.com/LoopKit/Loop/issues/1117). This PR change targets the master branch though. I also was not able to get PR [1324](https://github.com/LoopKit/Loop/pull/1324) to build IPAs without errors so opted for a smaller change in this PR that seems to work well.

The fix here resides in the copy-plugins.sh script, as I believe that is most closely the source of the issue. Extra architectures for frameworks (i386 and x86) are stripped in the copy-frameworks.sh script, via the call to `"${SRCROOT}/bin/carthage" copy-frameworks`. That script does not process plugin frameworks though, which are instead copied/installed via the copy-plugins.sh script. This PR will add a stripping mechanism for the plugin script as well for added parity between plugin frameworks vs non-plugin frameworks.

In terms of testing, I used Xcode 12.4 and checked the following. 

-Build directly to an iPhone 14.4.
-Simulator Builds
-Distribute an ad-hoc IPA.

### DIFF
--- a/Scripts/copy-plugins.sh
+++ b/Scripts/copy-plugins.sh
@@ -16,8 +16,14 @@ function copy_plugins {
       echo Copying device plugin: $plugin to frameworks directory in app
       plugin_path="$(readlink "$f" || echo "$f")"
       plugin_as_framework_path="${BUILT_PRODUCTS_DIR}/${FRAMEWORKS_FOLDER_PATH}/${plugin%.*}.framework"
+      
+      #Rename .plugin to .framework
       rsync -va --exclude=Frameworks "$plugin_path/." "${plugin_as_framework_path}"
-      # Rename .plugin to .framework
+      
+      #Remove sim architectures (for IPA distribution)
+      removeSimArchitectures "$plugin_as_framework_path"
+      
+      #Codesign
       if [ "$EXPANDED_CODE_SIGN_IDENTITY" != "-" ] && [ "$EXPANDED_CODE_SIGN_IDENTITY" != "" ]; then
         export CODESIGN_ALLOCATE=${DT_TOOLCHAIN_DIR}/usr/bin/codesign_allocate
         echo "Signing ${plugin} with ${EXPANDED_CODE_SIGN_IDENTITY_NAME}"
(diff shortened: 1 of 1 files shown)


Scripts/copy-plugins.sh
```diff
@@ -16,20 +16,30 @@ function copy_plugins {
       echo Copying device plugin: $plugin to frameworks directory in app
       plugin_path="$(readlink "$f" || echo "$f")"
       plugin_as_framework_path="${BUILT_PRODUCTS_DIR}/${FRAMEWORKS_FOLDER_PATH}/${plugin%.*}.framework"
```
gestrich, line 18: The fix is the call to the function I added removeSimArchitectures. That will strip out the architectures for the plugin frameworks.

I also added some comments to help organize the code a bit as I understand it. believe the second half of this function deals with nested frameworks within the plugin.

Scripts/copy-plugins.sh
```diff
@@ -16,20 +16,30 @@ function copy_plugins {
       echo Copying device plugin: $plugin to frameworks directory in app
       plugin_path="$(readlink "$f" || echo "$f")"
```
gestrich, line 17: The fix is the call to the function I added removeSimArchitectures. That will strip out the architectures for the plugin frameworks.

I also added some comments to help organize the code a bit as I understand it. believe the second half of this function deals with nested frameworks within the plugin.

dekimsey:
I'm somewhat confused as to why this issue doesn't get more visibility. What is it we are doing differently that's not affecting the maintainers? Is this because we are building an Archive and not just using the de-device local debug builds?

bill-foreflight:
@dekimsey I'm not sure if this is still an issue or not as I haven't been building master for a long time. I've only been building dev which doesn't need this fix but master did at one time. You are correct that the original issue would only affect those building Archives which is not a lot of people.

PR #1492, opened 2021-03-18
support for tuners

almanar:
This change adds support for parameter tuners to Loop,
with an accompanying UI to tune, review, and save tuned
parameters.

There is currently only one implemenation of tuners:
remote tuners package up relevant data, and send a tuning
request to a web service, which in turn responds with the
tuned parameters. The protocol is documented elsewhere [1].
This allows for experimentation with multiple tuning approaches,
and also to use tools that have rich support for data science
and machine learning (e.g., python).

A model based tuner is implemented in [1].

[1] https://github.com/mariusae/tune/tree/master/README.md


dabear:
Why was this closed? Looked interesting

PR #1494, opened 2021-03-20
Store device token to LoopSettings

gestrich:
I was getting an error "Could not find deviceToken in loopSettings." error while trying to send push notifications. It appears the token gets set to LoopSettings.deviceToken but that value isn't stored to LoopSettings.

I'm not very familiar with this area of code or how it interacts with `StoredSettings`, which also includes a device token. So this may not be the appropriate fix.

### DIFF
--- a/LoopCore/LoopSettings.swift
+++ b/LoopCore/LoopSettings.swift
@@ -295,6 +295,8 @@ extension LoopSettings: RawRepresentable {
             let dosingStrategy = DosingStrategy(rawValue: rawDosingStrategy) {
             self.dosingStrategy = dosingStrategy
         }
+        
+        self.deviceToken = rawValue["deviceToken"] as? Data
     }
 
     public var rawValue: RawValue {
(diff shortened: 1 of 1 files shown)


yodax:
I was having the same issue. This fixes it for me. 👍

ps2:
This is not the right fix. "Never cache the device token locally on the user's device." from https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622958-application .

iOS reports the token to Loop at some point in the startup cycle. If Loop uploads settings before then, the token may be nil.  The right fix is probably either prevent settings uploads until the token is reported, or find a way in NS to use the last non-nil token. Haven't really dug into this yet, but welcome anyone wanting to do so, considering the above.

bill-foreflight:
Ok cool. Thanks for looking into.

gestrich:
Further discussion on these comments in https://github.com/LoopKit/Loop/issues/1536

PR #1495, opened 2021-03-21
Remote Bolus/Carbs

gestrich:
This is a WIP. Expect git force push until to this moves to review.


ps2:
Thanks Bill! I assume this works with Nightscout playing the role of sending the request via APNS? If this is a Nightscout centric feature, would it make sense for more of this code to live in the Nightscout Service plugin? Just wondering if, for example some other service might have a different approach to validating the authenticity of a remote command.

Haven't really dug into this yet; I should set aside some time to really review how this is all intended to work.

gestrich:
@ps2 This draft PR is an update of Jose's PR #1293 which processes a push notification from Nightscout. Thus far it only resolves the conflicts and adds a row to the Settings UI for OTP just to confirm the feature is still functional in dev. (in the Therapy section temporarily)

From my very basic understandings of the plugin architecture, I can see this living in the Nightscout plugin. I did a little review of that architecture and here is a rough sketch of where this code could live:

- Move the OTP library and manager logic (`OTPManager`, developed by Jose) into NightScoutServiceKit. The `NightScoutService` could hold a reference to the `OTPManager`.
- Move the OTP UI components (ex: Jose's `OTPNavigationController`) into NightscoutServiceKitUI. Possibly Nightscout's `ServiceStatusView` could have a new row for OTP.

I was considering how the Loop app and plugin will communicate. There is bi-directional flow as the app receives the push notification and then consults with the plugin, and the plugin triggers or recommends bolus/carb entries. One thought was something like this:

- The `DeviceDataManager`, which receives push notifications today, asks the `RemoteDataServicesManager` to handle an incoming push notification. 
- The `RemoteDataServicesManager` checks with each of its services on whether they can handle the push. The `Service.identifer` string could be included in the push notification, and that could be used to map to its owning service. A simpler alternative is just to ask each service to handle it.
-The `RemoteDataService` procotol can include a new method, like `handleRemoteNotification`, that is called on the service by `RemoteDataServicesManager`.
-Once `NightScoutServiceKit` processes the notification, it can call a new protocol methods on the `ServiceDelegate`, similar to `serviceHasNewTherapySettings`. 
	`serviceHasNewTemporaryOverride(override: TemporaryScheduleOverride)`
	`serviceHasNewBolus(amount: Double)`
	`serviceHasNewCarbEntry(carbEntry: NewCarbEntry)`

Let me know if any thoughts on this generally or if you can see any gaps.

gestrich:
An alternative to the last bullet - The RemoteDataServicesManager could more explicitly request results from the services, based on what is in the push notification payload:

`let tempOverride = service.getTemporaryOverrideForNotifcation(notification...)`
`//Now Loop enacts the override`

I think the trade-off to this approach is the payload structure and allowed actions are defined in the Loop target, rather than being hidden and defined in the service. More power to the service vs more control by the Loop app.

bill-foreflight:
This PR can be closed or disregarded. I'll have a different PR against the NightScoutService repo instead.

PR #1508, opened 2021-05-23
Remote Bolus Dev

gestrich:
This PR includes changes to the Loop module to support remote/bolus in dev. See the [workspace PR](https://github.com/LoopKit/LoopWorkspace/pull/12) for more details on the overall effort. Comments added inline for review.


Loop/AppDelegate.swift
```diff
@@ -22,6 +22,7 @@ final class AppDelegate: UIResponder, UIApplicationDelegate, WindowProvider {
 
         loopAppManager.initialize(windowProvider: self, launchOptions: launchOptions)
         loopAppManager.launch()
+        UIApplication.shared.registerForRemoteNotifications()
```
ps2, line 25: This is already happening in NotificationManager; we should probably only do this once.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1210,6 +1211,36 @@ extension DeviceDataManager {
                 case .cancelTemporaryOverride:
                     log.default("Canceling temporary override from remote command")
                     loopManager.settings.scheduleOverride = nil
+                case .bolusEntry(let bolusValue):
```
ps2, line 1214: We should trigger a user notification when this happens. Or is the push notification including foreground text?

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1210,6 +1211,36 @@ extension DeviceDataManager {
...
+                       if bolusValue.isLessThanOrEqualTo(maxBolus) {
+                          self.enactBolus(units: bolusValue, automatic: true)
+                       } else {
+                           log.default("Remote bolus higher than maximum. Aborting...")
+                       }
+                    } else {
+                        log.default("No max bolus detected. Aborting...")
+                    }
+                case .carbsEntry(let newEntry):
+                    log.default("Adding carbs entry.")
```
ps2, line 1234: A local notification should happen with this too.

ps2 (changes requested):
Thanks Bill, could you address the comments?  Otherwise, this is looking great! Thanks!

Loop/AppDelegate.swift
gestrich: @ps2  This is part of the proposed fix outlined in [1536](https://github.com/LoopKit/Loop/issues/1536) among others). From what I can tell, we _only_ fetch the token via `registerForRemoteNotifications` during onboarding. Since we can't persist the token per iOS guidelines, it will be available in-memory only and not be available for later upload after an app reboot. Here's the cases where that could pose an issue:

1. User chooses to setup Nightscout after onboarding.
2. User goes offline during onboarding process.

I think attempting to fetch it on each startup would be good protection against the above scenarios -- it will always be available to in memory and uploaded to Nightscout when needed.

If that seems reasonable, I agree we don't need it in both places. How about I remove it from onboarding? Or are there other considerations I'm not thinking of?

gestrich: I updated this PR in this way. Please let me know if any any questions on that.

Loop/Managers/DeviceDataManager.swift
gestrich: Updated

Loop/Managers/DeviceDataManager.swift
gestrich: I've added notifications to indicate success and failure once the remote notification is processed. That should handle the foreground notification and allow for better troubleshooting for both foreground and background states. I've added screenshots and more discussion in Zulip https://loop.zulipchat.com/#narrow/stream/144182-development/topic/Remote.20bolus.20.2F.20remote.20carbs

Loop.xcodeproj/project.pbxproj
```diff
@@ -375,6 +375,8 @@
 		89FE21AD24AC57E30033F501 /* Collection.swift in Sources */ = {isa = PBXBuildFile; fileRef = 89FE21AC24AC57E30033F501 /* Collection.swift */; };
 		A90EF53C25DEF06200F32D61 /* PluginManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = C16DA84122E8E112008624C2 /* PluginManager.swift */; };
 		A90EF54425DEF0A000F32D61 /* OSLog.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4374B5EE209D84BE00D17AA8 /* OSLog.swift */; };
```
gestrich, line 377: Base32 and OneTimePassword are the new git submodules that are linked in the workspace. I believe they need added to the Loop project, and to be embedded, in order for things to work.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1210,6 +1211,67 @@ extension DeviceDataManager {
                 case .cancelTemporaryOverride:
                     log.default("Canceling temporary override from remote command")
                     loopManager.settings.scheduleOverride = nil
+                case .bolusEntry(let bolusAmount):
+                    log.default("Enacting remote bolus entry: %{public}@", String(describing: bolusAmount))
+
```
gestrich, line 1216: Add explicit guards for the various error conditions on a remote bolus. In there is an error, that will be reported back as a user local notification.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1210,6 +1211,67 @@ extension DeviceDataManager {
...
+                        log.default("No max bolus detected. Aborting...")
+                        return
+                    }
+                    
+                    guard bolusAmount.isLessThanOrEqualTo(maxBolusAmount) else {
+                        NotificationManager.sendRemoteBolusFailureNotification(for: RemoteCommandError.exceedsMaxBolus, amount: bolusAmount)
+                        log.default("Remote bolus higher than maximum. Aborting...")
+                        return
+                    }
+
```
gestrich, line 1235: When enacting the bolus here, we may get a specialized error from the enactBolus API which will also be presented in a local notification.... otherwise we show a local notification with a success message.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1210,6 +1211,67 @@ extension DeviceDataManager {
...
+                        return
+                    }
+                    
+                    self.enactBolus(units: bolusAmount, automatic: true) { error in
+                        if let error = error {
+                            NotificationManager.sendRemoteBolusFailureNotification(for: error, amount: bolusAmount)
+                        } else {
+                            NotificationManager.sendRemoteBolusNotification(amount: bolusAmount)
+                        }
+                    }
```
gestrich, line 1242: Same paradigm here for remote carb entries.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1210,6 +1211,67 @@ extension DeviceDataManager {
...
+                        log.info("Could not validate notification: %{public}@", String(describing: notification))
+                        return
+                    }
+                    
+                    guard candidateCarbsInGrams > 0.0 else {
+                        NotificationManager.sendRemoteCarbEntryFailureNotification(for: RemoteCommandError.invalidCarbs, amountInGrams: candidateCarbsInGrams)
+                        log.default("Invalid carb entry amount. Aborting...")
+                        return
+                    }
+
```
gestrich, line 1260: Added this check that we don't exceed the max carb entry amount.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1210,6 +1211,67 @@ extension DeviceDataManager {
...
+                        log.default("Invalid carb entry amount. Aborting...")
+                        return
+                    }
+                    
+                    guard candidateCarbsInGrams <= LoopConstants.maxCarbEntryQuantity.doubleValue(for: .gram()) else {
+                        NotificationManager.sendRemoteCarbEntryFailureNotification(for: RemoteCommandError.exceedsMaxCarbs, amountInGrams: candidateCarbsInGrams)
+                        log.default("Carbs higher than maximum. Aborting...")
+                        return
+                    }
+
```
gestrich, line 1266: Finally we attempt to add the carbs, and also show an error if that fails (unlikely - would be Core Data issue it seems)

Loop/Managers/LoopAppManager.swift
```diff
@@ -83,6 +83,8 @@ class LoopAppManager: NSObject {
         }
 
         registerBackgroundTasks()
```
gestrich, line 85: Registering for remote notifications on each startup -- this was discussed above and also in https://github.com/LoopKit/Loop/issues/1536.

Loop/Managers/LoopAppManager.swift
```diff
@@ -208,6 +210,12 @@ class LoopAppManager: NSObject {
     }
 
     // MARK: - Remote Notification
+
```
gestrich, line 213: Pete - Since you looked over this before I realized I should have this feature flag check. At least that is what is done in dev now.

Loop/Managers/LoopAppManager.swift
```diff
@@ -324,10 +332,13 @@ extension LoopAppManager: UNUserNotificationCenterDelegate {
         case LoopNotificationCategory.bolusFailure.rawValue,
              LoopNotificationCategory.pumpBatteryLow.rawValue,
              LoopNotificationCategory.pumpExpired.rawValue,
```
gestrich, line 334: New notifications for remote bolus/carbs. More on that below.

Loop/Managers/LoopDataManager.swift
```diff
@@ -798,7 +798,7 @@ extension LoopDataManager {
                                       maximumBasalRatePerHour: loopSettings.maximumBasalRatePerHour,
                                       maximumBolus: loopSettings.maximumBolus,
                                       suspendThreshold: loopSettings.suspendThreshold,
```
gestrich, line 800: This was discussed in https://github.com/LoopKit/Loop/issues/1536... But basically loopSettings doeesn't store the deviceToken so this will be nil. We may want to just remove that property altogether to avoid the confusion but I'd need to verify that is ok.

Loop/Managers/NotificationManager.swift
```diff
@@ -67,14 +67,6 @@ extension NotificationManager {
         center.requestAuthorization(options: authOptions) { (granted, error) in
             UNUserNotificationCenter.current().getNotificationSettings { settings in
                 completion(settings.authorizationStatus)
```
gestrich, line 69: Discussed in https://github.com/LoopKit/Loop/issues/1536 but need to make sure this is called on each startup since we are not persisting it anywhere.

Loop/Managers/NotificationManager.swift
```diff
@@ -156,6 +148,85 @@ extension NotificationManager {
             UNUserNotificationCenter.current().add(request)
         }
     }
```
gestrich, line 150: Helpers to send the new local notifications.

Loop/Managers/RemoteDataServicesManager.swift
```diff
@@ -361,6 +361,21 @@ extension RemoteDataServicesManager {
 
 }
```
gestrich, line 363: Check our data services for one to validate our notification here. As a future enhancement, data services like Nightscout could provide some kind of identifier in the push payload. That would allow Loop to dispatch the notification to the corresponding service rather than sending it to all the services like this.

Loop/Models/RemoteCommand.swift
```diff
@@ -8,15 +8,42 @@
 
 import Foundation
 import LoopKit
+import HealthKit
```
gestrich, line 12: I don't believe this was being used from what I can tell. But it looks like the right structure to track remote command errors like I needed.

Loop/Models/RemoteCommand.swift
```diff
@@ -8,15 +8,42 @@
...
 
-public enum RemoteCommandError: Error {
+public enum RemoteCommandError: LocalizedError {
     case expired
+    case invalidOTP
+    case missingMaxBolus
+    case exceedsMaxBolus
+    case exceedsMaxCarbs
+    case invalidCarbs
+
```
gestrich, line 20: I'm using this description in the local notifications that show.

Loop/Models/RemoteCommand.swift
```diff
@@ -34,6 +61,17 @@ extension RemoteCommand {
             self = .temporaryScheduleOverride(override)
         } else if let _ = notification["cancel-temporary-override"] as? String {
             self = .cancelTemporaryOverride
+        }  else if let bolusValue = notification["bolus-entry"] as? Double {
+            self = .bolusEntry(bolusValue)
+        } else if let carbsValue = notification["carbs-entry"] as? Double {
```
gestrich, line 66: Jose had this original comment on the default absorption value.  I hadn't dug in but it doesn't seem critical now either.

ps2:
Token management has been resolved in dev, I believe. This PR needs to be updated to resolve conflicts with the latest dev. I'll see if I can land the NightScoutService soon.

gestrich:
@ps2 conflicts + token workarounds have been removed.

ps2:
Merged these changes in https://github.com/LoopKit/Loop/commit/5b25c83282e0b0706e97a072fece727a0484782d.  Forgot to squash, so they're *all* separate in the log. :-/

PR #1523, opened 2021-08-15
Zero out the recommended dose when user starts to edit bolus

novalegra:
Currently, when a user decides to edit the recommended bolus, they have to delete the entry first, which adds more taps to this common interaction. This PR zeros out the bolus entry field if the user taps to edit the bolus and the currently-entered dose matches the recommended amount.

IMG_54DC4D28C48A-1


ps2:
Thanks @novalegra! I used this patch, but applied it to the `DismissibleKeyboardTextField` view instead, because I was seeing that tapping on the `U` units would clear the field, but not bring up the editor, which was confusing.

PR #1547, opened 2021-09-14
fix Retrospective Correction subtitle

dm61:
This adds back the format string describing retrospective glucose prediction comparison, which fixes the Retrospective Correction subtitle on the Predicted Glucose screen.


ps2:
Thanks @dm61. Sorry this got lost; did get fixed here: https://github.com/LoopKit/Loop/pull/1608

PR #1591, opened 2022-01-22
Update capture-build-details.sh

marionbarker:
This change reports the LoopWorkspace build if Loop is at a detached HEAD instead of a branch and explicitly adds either Loop or LoopWorkspace to the string.
It also limits the SHA-1 to 7 characters.


Scripts/capture-build-details.sh
```diff
@@ -64,13 +64,25 @@ info "Gathering build details in ${git_source_root}"
 cd "${git_source_root}"
 
 if [ -e .git ]; then
-  rev=$(git rev-parse HEAD)
-  plutil -replace com-loopkit-Loop-git-revision -string ${rev} "${info_plist_path}"
+  # if branch exists, report Loop branch and SHA
```
ps2, line 67: Are you trying to tell the difference between a LoopWorkspace build and a Loop standalone build by whether the current directory is on a branch? This will fail if people are building LoopWorkspace, but have a branch checked out here.  A better test might be to look for the presence of`../Loop.xcworkspace`

marionbarker: Use case - handle the average user with a LoopWorkspace build where the configuration is defined by the branch and SHA-1 of LoopWorkspace and where there is HEAD detached in Loop.

Use case - handle the developer who may be working in a particular branch in the Loop folder and wants to have that information captured.

This is not a panacea for gathering all configuration information, but I thought it was an improvement now that Xcode 13 has pushed us to Workspace builds. I did not think to test it for a zip download. Just offering it for your consideration.

Example:
```
## LoopVersion
* Version: FreeAPS v2.2 (204)
* profileExpiration: 2023-01-11 21:24:21 +0000
* gitRevision: LoopWorkspace 24eacda
* gitBranch: LoopWorkspace rk-dev-02
* sourceRoot: /Users/marion/Downloads/ManualClones/RK-dev-02_mdb_20220105/LoopWorkspace/Loop
* buildDateString: Sat Jan 15 00:04:15 PST 2022
* xcodeVersion: 13C100
```

Scripts/capture-build-details.sh
```diff
@@ -64,13 +64,25 @@ info "Gathering build details in ${git_source_root}"
...
+  # else, go up one level and report LoopWorkspace values
   branch=$(git branch --show-current)
   if [ -n "$branch" ]; then
-    plutil -replace com-loopkit-Loop-git-branch -string "${branch}" "${info_plist_path}"
+    rev=$(git rev-parse HEAD)
+    plutil -replace com-loopkit-Loop-git-revision -string "Loop ${rev:0:7}" "${info_plist_path}"
```
ps2, line 72: The git-revision shouldn't have "Loop" in it, nor should the git-branch; app name is taken from other parts of the plist.

ps2: You could add a 'com-loopkit-Loop-build-type' pref if you wanted. But I'm not exactly sure if "build type" is what you're trying to track.

ps2 (changes requested):

Scripts/capture-build-details.sh
ps2: Yes, both those changes sound good. It might be cleaner to just not add those plist values for a non-workspace build.

PR #1602, opened 2022-02-09
increase log duration, add markdown header mark

marionbarker:
This PR, along with [Update capture-build-details.sh](https://github.com/LoopKit/Loop/pull/1591) will assist in providing full pod history for support of both Eros and Dash pods.

This PR:
* Removes some of the verbose portions of the Loop Report that are now available in the Export Critical Event Logs tab
* Extends the duration of the reported log to 84 hours, which should cover the full history of a pod for most cases
* Adds a header marker for the code Version in the markdown output

The last 2 items on the list simplify the work required to parse and report on Dash pods as we are testing the code to support Dash implementation.

The extended duration is helpful for supporting both Eros and Dash when Loop users report problems with their pods online.

Note that without the full history (and the corresponding 0x0115 and 0x011b messages from the pod), the lot, sequence number and firmware versions are not available.

The other PR, #1591, reports the commit for the associated LoopWorkspace, which for most builds, will identify the full build with submodules for the workspace build.

### DIFF
--- a/Loop/Managers/DeviceDataManager.swift
+++ b/Loop/Managers/DeviceDataManager.swift
@@ -606,7 +606,7 @@ final class DeviceDataManager {
                         "",
                         String(reflecting: self.statusExtensionManager!),
                         "",
-                        // loopReport,
+                        loopReport,
                         alertReport
                         ].joined(separator: "\n")
 
--- a/Loop/Managers/DeviceDataManager.swift
+++ b/Loop/Managers/DeviceDataManager.swift
@@ -577,7 +577,7 @@ final class DeviceDataManager {
                     }
 
                     let report = [
-                        "## Version",
+                        "## Build Details",
                         "* appNameAndVersion: \(Bundle.main.localizedNameAndVersion)",
                         "* profileExpiration: \(Bundle.main.profileExpirationString)",
                         "* gitRevision: \(Bundle.main.gitRevision ?? "N/A")",
--- a/Loop/Managers/DeviceDataManager.swift
+++ b/Loop/Managers/DeviceDataManager.swift
@@ -578,7 +578,7 @@ final class DeviceDataManager {
 
                     let report = [
                         "## Version",
-                        "* codeVersion: \(Bundle.main.localizedNameAndVersion)",
+                        "* appNameAndVersion: \(Bundle.main.localizedNameAndVersion)",
                         "* profileExpiration: \(Bundle.main.profileExpirationString)",
                         "* gitRevision: \(Bundle.main.gitRevision ?? "N/A")",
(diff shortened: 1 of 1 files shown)


Loop/Managers/DeviceDataManager.swift
```diff
@@ -600,11 +602,11 @@ final class DeviceDataManager {
                         "## Device Communication Log",
                         deviceLogReport,
                         "",
-                        String(reflecting: self.watchManager!),
-                        "",
-                        String(reflecting: self.statusExtensionManager!),
-                        "",
-                        loopReport,
+                        // String(reflecting: self.watchManager!),
```
ps2, line 605: Why are these being commented out? They're useful to see the state of watch and status manager.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -576,7 +577,8 @@ final class DeviceDataManager {
                     }
 
                     let report = [
-                        Bundle.main.localizedNameAndVersion,
+                        "## Version",
+                        "* codeVersion: \(Bundle.main.localizedNameAndVersion)",
```
ps2, line 581: This is the app name and version. not sure what "code version" is. That sounds like something that should be tracked by gitRevision/branch.

ps2 (changes requested):

Loop/Managers/DeviceDataManager.swift
marionbarker: I was trying to reduce the size of the log file while increasing duration for pods.
I wrote up the sizes in thie [zulipchat conversation](https://loop.zulipchat.com/#narrow/stream/312259-Omnipod-DASH/topic/Log.20Files.20and.20Parser/near/269952885)

marionbarker: Restored watch and status manager.

Loop/Managers/DeviceDataManager.swift
ps2: The `## Version` header is a header for a section that is describing much more than just version. Could we call the section `## Build Details` ?

ps2 (changes requested):

Loop/Managers/DeviceDataManager.swift
marionbarker: I was about to suggest something like that.  Makes a lot more sense.

marionbarker: Done

ps2:
Thanks! I do think it would be best if we could either update tools to use critical event log, or make a pod specific log exporter (could even do more details, like bluetooth comms/etc), but this will work for now.

PR #1607, opened 2022-02-21
Add LoopWorkspace to Build Info and Move Alerts

marionbarker:
Replaces PR #1591 

* report LoopWorkspace Build Info
* move alerts section closer to beginning of Loop Report

Sample Output in Loop Report:
```
## Build Details
* appNameAndVersion: Loop v2.3.0 (57)
* profileExpiration: 2023-01-20 16:11:22 +0000
* gitRevision: 8ec85bd
* gitBranch: workspace_build_info
* isWorkspaceBuild: Yes
* gitRevWorkspace: a6b5278
* gitBranchWorkspace: dash-integration
* sourceRoot: /Users/marion/Downloads/ManualClones/Loop_dev_dash_20220124/LoopWorkspace/Loop
* buildDateString: Sun Feb 20 18:17:17 PST 2022
* xcodeVersion: 13C100
```

followed by Feature Flags, then Alerts.


Common/Extensions/NSBundle.swift
```diff
@@ -85,6 +85,19 @@ extension Bundle {
         }
     }
 
+    // These strings are only configured if it is a workspace build
+    var workspaceBuild: String? {
+        return object(forInfoDictionaryKey: "com-loopkit-LoopWorkspace-y-n") as? String
```
ps2, line 90: I'm not sure what the y-n means? If it's to communicate whether or not is a workspace build, it should be something like com-loopkit-isWorkspaceBuild and be a boolean.  Also, it may be unneeded if the following are only set if it's a workspace build, as we can tell by their presence.
All three var names should probably share the same naming scheme.
Prefer full names to abbreviations: e.g. workspaceGitRevision vs gitRevWs

Loop/Managers/DeviceDataManager.swift
```diff
@@ -587,13 +587,18 @@ final class DeviceDataManager {
...
+                        "* gitRevWorkspace: \(Bundle.main.gitRevWs ?? "N/A")",
+                        "* gitBranchWorkspace: \(Bundle.main.gitBranchWs ?? "N/A")",
                         "* sourceRoot: \(Bundle.main.sourceRoot ?? "N/A")",
                         "* buildDateString: \(Bundle.main.buildDateString ?? "N/A")",
                         "* xcodeVersion: \(Bundle.main.xcodeVersion ?? "N/A")",
                         "",
                         "## FeatureFlags",
                         "\(FeatureFlags)",
                         "",
+                        alertReport,
```
ps2, line 600: Just a couple notes about ordering: I assume this is how *you* would sort it because you find this more useful than the other fields. As I generally have to look into algorithm issues, the dosing info (carbs/doses) are more important for me. I'll take this change as the file is already so unwieldy already we generally need to navigate by search anyways. But please don't look at this a file to optimize and spend time on for general users. If there are things in here that users need on a regular basis, we should be spending time figuring out how to expose them in the UI.

ps2 (changes requested):
Thanks! Mainly just some naming comments.

Loop/Managers/DeviceDataManager.swift
marionbarker: I was thinking of the use case where there was a pod failure and the user wanted to see the alert on their phone.
It is very painful to scroll long distances in a file on the phone interface and this would make it easier to find.  
Note - I did this before you added the UI that showed it for last pod.
Even with that feature, there may be cases of back-to-back faults and user could see both in the report.

Common/Extensions/NSBundle.swift
marionbarker: I kind of fumbled around trying to figure out how to do this.
I don't know how to do the boolean, so I took it out.
Redid the names and tested locally.

marionbarker:
I believe this is ready for review again.  Thanks for the constructive suggestions.

Loop/Managers/DeviceDataManager.swift
ps2: Thanks for the changes. Tracking LoopWorkspace git revision/branch will probably be useful to those who need to dig into Loop issue reports, and the shorter sha hash is nice.

ps2 (approved):

PR #1619, opened 2022-03-02
Initial Afrezza Support

damonbayer:


Common/FeatureFlags.swift
```diff
@@ -81,6 +82,13 @@ struct FeatureFlagConfiguration: Decodable {
         #else
         self.lyumjevInsulinModelEnabled = true
         #endif
+
+        // Swift compiler config is inverse, since the default state is enabled.
+        #if LYUMJEV_INSULIN_MODEL_DISABLED
```
elnjensen, line 87: Shouldn’t this line reference Afrezza rather than Lyumjev?

damonbayer: Fixed, thanks!

PR #1628, opened 2022-03-08
Bolus clear fix

ps2:
There are some reports of the existing clear-on-edit mechanism not actually clearing out the editable bolus amount field in the bolus view. This uses a different strategy to clear the field. 

Looking to test reports from those who have experienced the failure to clear. 

It also contains changes to streamline forecast and recommendation generation; previously, bringing up the view would cause at least 7 forecast predictions to be run, and would cause flashing as the various parts of the UI updated, usually rendering a view without the bolus and then rendering one with the bolus.

### DIFF
--- a/Loop/Managers/DeviceDataManager.swift
+++ b/Loop/Managers/DeviceDataManager.swift
@@ -1108,9 +1108,6 @@ extension DeviceDataManager: GlucoseStoreDelegate {
 // MARK: - TestingPumpManager
 extension DeviceDataManager {
     func deleteTestingPumpData(completion: ((Error?) -> Void)? = nil) {
-//        guard FeatureFlags.scenariosEnabled else {
-//            fatalError("\(#function) should be invoked only when scenarios are enabled")
-//        }
 
         guard let testingPumpManager = pumpManager as? TestingPumpManager else {
             assertionFailure("\(#function) should be invoked only when a testing pump manager is in use")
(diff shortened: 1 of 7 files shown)


marionbarker:
This worked the second time (when I was prompt and tapped the Bolus line and got a 0 and a save without bolusing button).

The first time i tried it, I got in an infinite loop of typing in a 0 U bolus, saying done and the only button offered is Enter Bolus which brings up the keyboard. 

See if I can repeat this issue (tried several times): 
* delete previous fake carbs
* add new carbs to test phone
* bolus screen shows up
* tap the Bolus line to bring up keyboard, hit done without entry
* See a Save without Bolus button
* Wait until next CGM value
* Both the Recommended Bolus and Bolus lines updated to new value.
* Tap on Bolus line, it brings up keyboard but does not zero out the amount.
* I zeroed the amount with the backspace button on the keyboard.
* I did not get stuck in the "loop" this time

to be explicit - this is the change I was testing
```
Loop/Views/BolusEntryView.swift
@@ -57,7 +57,8 @@ struct BolusEntryView: View {
                     self.shouldBolusEntryBecomeFirstResponder = false
                 } else {
                     if !editedBolusAmount {
-                        typedBolusEntry.wrappedValue = ""
+                        enteredBolusAmount = ""
+                        self.viewModel.enteredBolus = HKQuantity(unit: .internationalUnit(), doubleValue: 0)
                         editedBolusAmount = true
                     }
                 }
```

PR #1636, opened 2022-03-14
Reduce WatchApp visible duration from 6 hours to 2 hours

mitchellhenke:
Based on some discussion in zulip [here](https://loop.zulipchat.com/#narrow/stream/144182-development/topic/WatchApp.20on.20dev/near/271923672)

The shorter window has made it easier for me to see the recent trends and near-term projections, which is where the watch is most useful 🙂


ps2:
Thanks. Would still like to see this as a preferences item, but can take this for now.

PR #1653, opened 2022-03-26
Support more precision in remote bolus notifications

gestrich:
Remote bolus notifications were not formatted properly for values like 1.02, which would round to 1.1 in the visible notifications. (would not affect delivery though). I was only supporting 1 precision unit. The screenshot below is with the fix.

IMG_8CCF2067FDE8-1


ps2:
If we're going the route of quantity formatters, you could use the extensions in LoopKit, which include the `U` part of the string, and then it would support localizations more directly. But it's not significant win.

PR #1655, opened 2022-03-27
Create new dosing decision after forecast change, for external clients

ps2:
When the forecast changes, make a new dosing decision (without actually looping), and store for external clients.


Nightfoxy:
Thanks for the work on this! Went from missing a feature to significantly improved over master very quickly!

PR #1668, opened 2022-04-11
add more german localizations

Joerg-Schoemer:


Joerg-Schoemer:
Is it possible to run unit-tests for internationalization?

Joerg-Schoemer:
It’s really hard to find the right places to put the translation.

Issue #1643, opened 2022-03-19
Duplicate carb entries can be saved 


damonbayer:
**Describe the bug**
Loop sometimes saves duplicate carb entries if "Save without blousing" is tapped multiple times in quick succession. Initially, none of these entries appear in the carbohydrates screen, but they will appear after going back to the main screen and returning to the carbohydrates screen. Video is attached.

**To Reproduce**

**Expected behavior**
Only one carb entry should be saved and it should appear immediately on the carbohydrates screen.

**Screenshots**
[Video.](https://youtu.be/RuUUOqjisTs)

**Phone**
 - Hardware: iPhone 13 Minin
 - OS Version: 15.4

## Build Details
* appNameAndVersion: Loop v2.3.0 (57)
* profileExpiration: 2023-01-21 21:50:57 +0000
* gitRevision: d1890e4
* gitBranch: N/A
* workspaceGitRevision: e0ca2ce
* workspaceGitBranch: dev
* sourceRoot: /Users/damon/Downloads/LoopWorkspace/Loop
* buildDateString: Mon Mar 14 10:26:03 PDT 2022
* xcodeVersion: 13C100

Joerg-Schoemer:
We've seen this yesterday. Our son had saved accidentally 30g carbs instead of 10g carbs.
This leads to a hypo after approximately an hour.
He expected a fingerprint reading but it doesn't showed up, so he pushed multiple times 😞

NaomiDonahue:
Had the same issue. Here is the report:

Generated: 2022-04-29 00:56:15 +0000



## Build Details
* appNameAndVersion: Loop v2.3.0 (57)
* profileExpiration: 2023-03-28 16:44:18 +0000
* gitRevision: 549a41a
* gitBranch: N/A
* workspaceGitRevision: 87c835a
* workspaceGitBranch: dev
* sourceRoot: /Users/joedonahue/Downloads/BuildLoop/Loop-Dev-220422-1022/LoopWorkspace/Loop
* buildDateString: Sat Apr 23 14:43:10 MST 2022
* xcodeVersion: 13E113

## FeatureFlags
* cgmManagerCategorizeManualGlucoseRangeEnabled: false
* criticalAlertsEnabled: false
* entryDeletionEnabled: true
* fiaspInsulinModelEnabled: true
* lyumjevInsulinModelEnabled: true
* includeServicesInSettingsEnabled: true
* mockTherapySettingsEnabled: false
[...]

Issue #1633, opened 2022-03-12
Dev Branch - Cannot Enter and Save Carbs When There Isn't an Insulin Pump Added and Configured


rolandet:
**Describe the bug**

I was switching from the OmniPod Eros to the OnmiPod DASH using a Dev Branch build from March 2nd and after removing the OmniPod Eros pump from the Configuration settings I was not able to enter Carbs.

**Attach an Issue Report**

**To Reproduce**
1. Without a Pump configured attempt to add Carbs
2. Click "Save..." and it will not let you save the Carb entry. I don't remember the exact error. I'll grab a screen capture the next time if that is needed.

**Expected behavior**
I would expect that you should be able to log your Carbs even if a Pump is not added and configured.

**Screenshots**
I did not capture a screen capture at the time.

**Phone**
 - Hardware: iPhone 12
 - OS Version: iOS 15.3.1

**Loop Version**
 - Version Number: 2.3.0 (57) Built from March 2nd Dev Branch
 - Repo: LoopKit/Loop WorkSpace

**CGM**
 - Device: Dexcom G6
 - Manager app: Dexcom App

**Pump**
 - Manufacturer:  Insulet
 - Model: DASH
 - Firmware version: 3.27.0

**Additional context**

rolandet:
Attaching a screen capture with the message when trying to enter carbs with no pump configured.
![IMG_0437](https://user-images.githubusercontent.com/4328088/160317811-48abe617-5250-4ef8-84dc-8ea30aceef8e.PNG)

marionbarker:
I tested with the latest version of Loop-dev: Workspace commit e68c4e9.

This is still an issue.
I tested with PR #1678 applied and that PR fixes this issue - I will update this use case in that PR.

Use Case for needing to add carbs when no pump is present.
* looper is between pumps for some reason - either no pump is configured, or no pod is paired but wants Loop to continue predicting future BG

PR #1678, opened 2022-05-01
Modify saveWithoutBolus to save Carbs with no active pump

marionbarker:
Edited to add that I believe this fixes the problem reported in Issue #1633 Cannot Enter and Save Carbs when no pump configured.

~In response to Issue #1643 Duplicate carb entries can be saved:~

Add a saveWithoutBolusing function that does not have all the checks (like is bolus already running) that saveAndDeliver requires.

The new saveWithoutBolusing function has been tested in these cases:
* with CGM data, add carb entry, bolus recommended, modify to 0 U - saves carbs and dismisses screen
* missing BG data - tap carb entry:
  * enter FS data, add carb entry, bolus recommended, modify to 0 U - saves carbs and FS data 
* missing BG data - tap manual bolus entry:
  * enter FS data, bolus recommended, modify to 0 U - saves FS data 
* start a large bolus (on test phone)
  * add carb entry, bolus recommended, modify to 0 U - saves carbs and dismisses screen (while bolus is still continuing)


marionbarker:
I could have sworn that I was able to reproduce this error yesterday (before this modification) by having the app delivering insulin when I added carbs and selected Save without Bolusing, but I could not make it fail today.

Perhaps the change proposed [here](https://github.com/Joerg-Schoemer/Loop/commit/89f4837f4be2617daa48c1837fd3cbc340231abc) should be selected instead of or in addition to.

ps2:
Please describe the problem that this is fixing.

marionbarker:
This was in response to [Issue 1643: Duplicate carb entries can be saved](https://github.com/LoopKit/Loop/issues/1643) which happened to me.  The issue is that sometimes `Save without Bolusing` does not work properly - it saves the carbs when you hit it but it continues to show the BolusEntryView screen and allowed me to tap Save without Bolusing again - and saved carbs again.

See also this [zulipchat post](https://loop.zulipchat.com/#narrow/stream/198661-ux/topic/update.20BolusEntryView/near/280815271)

I convinced myself I could reproduce it by starting a bolus and entering carbs and trying to `Save without Bolusing` while the first bolus was still going. 

The next day, I went back and tried to make issue 1643 happen again (with original dev, not with this code) and I could not reproduce it.

At that time, I looked at https://github.com/Joerg-Schoemer/Loop/commit/89f4837f4be2617daa48c1837fd3cbc340231abc and was not sure if that was the correct fix either.

I don't think the code is "wrong" but it may not be necessary. And it may not be sufficient to fix this bug.

marionbarker:
While reviewing open issues, I came across Issue #1633 Cannot Enter and Save Carbs When There Isn't an Insulin Pump Added and Configured.

I believe that this PR solves that issue.
Use Case 1: No Pump Configured
* new Looper might be wanting to use Loop to make predictions while they are learning the interface and manually controlling a pump
* experienced Looper might be switching between insulin delivery methods, but needs to enter carbs in the interim to keep predictions accurate

Use Case 2: 
* Pod Looper is between pods and needs to enter carbs in the interim to keep predictions accurate

Graphic showing success when PR 1678 has been applied when between pods:

PR_1678_fix_SaveWithoutBolusing

marionbarker:
If user wants to save carbs and enter BG without really attaching a pump they can add a simulator (if they want dosing suggestions) or add a pod and cancel at pairing if they just want to save carbs / BG.

PR #1679, opened 2022-05-01
Modify BolusEntryView

marionbarker:
Revise this idea - removing graphic.  Update at end of conversation.

~~I'm not sure if this idea is acceptable, so I have not taken the time to add an actual button with the correct color coding and action. (The placeholder code is just a text box that doesn't respond, i.e., it is a dummy display for illustration purposes. I added the blue box around it with an image editer.)~~

~~The modification is applied to the bolusEntryRow~~
~~* if shouldBolusEntryBecomeFirstResponder is false (as when first displayed), do not show the DismissibleKeyboardTextField, instead display a button at that location similar to what is shown in the graphic below.~~
~~* once the button is tapped, then the display reverts to the current behavior with DismissibleKeyboardTextField showing with a default entry of 0 U.~~

~~Graphic with example text (would need formatting and be updated to be a button with appropriate action).~~

### DIFF
--- a/Loop/View Models/BolusEntryViewModel.swift
+++ b/Loop/View Models/BolusEntryViewModel.swift
@@ -818,7 +818,7 @@ extension BolusEntryViewModel {
         }
     }
     
-    private var hasBolusEntryReadyToDeliver: Bool {
+    public var hasBolusEntryReadyToDeliver: Bool {
         enteredBolus.doubleValue(for: .internationalUnit()) != 0
     }
 
--- a/Loop/Views/BolusEntryView.swift
+++ b/Loop/Views/BolusEntryView.swift
@@ -182,6 +182,10 @@ struct BolusEntryView: View {
             if viewModel.isManualGlucoseEntryEnabled || viewModel.potentialCarbEntry != nil {
                 recommendedBolusRow
             }
+            
+            if viewModel.hasBolusEntryReadyToDeliver && viewModel.potentialCarbEntry != nil {
+                saveCarbsWithoutBolusButton
+            }
 
             bolusEntryRow
         }
@@ -289,32 +293,22 @@ struct BolusEntryView: View {
 
     private var bolusEntryRow: some View {
         HStack {
-            if !shouldBolusEntryBecomeFirstResponder {
-                Text("Accept with Save", comment: "Label for bolus entry row on bolus screen when shouldBolusEntryBecomeFirstResponder is false")
-                Spacer()
-                HStack(alignment: .firstTextBaseline) {
-                    // Instead of Text, insert a button with action of
-                    //  shouldBolusEntryBecomeFirstResponder = true
-                    Text("Edit Bolus")
-                }
-            } else {
-                Text("Bolus", comment: "Label for bolus entry row on bolus screen")
-                Spacer()
-                HStack(alignment: .firstTextBaseline) {
-                    DismissibleKeyboardTextField(
-                        text: enteredBolusStringBinding,
-                        placeholder: Self.doseAmountFormatter.string(from: 0.0)!,
-                        font: .preferredFont(forTextStyle: .title1),
-                        textColor: .loopAccent,
-                        textAlignment: .right,
-                        keyboardType: .decimalPad,
-                        shouldBecomeFirstResponder: shouldBolusEntryBecomeFirstResponder,
-                        maxLength: 5,
-                        doneButtonColor: .loopAccent,
-                        textFieldDidBeginEditing: didBeginEditing
-                    )
-                    bolusUnitsLabel
-                }
+            Text("Bolus", comment: "Label for bolus entry row on bolus screen")
+            Spacer()
+            HStack(alignment: .firstTextBaseline) {
+                DismissibleKeyboardTextField(
+                    text: enteredBolusStringBinding,
+                    placeholder: Self.doseAmountFormatter.string(from: 0.0)!,
+                    font: .preferredFont(forTextStyle: .title1),
+                    textColor: .loopAccent,
+                    textAlignment: .right,
+                    keyboardType: .decimalPad,
+                    shouldBecomeFirstResponder: shouldBolusEntryBecomeFirstResponder,
+                    maxLength: 5,
+                    doneButtonColor: .loopAccent,
+                    textFieldDidBeginEditing: didBeginEditing
+                )
+                bolusUnitsLabel
             }
         }
         .accessibilityElement(children: .combine)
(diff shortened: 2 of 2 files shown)


marionbarker:
To explain a bit more:

My idea is when you are first shown the screen there are 2 obvious buttons.  The big button at the bottom is Save and Bolus - tapping that does the action and closes the screen.  The little button (that used to be a colored bolus value) is proposed to say "Edit Bolus".

I have not coded it yet, but my idea is that when you tap on "Edit Bolus", it sets the bolus value to 0 (instead of recommended); removes the little button and shows the editable 0 U Bolus value, which will cause the big blue button at the bottom to become Save Without Bolusing.

At this point, if you wanted some Bolus but not the recommended amount; you can tap on the Bolus value and enter the desired bolus and Save and Bolus. 

For the case where you want to bolus some but not the recommended amount, it is one extra tap.
For the case where you want to save without bolusing, it is significantly easier.
For the case where you want to bolus the recommended amount, it it unchanged.

marionbarker:
There are several threads in zulipchat on this interface. I will add the last conversation in each thread (at the time of this comments).  

For those not familiar with zulipchat: Tap on the Topic name for the conversation, and zulipchat will show just the one conversation.

* https://loop.zulipchat.com/#narrow/stream/198661-ux/topic/update.20BolusEntryView/near/280873826
* https://loop.zulipchat.com/#narrow/stream/198661-ux/topic/missing.20bolus.20due.20to.20wrong.20time.20entry.20for.20carbs/near/280815323


This one goes back to Aug 2021 - before the automatic zeroing of the bolus field when tapping on it was implemented.  It also has discussion of the difficulty some folks have using the current interface with the Save without Bolusing action. Also - the first report of bug #1643.
* https://loop.zulipchat.com/#narrow/stream/144182-development/topic/Clear.20bolus.20field.20when.20editing/near/275933845

marionbarker:
I had trouble with this initially because I was trying to reuse the existing Save without Bolusing button (see actionButtonAction in BolusEntryViewModel.swift).

This became much simpler when I added a new button called saveCarbsWithoutBolusButton.
This button only shows up when new carbs need to be saved and a bolus is recommended.

New graphics:
When bolus has a non-zero value and carbs are being added, a new button appears between the Recommended Bolus and Bolus lines. Note the Bolus line does not show up on a small phone.

IMG_EC02B212F817-1

This graphic shows the Bolus row after scrolling down.

IMG_18459E257BF7-1

Note: putting this out for comment - I'm still doing testing on it.

marionbarker:
If this is deemed the correct approach and PR #1678 is also accepted, then this code might be modified slightly.

Nightfoxy:
I liked your edit bolus button idea. I could go with either.

ps2:
The graph that's shown is intended to communicate what will happen if you take the suggested action. Having two suggested actions (two large call to action buttons), one of which does *not* implement the shown course of action, is conflicting. Just having two call to action buttons is a design smell. 

If a save without bolusing action button is to be presented, maybe a small "x" next to the bolus amount that clears the input would be better. Then the preview would update, and the action would be just to save the carb entry.

The other option to explore is to have a "save carbs" on the previous page, but that opens up some design questions as well.

marionbarker:
I'm trying to provide an acceptable design change to the Meal Bolus flow that enables single-button tap for these two common actions:
* Save Carbs just entered without an immediate Bolus
* Bolus as recommended and Save Carbs just entered

I use this screen a lot.  My two most likely actions for the meal entry are to save with the suggested bolus or to save carbs without bolusing (and let Loop do my prebolus). There is no change to the design of the Meal Bolus flow when no bolus is recommended or if someone taps on the Bolus Row amount to edit the desired Bolus; or to the Bolus flow (manual bolus is selected from the toolbar).

The zulipchat links posted above show that a lot of people find the current method to save without bolusing to be difficult. I don't think anyone needs to see an updated glucose graph to decide whether to save the carbs without bolusing. Furthermore, the very next screen visible is the home screen which contains the updated glucose projection plot.

It is very important to me that some design change be implemented to make this common action easier than the current design.  I'll keep working with you until we come up with an acceptable solution.

Commit 15430a4 modifies the save carbs without bolusing ActionButtonStyle to .secondary.

Commit 54e7b81 then modifies the placement of the secondary button, which I strongly suspect you will prefer - it does mean that folks with small phone will need to scroll up then tap to choose this action.

Both graphics below show the Meal Entry Flow with the new button:
From left to right:
* Screen seen on a small phone
* Screen seen on a small phone after scrolling up (large phone shows this initially)
* Screen seen after tapping on the Save Carbs without Bolusing button

This version matches Commit 15430a4 with the new button above the Bolus row:

proposed-meal-entry-rev3

This version matches Commit 54e7b81 (current PR) with the new button below the Bolus row:

proposed-meal-entry-rev4

marionbarker:
For completeness.  This graphic is the current design flow for this same action.  Some users have issues with the keyboard.

Meal Entry Flow with the current design:
From left to right:
1. User taps on the Bolus value
2. Bolus value is automatically zeroed and keyboard is displayed
3. User taps on the Done button
4. User taps on the Save without Bolusing button

current-meal-entry-saveWithoutBolusing

Nightfoxy:
I could go either way.
An X or clearing button would be convenient enough for me, especially since
it doesn't bring up the annoying keyboard just to clear the amount. The
keyboard is the most jarring problem. Also, I don't think an option of "0"
is intuitive without the clear button for new people.

I can also see that having 2 buttons complicates the prediction in that its
not accurate. Seeing the prediction with a 0 bolus is a little comforting
for those that understand how the algorithm works. But, I think most people
are doing what Marion describes: all or nothing. Rarely, only a portion is
applied manually.

Trpl7ca:
I like 54e7b81. I rarely look at the prediction graph, probably should, but I do usually make two entries for fast acting carbs where I give about half of the recommended as a pre bolus  and then enter slow acting carbs in the future with a save without blousing. So having the recommended on top of my desired bolus would be my choice. Then the future entry is easy with the Save without blousing option right there.  But either way improvement

britnischipper:
Either of these would be a huge improvement to the current design. The number of times I have hit a number instead of the done on the keyboard and almost bolused an unnecessary and large bolus of insulin is unreal!

aug0211:
I'm not overly informed about this as we do not enter often without bolusing (currently), but I wonder what this would like like with:

1. "0.0u" and radio button left of the recommended amount (1 tap to zero out bolus, not keyboard needed)
2. "Edit" and radio button between "0.0" and recommended buttons; brings up keyboard and allows override of recommended bolus ("Edit" text is replaced with whatever is entered if overridden)

The goal being to:

1. Clearly present the 3 common actions (take recommended bolus - default, give 0.0 units, or manually override)
2. Preserve current single line usage so everything is presented without a need to scroll (reduces risk of someone simply never discovering it... and less annoying to never scroll)

Super rough mock-up I threw together on my phone - I'm sure cleaner designs/fonts/sizes/colors/buttons can be achieved :)
2DE02FEA-2DA9-46B4-9F59-EB7EC817A527

Trpl7ca:
So if you want to save without bolus you would tap 0.0U and then Save and Deliver?

aug0211:
Yep! Maybe the big blue CTA button updates dynamically to say "Save Without Blousing" if you tap the 0.0u radio button - if that is an option/doable.

damonbayer:
I would be interested in a prototype where there was an arrow in the upper right that would let you "save without blousing", like there is for "Carb Entry" in the upper left.

britnischipper:
What if instead on the carb entry screen there was a button that you selected to “save without bolusing” and it simply saved the carbs and closed out and then another button that allowed you to save carbs and move to the normal bolus screen?

marionbarker:
Thanks for all the comments. I plan to work on the carb entry screen in a new PR.

marionbarker:
I'm going to close this because it will not be done this way. 
I don't want to detract from other PR.

Issue #1654, opened 2022-03-27
Remove Pre-Meal Settings


gigich:
It is not possible to remove the settings of pre-Meal. If they could be deleted, it would be passinle to disable pre-meal button. Once set you can’t do that.

PR #1680, opened 2022-05-02
Make Pre-Meal icon inactive if nil

marionbarker:
This modification disables the Pre-Meal icon on the app home screen for the case where the Pre-Meal ranges are not set, i.e., equal to [:].

This change, in conjunction with [LoopKit PR #400: add a reset button to Pre-Meal screen](https://github.com/LoopKit/LoopKit/pull/400) should complete the required changes needed to close Issue #1654


Nightfoxy:
good catch!

ps2:
Thanks! Just want to check that you've tested the transitions; i.e. setting the range and clearing it in a session makes it enable/disable appropriately.

ps2:
Tested, and appears to work correctly. Thanks!

Issue #1683, opened 2022-05-11
Display last loop time instead of a fixed per loop color strings


itsmojo:
The current Loop dev (currently labeled v2.3.0, presumably eventually v3.0) has an updated HUD with less information being displayed (e.g.no last loop time and no last update times for CGM, basal & pump) for a cleaner display. Although I don't actually like this approach in this case as a Loop user (and as someone providing Loop support or wanting to help those who do), it seems like this matter has been already been decided upon and I'm not going to try to rock the boat now. However, I do think that incremental improvements can be made which shouldn't interfere with the new modus operandi.

The green, yellow & red loops are the quick indicator of loop's health. But the fixed strings when the loop is pressed always giving ranges of time is lame and not particularly helpful after the first time you've read all 3 messages.

![original_green_loop](https://user-images.githubusercontent.com/23020111/167765771-7fb504db-0864-4533-abc9-a2222e69c498.PNG)
![original_yellow_loop](https://user-images.githubusercontent.com/23020111/167765788-193996e2-15fe-41cb-9b0a-a52db12f5f98.PNG)
![original_red_loop](https://user-images.githubusercontent.com/23020111/167765803-7fb6a70a-9b10-498b-8826-d1c5e6c57ce1.PNG)

A minimal approach within the current framework can be used where these 3 messages contain the actual time of the last loop instead of using fixed strings. I've created PR #1682 which will display the following messages for example last loop times of 1, 8 and 16 minutes. During this work, I decided to change the yellow loop message title from "Loop Failure" to "Loop Warning" to better differentiate from the red loop message and to more accurately describe Loop's yellow loop state.

![new-01-min-green-loop](https://user-images.githubusercontent.com/23020111/167767478-0570bd84-db8d-492b-926a-ec1fffca9983.PNG)
![new-08-min-yellow-loop](https://user-images.githubusercontent.com/23020111/167767496-094889d1-d165-469e-911f-e324c554aef9.PNG)
![new-16-min-red-loop](https://user-images.githubusercontent.com/23020111/167767515-e4caca12-9e17-41a1-8779-16f8dd6dcd39.PNG)

Another possible extension of this idea would be to include additional information that formerly used to previously be in the HUD displays such as the last CGM and Pump update times in the Loop Completion HUD detail message. This is a possible examples of what this might look like:

[...]

PR #1684, opened 2022-05-11
shorten loop icon messages

marionbarker:
* Shorten the Loop Icon messages and add date-time of last communications. 
* Remove instances of app name in some cases.
* Modify the title to always say Open or Closed loop, therefore do not need to repeat that in the message.

This builds off PR #1682 (already accepted) and the comments of Issue #1683.

Notes: 
1. I kept the Warning and Failure strings to go with the Yellow and Red loop colors.
2. Once the information about last pump message and last CGM messages is accessible, it will be an easy addition to add that to the code.
3. I was not sure whether to add the time-stamp to the large font version or the accessible version.

Next few comments are graphics created with this PR.

### DIFF
--- a/LoopUI/Views/LoopCompletionHUDView.swift
+++ b/LoopUI/Views/LoopCompletionHUDView.swift
@@ -129,6 +129,13 @@ public final class LoopCompletionHUDView: BaseHUDView {
         return formatter
     }()
 
+    private lazy var timeFormatter: DateFormatter = {
+        let timeFormatter = DateFormatter()
+        timeFormatter.dateStyle = .short
+        timeFormatter.timeStyle = .short
+        return timeFormatter
+    }()
+
     @objc private func updateDisplay(_: Timer?) {
         lastLoopMessage = ""
         if let date = lastLoopCompleted {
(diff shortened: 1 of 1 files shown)


marionbarker:
This graphic is with nominal text size for green, yellow and red loops for closed loop (top row) and open loop (bottom row) using this PR.

loop_message_proposed

marionbarker:
This graphic uses the accessible feature of larger text.  Left is Closed Loop, right is Open Loop - both are in the Green Loop state.

loop_message_proposed_large_font

marionbarker:
This graphic was prepared using the updated code.  

* Figure on Left was after turning off the Dash simulator overnight
  * Time since last cycle was > 4 hours, so date and time of last cycle is provided
* Figure on Right was after resuming the Dash simulator
  * Time since last cycle was < 4 hours, so only time of last cycle is provided
loop_message_proposed_vs2

TLH54:
@marionbarker Looks good! Your comments about <>4 hours seem to indicate a programming period I’m not aware of (haven’t looked thru the code, don’t recall from docs or what I’ve reviewed in discussions). Am I reading into your comment, is that a built in time of import, or just an example you used?

marionbarker:
Someone pointed out that you generally don’t need the date displayed. I picked 4 hours as a nice number. If it’s been more than 4 hours since loop completed, a date and time are displayed (like in the original graphics).  If it’s less than 4 hours, just time is displayed. It was a design choice without much thought behind it.

ps2:
This display is not intended to convey whether "communication" completed (what does that even mean?), but whether Loop has been successfully completing its work.  Communication failures are just one kind of failure.

I think the date formatting changes are good. You could update this to just include those, or close this one and create a new ticket with just those changes.

PR #1689, opened 2022-05-15
Update status message for loop icon

marionbarker:
Update status message for loop icon
* add time stamp to last loop message
* if more than 4 hours ago, add date to time stamp
* include time stamp as first line for both open and closed loop when green

For each graphic below:
* Top row is open loop
* Bottom row is closed loop
* Left to right Green, Yellow, Red

First graphic shows 6 different loop icon status messages with current version of dev.

loop-icon-messages-current

Second graphic shows version generated with commit bf3d510.
For this example, the open loop failure message shows the format for > 4 hours since loop last completed. In this case the date is added and the time stamp no longer includes seconds.  (The code uses "ago at" instead of "ago on", so graphic is not exact.)

loop-icon-messages-proposed


marionbarker:
Merged in changes from dev.
I made a few more formatting changes.

This graphic shows 3 views with commit 955c9ff:
Left: Open Loop Message with Manual Temp Basal active
Middle: Open Loop when Yellow with Warning Message
Right: Closed Loop when Red with Failure Message > 4 hours

loop-message-proposed_v2-withMTB

LoopUI/Views/LoopCompletionHUDView.swift
```diff
@@ -150,11 +165,18 @@ public final class LoopCompletionHUDView: BaseHUDView {
...
+                    timeDateString = timeFormatter.string(from: date)
+                } else {
+                    timeDateString = timeDateFormatter.string(from: date)
                 }
+
+                let timeString = String(format: LocalizedString("%1$@ ago at \n%2$@", comment: "Format string describing the time interval since and time of the last completion date. (1: The localized date components, (2: the date"), formatter.string(from: ago)!, timeDateString)
+
+                accessibilityLabel = String(format: LocalizedString("Loop ran %@", comment: "Accessbility format label describing the time interval since and time of the last completion date. (1: last loop time string)"), timeString)
+
+                lastLoopMessage = String(format: LocalizedString("%1$@ last successful loop was\n%2$@", comment: "Last time loop completed message (1: app name) (2: last loop time string)"), Bundle.main.bundleDisplayName, timeString)
```
ps2, line 179: "Loop last successful loop.." is odd grammar. Prefer previous wording here. "Loop last  successfully completed a Loop" or drop the app name: "Last successful loop was ..."

LoopUI/Views/LoopCompletionHUDView.swift
```diff
@@ -150,11 +165,18 @@ public final class LoopCompletionHUDView: BaseHUDView {
...
-                if let fullTimeStr = formatterFull.string(from: ago) {
-                    lastLoopMessage = String(format: LocalizedString("%1$@ last successfully completed a loop %2$@ ago.", comment: "Last loop time completed message (1: app name) (2: last loop time string)"), Bundle.main.bundleDisplayName, fullTimeStr)
+                var timeDateString: String = ""
+                if ago < timeAgoToIncludeDate {
+                    timeDateString = timeFormatter.string(from: date)
+                } else {
+                    timeDateString = timeDateFormatter.string(from: date)
                 }
+
+                let timeString = String(format: LocalizedString("%1$@ ago at \n%2$@", comment: "Format string describing the time interval since and time of the last completion date. (1: The localized date components, (2: the date"), formatter.string(from: ago)!, timeDateString)
```
ps2, line 175: Relative time alone is better for recent Loops, timestamps are more clutter in this case. If you want to put a date string in when last loop is further in the past, that's fine, and at the point we should remove the relative time string.  4 hours is an ok cutoff, imo.

ps2 (commented):
Thanks. Sorry for the delayed review. Please see comments.

marionbarker:
It's been a while since I looked at this.
There have been a number of PR and zulipchat discussions on this topic.
I reviewed the PR - summarized in details below.

Initial response to comments:
* I agree the language got a little confused, I'll fix that.
* I'm not sure I understand the request to remove absolute time - or maybe I'm misunderstanding.
  * The whole point of this PR is to add absolute time to the relative time.
* I agree that when it has been > 4 hours, removing relative time is a good idea.

Rationale for absolute time:
* With one tap, you get the last time loop completed displayed to the nearest second (for most cases)
* This can be useful for trouble shooting while leaving the HUD pristine

Personal Use:
* I have been using this version on my personal phone ever since I posted it
* I key off either relative time or absolute time depending on why I tapped on the icon
* I find both useful at different times
* I never read the words, which is why I didn't even notice the awkward language

## Details:

Related PR:
* Joe added the relative time in PR #1682, which was merged and closed.
* I put in a PR #1684 that had too many changes for you to accept, which I subsequently closed and, based off feedback in that PR, created this current PR.

For ease of reference, repeating the ps2 comments in PR #1684:
* "I think the date formatting changes are good. You could update this to just include those, or close this one and create a new ticket with just those changes."

Related Open Issue:
* Joe opened an issue (#1683), that is not addressed by this PR and is a different topic altogether, just adding it for completeness - not to derail discussion here - any comment on that issue should be posted over there.
  * It requests a time stamp for last CGM and Pump communication in addition to last loop time
  * I don't know how to incorporate that information in the message

marionbarker:
I updated my branch to current dev and modified the last loop message.  The graphics below are Closed Loop messages with commit 8ac0e4e2.
* left: green loop
* middle: yellow loop
* right: red loop > 4 hrs

pr_1689_last_loop_message

marionbarker:
The Open Loop message is different from the Closed Loop when loop is green, but otherwise is identical to the Closed Loop messages. I could add logic to indicate that the yellow / red indications are in open loop for those cases.  But will only do that upon request.

LoopUI/Views/LoopCompletionHUDView.swift
marionbarker: Modified language

LoopUI/Views/LoopCompletionHUDView.swift
marionbarker: See notes below. Alternative code committed.

marionbarker:
Tested with Xcode 14 - found a typo that Xcode 13 let slip through.
Fix typo and pushed commit.

ps2:
I think this is moving in the right direction, but I would still like to see a few changes:
* A loop failure, is still a loop failure, even if it was only one cycle, so I think keeping the title Loop Failure is ok.
* When last loop is fairly recent:
  * Last loop time should be displayed as relative minutes, as it currently is. I don't not feel it's the job of this interface to be providing precise timestamps for Looping. This is just to let the user know how long it's been since the last loop. "Loop last successfully completed a loop X minutes ago".
  * No absolute time string included in this state.
* When last Loop is quite stale (maybe the threshold is > 1 hour?)
  * Display absolute time string.
  * Include AM/PM (latest screenshots seem to be missing this, or are 24 hour time?).
  * Do not include relative time string.
  * Do not include seconds.

marionbarker:
Can we compromise?
* I don't believe I modified warning vs failure but I'll certainly confirm (there are a lot of commits here)
* New proposal
  * If Loop is less than 20 minute old, include only time ago
     - I like minutes with sec but will code as you requested
     - Or maybe sec if < 1 minute, otherwise minutes
  * If Loop is 20 min or greater and less than 4 hours, include time stamp (without time ago) using hr:mm with AM/PM
  * If Loop is greater than 4 hours, include date and time stamp (without time ago) using Date, hh:mm with AM/PM

Should I start a fresh PR with just the agreed upon format (so you won't have 12 commits going through so many version so dev)?

ps2:
@marionbarker That sounds good. Thanks! No need to create a fresh PR. The diff is small enough; if all the commits bother you, you can squash locally and force push to this PR, but they don't bother me; they'll be squashed while merging, regardless.

marionbarker:
I confirmed that the original code uses:
* "Warning - watch for" when aging 
* "Failure - check for" when stale

marionbarker:
There is no difference in the provided message for loopStateView.open true or false when aging or stale.
Let me know if you want that changed to distinguish for loopStateView.

Original code logic:
* switch freshness
* case .fresh
   * if loopStateView.open - open loop message
   * else closed loop message
* case .aging (single "Warning" message + "will continue trying to complete a loop")
* case .stale (single "Failed" message + "will continue trying to complete a loop")

marionbarker:
Updated 30-Nov-2022:

Design:
* Use ago to determine "time portion" of lastLoopMessage
* "Last completed loop "
   * ago <1 min, "<1 min ago"
   * ago <=5 min, "# min ago"
   * ago >=20 min, "# min ago at HH:MM am/pm" 
   * ago >=4 hr, "was at DATE HH:MM am/pm" 

Process:
* Restored this branch back to original code for dev.
* Made minimal changes to provide desired messages.
* Chose to use "<1 min" instead of seconds because ago doesn't update frequently and I figured 
"some" people will poke at the loop repeatedly and say: why does it still say 2 sec.

Graphic:

* First row (after build on blank phone):
    * open or closed Loop, after build on blank phone; same message before and after adding CGM/Pump and before first loop cycle completes
* Left to right, second row (green Loop):
    * open Loop, green
    * closed Loop, green, ago <1 min
    * closed Loop, green, ago >1 and <=5 min
* Left to right, third row, yellow and red, open and closed - same message:
    * open or closed Loop, yellow, ago >5 min
    * open or closed Loop, red, ago >=20 min
    * open or closed Loop, red, ago >=4 hour


Issues to discuss:
* Am I handling the LocalizedString comments correctly for fullTimeStr?
* Any other changes you want?
* I don't understand the caption?.txt portion
* I did not make a change but I noticed lines 212-213 (this commit), under if loopStateView.open, NSLocalizedString is used and everywhere else in file LocalizedString is used.

pr-1689-messages

ps2:
I think 4 hours might be a little early to be bringing in the date, but that's a small tweak if we want to make it later.

Issue #1642, opened 2022-03-17
Loop is doing Remote Data Synchronization of SGVs in dev branch without option enabled


loudestnoise:
**Describe the bug**
As discussed [here](https://loop.zulipchat.com/#narrow/stream/144182-development/topic/Nightscout.20Client.20CGM/near/275252524) in Zulip, users of `dev` branch of Loop are missing the trend arrow in their Nightscout instances. It would appear Loop is uploading the SGV values, despite the option for "Upload Readings" not being enabled in the CGM manager.

**To Reproduce**
1. Build Loop `dev` branch on iPhone.
2. Configure Nightscout as one of your Services.
3. View Nightscout site and observe missing trend arrow. 
4. Check Loop and confirm Upload Readings is not enabled in CGM Manager.

**Expected behavior**
Unless enabled, Nightscout should ingest SGVs from Nightscout user's already configured setting.

**Screenshots**
Dave's example of how if he added a manual value to Apple Health that this was shown in Loop and then uploaded to his Nightscout.
![image](https://user-images.githubusercontent.com/194853/158906655-9dd58905-dfff-4d94-9559-678af92ccb18.png)

**Phone**
 - Hardware: iPhone 13 Mini
 - OS Version: iOS 15.4

**Loop Version**
 - Version Number: v2.3.0 (57)
 - Repo: LoopKit/LoopWorkspace `dev` branch

**CGM**
 - Device: Dexcom G6
 - Manager app: Dexcom App

**Pump**
 - Manufacturer: Insulet
 - Model: Omnipod DASH

**Additional context**
Attached entries.json from ` /api/v1/entries.json` text file from Dave's Nightscout - https://cgms.davediabet.es

[entries.txt](https://github.com/LoopKit/Loop/files/8293455/entries.txt)

gestrich:
A few hints for whenever this can be looked at:

The CGMManager includes a `shouldSyncToRemoteService` bool; however, that flag is not being respected in dev. See the screenshot below for where that flag is read in master. 

In dev, the RemoteDataServicesManager handles carb uploads. That is probably where we need to access the `shouldSyncToRemoteService`. That class does not appear to have a reference to the CGMManager, so a new delegate between RemoteDataServicesManager and DeviceDataManager may help here (RemoteDataServicesManager has a reference to CGMManager)

[...]

PR #1700, opened 2022-05-24
Avoid uploading glucose data to Remote Services when disabled

bill-foreflight:
This is a potential fix for https://github.com/LoopKit/Loop/issues/1642. In the master branch we were only uploading Glucose data when `CGMManager.shouldSyncToRemoteService` returned true. That feature was lost in dev with subsequent refactors.

This change will respect that value but checking just before a glucose upload is performed. I'm not that familiar with the history of this setting so I may be missing the bigger picture to how this is intended to work.

### DIFF
--- a/Loop/Managers/DeviceDataManager.swift
+++ b/Loop/Managers/DeviceDataManager.swift
@@ -353,7 +353,8 @@ final class DeviceDataManager {
         dosingDecisionStore.delegate = self
         glucoseStore.delegate = self
         doseStore.insulinDeliveryStore.delegate = self
-
+        remoteDataServicesManager.delegate = self
+        
         setupPump()
         setupCGM()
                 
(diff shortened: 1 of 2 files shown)


Loop/Managers/RemoteDataServicesManager.swift
```diff
@@ -333,6 +335,11 @@ extension RemoteDataServicesManager {
     }
```
bill-foreflight, line 336: This is the common method I see that uploads glucose data.

Loop/Managers/RemoteDataServicesManager.swift
```diff
@@ -24,6 +24,8 @@ enum RemoteDataType: String {
 final class RemoteDataServicesManager {
 
     public typealias RawState = [String: Any]
```
bill-foreflight, line 26: There was not a reference to cgmManager here so I added a delegate to capture that.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1530,6 +1531,14 @@ struct CancelTempBasalFailedError: LocalizedError {
     }
 }
 
+//MARK: - RemoteDataServicesManagerDelegate protocol conformance
+
+extension DeviceDataManager : RemoteDataServicesManagerDelegate {
+    var shouldSyncToRemoteService: Bool {
+        return cgmManager?.shouldSyncToRemoteService ?? false
```
ps2, line 1538: If we don't have a cgmmanager, but do have glucose values for some reason (unlikely, but possible), then we should probably still upload them.

bill-foreflight: Good point - done.

PR #1705, opened 2022-06-01
Feature Request - add manual additions to insulin in event of pen usage. 

jmp1989:
Please consider allowing manual additions to insulin given.  Currently, I see the ability to delete, but never to add.

**Why**:
Pods may fall off or need to be replaced and on occasion, a new pod may not be able to be placed on immediately.  In these situations, a user may be able to utilize an insulin pen to deliver insulin.  This insulin tracking is then "lost" once a new pod is activated.

**Example**:
Swimming or hiking during the summer.  Current blood sugar is around 200.   Your pod falls off or has a bug.  You aren't home, but fortunately you have an insulin pen and you give yourself 6 units and finish your swim.  By the time you get home, given that you haven't had an active pod and you are still high Loop is recommending 10 units.  You know that you already gave yourself 6 and would like to be able to put that in so the decay curve is properly calculated.


elnjensen:
This feature already exists in the dev branch (meaning that it will be included in the next Loop release). It is also possible to do this in the existing release - if you log an insulin dose in the Health app, Loop will see it and take it into account in its dosing calculations.  See https://loopkit.github.io/loopdocs/build/health/

If you feel that this addresses your need, then please close this request. If not, then add more detail. Thanks!

ps2:
As @elnjensen said, this already exists. Please try it out and report any feedback on zulip. Thanks!

Issue #1644, opened 2022-03-21
v2.3.0 Dev - Correction Range Bar Disappears Using Overrides with Empty Target Values


rfmurphy81:
**Describe the bug**
Throughout the day, I have two different correct ranges. One while awake and one while asleep (slightly higher). Using the v2.3.0 Dev branch from March 9, it appears that when an override is enabled that does not have a target range, the blue bar on the predicted glucose view disappears from view. When using another override with target range values set or disabling an override, the blue bar on the predicted glucose view displays as expected. 

**Attach an Issue Report**
See attached. 

**To Reproduce**
1. Set up an override as just a percentage without any values in tha target range
2. Ensure you have a correct range established and it displays as a blue bar in the predicted glucose view
3. Enable said override

**Expected behavior**
The blue bar for the correction range should not disappear when an override without a target range value is enabled

**Screenshots**
See attached for video. 

**Phone**
 - Hardware: iPhone 13 Pro
 - OS Version: iOS 15.2.1

**Loop Version**
 - Version Number: v.2.3.0
 - Repo: LoopWorkspace Dev from March 9

**CGM**
 - Device: Dexcom G6
 - Manager app: Dexcom App

**Pump**
 - Manufacturer: OmniPod
 - Model: Dash
 - Firmware version: 4.10.0

**Additional context**
N/A

[Loop Report 2022-03-20 19:22:28-07:00.md](https://github.com/LoopKit/Loop/files/8312730/Loop.Report.2022-03-20.19.22.28-07.00.md)

https://user-images.githubusercontent.com/26353970/159198354-8bf47458-7d5c-4a49-b82a-2973ea11a342.mp4

Nightfoxy:
I was about the report this same thing. Nicely documented.
I think the behavior currently in master is sufficient: the existing correction range is boldened or made less opaque on the graph.

jgefaell:
#1620

biesbjerg:
Don't know if it helps, but if `Pre-Meal Preset` is activated before activating a preset, the correction range is shown.

![image](https://user-images.githubusercontent.com/230182/180038324-df753977-2099-4580-8925-41324f9bbf93.png)

marionbarker:
I just built with the "fix" for this. It now shows the correction range, but it is not enhanced with a dark bar - can't tell when override starts and stops from the graph.
LoopWorkspace: 551baa72
Loop: f3e5cbcd

Graphic below - left side before rebuild, right side after rebuild.
[...]

Issue #1620, opened 2022-03-03
DEV: Target indicator not shown when override without a target is selected


jgefaell:
**Describe the bug**
The target indicator on the HUD is not shown when an override that does not specify a target is selected

**Attach an Issue Report**
N/A

**To Reproduce**
1. Select an override that does not specify a target
3. Observe that the target indicator is absent from the HUD for the duration of the override

**Expected behavior**
Activating an override that does not specify a target should display the current target indicator on the HUD.

**Screenshots**

1. Override does not specify a target
![NoTargetSpecified](https://user-images.githubusercontent.com/653168/156482072-2c4f6635-b2b3-4290-a989-8107ada4f572.jpg)

2. The target indicator is absent from the HUD for the duration of the override
![NoTargetShown](https://user-images.githubusercontent.com/653168/156482082-e704bd0f-8ab5-419b-a753-2d75595a65d7.jpg)

**Phone**
 - Hardware: iPhone Xs Max
 - OS Version: 15.2.1

**Loop Version**
 - Version Number: 2.3.0
 - Repo: dev

```
* gitRevision: 4b1951e
* gitBranch: N/A
* workspaceGitRevision: 9363702
* workspaceGitBranch: dev
```

**CGM**
 - Device: Dexcom G6
 - Manager app: Dexcom App

**Pump**
 - Manufacturer: Insulet
 - Model: Dash (TWI)  
 - Firmware version: 4.10.0

**Additional context**
N/A

marionbarker:
Thanks for posting this Jon.  I noticed this as well but never got around to creating an issue.
In my case - when I hit the 3rd day pod override (which is indefinite in duration) I can wind up with no correction range showing up at all on my HUD.

jgefaell:
This behavior is still present 

## Build Details
* appNameAndVersion: T1Pal Loop Dev v2.3.0 (99)
* profileExpiration: 2023-06-03 12:12:26 +0000
* gitRevision: 6e7f0e1
* gitBranch: N/A
* workspaceGitRevision: 0bdc76c
* workspaceGitBranch: dev
* sourceRoot: /Users/jongefaell/src/LoopDevSwitcher/Loop
* buildDateString: Fri Jun  3 08:23:24 EDT 2022
* xcodeVersion: 13E500a

ps2:
This should be fixed in dev. Please re-open with screenshots and details if you see this on current dev.

jgefaell:
I confirm that this is corrected. Thank you!

PR #1756, opened 2022-07-23
Fix missing range bar in glucose chart

biesbjerg:
Fixes missing range bar in the glucose chart when an override without a target range is active.

Reported in https://github.com/LoopKit/Loop/issues/1620, https://github.com/LoopKit/Loop/issues/1644


ps2:
Thanks!

Issue #1741, opened 2022-07-06
activating pre-meal shows incorrect start of override


Joerg-Schoemer:
**Description**
When pre-meal is active and an override is scheduled, the start of the override starts at the end of the pre-meal even when the scheduled start is 1h after the end of the pre-meal.

**To Reproduce**
1. Go to 'overrides'.
2. Create a scheduled override to start in two hours.
3. Activate pre-meal
4. The start of the override is drawn right after the pre-meal

**Expected behavior**
The start of the override should be drawn to start in two hours.

**Screenshots**
scheduled override:
![before activating pre-meal](https://user-images.githubusercontent.com/8568035/177558942-b5b20591-98bf-4c0e-9da5-8b03a3dd9e63.jpeg)

after activating pre-meal:
![after activating pre-meal](https://user-images.githubusercontent.com/8568035/177559154-80b48a85-67b4-4906-b079-74c288013628.jpeg)

**Phone**
 - Hardware: iPhone 13 mini
 - OS Version: 15.5

**Loop Version**
 - Version Number: v2.3.0 (57)
 - Repo: LoopKit/Loop

biesbjerg:
Hello @Joerg-Schoemer 

I've created a pull request (https://github.com/LoopKit/Loop/pull/1757) fixing the reported issue :-)

PR #1757, opened 2022-07-27
Fix pre-meal + scheduled override start date

biesbjerg:
Fixes https://github.com/LoopKit/Loop/issues/1741

### DIFF
--- a/LoopUI/Charts/PredictedGlucoseChart.swift
+++ b/LoopUI/Charts/PredictedGlucoseChart.swift
@@ -107,7 +107,7 @@ extension PredictedGlucoseChart {
 
                 if displayedScheduleOverride != nil {
                     if displayedScheduleOverride!.scheduledEndDate > preMealOverride.scheduledEndDate {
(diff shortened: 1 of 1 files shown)


LoopUI/Charts/PredictedGlucoseChart.swift
```diff
@@ -107,7 +107,7 @@ extension PredictedGlucoseChart {
 
                 if displayedScheduleOverride != nil {
                     if displayedScheduleOverride!.scheduledEndDate > preMealOverride.scheduledEndDate {
-                        displayedScheduleOverride!.scheduledInterval = DateInterval(start: preMealOverride.scheduledEndDate, end: displayedScheduleOverride!.scheduledEndDate)
+                        displayedScheduleOverride!.scheduledInterval = DateInterval(start: displayedScheduleOverride!.startDate, end: displayedScheduleOverride!.scheduledEndDate)
```
ps2, line 110: Thanks for the PR. I think this makes sense for when the scheduled override starts after premeal ends, but what about the case where the scheduled override starts *before* premeal ends?  I think this will draw a section of the timeline that shows two target ranges, which would be confusing. Perhaps the max of (premeal.end and scheduled.start) should be used here.

PR #1771, opened 2022-08-13
Allow Remote Carb Entries in Past or Future

gestrich:
The Careportal includes an Event Time field when entering remote carbs. That date is not included in the Push Notification. That has led to some confusion and also a feature request to support remote carb entries in the future or past. 

I have a separate PR for Nightscout to begin including that date ([7512](https://github.com/nightscout/cgm-remote-monitor/pull/7512)) I don't think this Loop PR needs to wait for that one to merge though.

This Loop PR will parse the "created-date" from the push notification when it is included (when a user has the code in PR 7512).

Screen Shot 2022-08-13 at 1 03 35 PM

### DIFF
--- a/Loop.xcodeproj/project.pbxproj
+++ b/Loop.xcodeproj/project.pbxproj
@@ -386,6 +386,7 @@
 		A9DF02CB24F72B9E00B7C988 /* CriticalEventLogTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A9DF02CA24F72B9E00B7C988 /* CriticalEventLogTests.swift */; };
 		A9DFAFB324F0415E00950D1E /* CarbBackfillRequestUserInfoTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A9DFAFB224F0415E00950D1E /* CarbBackfillRequestUserInfoTests.swift */; };
 		A9DFAFB524F048A000950D1E /* WatchHistoricalCarbsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A9DFAFB424F048A000950D1E /* WatchHistoricalCarbsTests.swift */; };
+		A9E8A80528A7CAC000C0A8A4 /* RemoteCommandTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A9E8A80428A7CAC000C0A8A4 /* RemoteCommandTests.swift */; };
 		A9F5F1F5251050EC00E7C8A4 /* ZipArchiveTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A9F5F1F4251050EC00E7C8A4 /* ZipArchiveTests.swift */; };
 		A9F66FC3247F451500096EA7 /* UIDevice+Loop.swift in Sources */ = {isa = PBXBuildFile; fileRef = A9F66FC2247F451500096EA7 /* UIDevice+Loop.swift */; };
 		A9F703732489BC8500C98AD8 /* CarbStore+SimulatedCoreData.swift in Sources */ = {isa = PBXBuildFile; fileRef = A9F703722489BC8500C98AD8 /* CarbStore+SimulatedCoreData.swift */; };
@@ -440,7 +441,6 @@
 		C13BAD941E8009B000050CB5 /* NumberFormatter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 43BFF0B31E45C1BE00FF19A9 /* NumberFormatter.swift */; };
 		C13DA2B024F6C7690098BB29 /* UIViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = C13DA2AF24F6C7690098BB29 /* UIViewController.swift */; };
 		C148CEE724FD91BD00711B3B /* DeliveryUncertaintyAlertManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = C148CEE624FD91BD00711B3B /* DeliveryUncertaintyAlertManager.swift */; };
-		C1549B782837DF4E002B190C /* LoopAlertManagerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = C1549B772837DF4E002B190C /* LoopAlertManagerTests.swift */; };
 		C159C825286785E000A86EC0 /* LoopUI.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 4F75288B1DFE1DC600C322D6 /* LoopUI.framework */; };
 		C159C828286785E100A86EC0 /* LoopKitUI.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = C159C8192867857000A86EC0 /* LoopKitUI.framework */; };
 		C159C82A286785E300A86EC0 /* MockKitUI.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = C159C8212867859800A86EC0 /* MockKitUI.framework */; };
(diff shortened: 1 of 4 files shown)


Loop.xcodeproj/project.pbxproj
```diff
@@ -440,7 +441,6 @@
 		C13BAD941E8009B000050CB5 /* NumberFormatter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 43BFF0B31E45C1BE00FF19A9 /* NumberFormatter.swift */; };
 		C13DA2B024F6C7690098BB29 /* UIViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = C13DA2AF24F6C7690098BB29 /* UIViewController.swift */; };
 		C148CEE724FD91BD00711B3B /* DeliveryUncertaintyAlertManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = C148CEE624FD91BD00711B3B /* DeliveryUncertaintyAlertManager.swift */; };
```
gestrich, line 443: This appears to be a dead file reference that Xcode is cleaning up since I touched the project file.

Loop.xcodeproj/project.pbxproj
```diff
@@ -386,6 +386,7 @@
 		A9DF02CB24F72B9E00B7C988 /* CriticalEventLogTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A9DF02CA24F72B9E00B7C988 /* CriticalEventLogTests.swift */; };
 		A9DFAFB324F0415E00950D1E /* CarbBackfillRequestUserInfoTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A9DFAFB224F0415E00950D1E /* CarbBackfillRequestUserInfoTests.swift */; };
 		A9DFAFB524F048A000950D1E /* WatchHistoricalCarbsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A9DFAFB424F048A000950D1E /* WatchHistoricalCarbsTests.swift */; };
```
gestrich, line 388: This appears to be a dead file reference that Xcode is cleaning up since I touched the project file.

Loop.xcodeproj/project.pbxproj
gestrich, line 443: This appears to be a dead file reference that Xcode is cleaning up since I touched the project file.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1291,100 +1291,110 @@ extension Notification.Name {
 extension DeviceDataManager {
     func handleRemoteNotification(_ notification: [String: AnyObject]) {
```
gestrich, line 1292: I was trying to cleanup this function by removing the deep nesting, hence the additions of defer and a few new guards. 

The only real logical change is RemoteCommand now throws. That allows for improved logging when we fails to create a RemoteCommand, rather than just returning nil.

Loop/Models/RemoteCommand.swift
```diff
@@ -49,7 +49,7 @@ enum RemoteCommand {
 
 // Push Notifications
 extension RemoteCommand {
```
gestrich, line 51: Allow the "now date" to be optionally injected to make this method testable.

Loop/Models/RemoteCommand.swift
```diff
@@ -70,10 +70,27 @@ extension RemoteCommand {
                 absorptionTime = TimeInterval(hours: absorptionOverride)
             }
             let quantity = HKQuantity(unit: .gram(), doubleValue: carbsValue)
-            let newEntry = NewCarbEntry(quantity: quantity, startDate: Date(), foodType: "", absorptionTime: absorptionTime)
+
```
gestrich, line 73: Parse the created-date. I'm using the same date format as used for the expire date in notifications. 

I'm making this field optional for backwards compatibility with NS versions that don't pass this value. But when it is there, this expects it to have the right format.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1291,100 +1291,110 @@ extension Notification.Name {
...
                 }
-                // Wait up to 25 seconds for uploads triggered by these commands to finish
-                let _ = remoteDataServicesManager.waitForUploadsToFinish(timeout: .now() + TimeInterval(25))
-            } else {
-                log.error("Unhandled remote notification: %{public}@", String(describing: notification))
             }
         }
-        log.info("Finished handling remote notification")
+        // Wait up to 25 seconds for uploads triggered by these commands to finish
+        let _ = remoteDataServicesManager.waitForUploadsToFinish(timeout: .now() + TimeInterval(25))
```
gestrich, line 1397: Unhandled notifications errors will now be thrown by the RemoteCommand initializer.

LoopTests/Models/RemoteCommandTests.swift
```diff
@@ -0,0 +1,107 @@
...
+
+class RemoteCommandTests: XCTestCase {
+
+    override func setUpWithError() throws {
+    }
+
+    override func tearDownWithError() throws {
+    }
+    
+
```
gestrich, line 21: Added RemoteCommand tests, just for the feature I am changing - Carb entries. This could be a good place to add the other tests as we make more changes.

Loop/Models/RemoteCommand.swift
```diff
@@ -70,10 +70,27 @@ extension RemoteCommand {
                 absorptionTime = TimeInterval(hours: absorptionOverride)
             }
             let quantity = HKQuantity(unit: .gram(), doubleValue: carbsValue)
-            let newEntry = NewCarbEntry(quantity: quantity, startDate: Date(), foodType: "", absorptionTime: absorptionTime)
+            
+            var startDate = nowDate
+            if let createdAtStr = notification["created-at"] as? String {
```
ps2, line 75: 'created-at' is generally a term in Loop that we use to record the date that a record was created at, which can be different from the time of eating. The 'startTime' is the the time the user entered, to represent the time of eating.

ps2 (commented):
Thanks! And thanks for the tests. Just a few comments.

Loop/Managers/DeviceDataManager.swift
gestrich: Just some background on my usage of the throwing paradigm -- I've been adopting Swift Concurrency lately where throwing methods seem useful for simpler control flow down the asynchronous stack. i.e. Just `try` failable things in throwing methods and don't check for the return results until you must show an error to the user. It has been nice to not need to check/unwrap things, but only when you don't care why it fails, you just need to know it succeeded. That has been seeping into my non-concurrency world too like this PR

But I can see the rationale for using Result at times too (as I think is what you meant by returning the Error). I think the type guarantees you are suggesting are around using exhaustive switches on the exact error type to maybe take the next best action based on what kind of error happened.

I went ahead and updated this PR to use Result which returns the Error as I think you were suggesting, just to see how that looks for comparison. I did not end up leveraging it here yet as I am just logging the error as you mentioned but it is interesting to see the difference.

Loop/Models/RemoteCommand.swift
```diff
@@ -70,10 +70,27 @@ extension RemoteCommand {
...
+
+            let newEntry = NewCarbEntry(quantity: quantity, startDate: startDate, foodType: "", absorptionTime: absorptionTime)
             self = .carbsEntry(newEntry)
         } else {
-            return nil
+            throw RemoteCommandError.unhandledNotication("\(notification)")
         }
     }
+    
+    enum RemoteCommandError: LocalizedError {
```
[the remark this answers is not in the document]
gestrich: Yeah I missed that namespace was already being used at the top of the file. I updated to indicate this is for command creation/parsing. I thought about using the same error type for everything but it feels like the notification parsing and command delivery errors are different domains.

Loop/Models/RemoteCommand.swift
gestrich: Thanks - This has been updated.

gestrich:
Just an update - The Nightscout PR has merged for this one.

PR #1798, opened 2022-09-19
New Widget

SwiftlyNoah:
Open to suggestions, beta testing, and feedback!


SmallStatusWidget/SmallStatusWidget.swift
```diff
@@ -0,0 +1,239 @@
...
+            .foregroundColor(.primary)
+            
+            
+            HStack(alignment: .center) {
+                BasalView(entry: entry)
+                    .frame(maxWidth: .infinity, alignment: .center)
+                
+                VStack {
+                    if let eventualGlucose = entry.eventualGlucose {
+                        Text("Ev \(eventualGlucose)")
```
ps2, line 207: "Ev" is too cryptic. Please update design to either say "Eventually", or "In x hours", or "Predicted". Something that conveys what this is, and add units.

SmallStatusWidget/GlucoseView.swift
```diff
@@ -0,0 +1,84 @@
...
+import HealthKit
+
+struct GlucoseView: View {
+    @Environment(\.sizeCategory) var sizeCategory
+    var entry: SmallStatusEntry
+    
+    var body: some View {
+        VStack(alignment: .center, spacing: 0) {
+            HStack(spacing: 2) {
+                if let quantity = entry.currentGlucose?.quantity,
```
ps2, line 20: This should be checking the age of the glucose, and indicating staleness. On the watch, we use three dashes ('---') .

SmallStatusWidget/LoopCircleView.swift
```diff
@@ -0,0 +1,50 @@
...
+
+import SwiftUI
+import LoopCore
+
+struct LoopCircleView: View {
+    var entry: SmallStatusEntry
+
+    var body: some View {
+        let closeLoop = entry.closeLoop
+        let lastLoopCompleted = entry.lastLoopCompleted ?? Date()
```
ps2, line 17: Should not assume fresh loop, if nothing was stored in UserDefaults for lastLoopCompleted.

ps2 (changes requested):
Looking great, Noah! Thanks for your work on this. Sorry it took so long to review. I went through it this morning, and while there are a few things I would like to see addressed, the overall implementation is good, and nice work re-implementing some of the existing UIKit views in SwiftUI!

Loop/Managers/LoopDataManager.swift
```diff
@@ -423,6 +424,11 @@ final class LoopDataManager {
         dosingDecisionStore.storeDosingDecision(dosingDecision) {}
 
         NotificationCenter.default.post(name: .LoopCompleted, object: self)
+
```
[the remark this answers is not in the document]
SwiftlyNoah: Done.

SmallStatusWidget/BasalView.swift
```diff
@@ -0,0 +1,80 @@
...
+                            .stroke(entry.isOld ? Color(UIColor.systemGray3) : Color("insulin"), lineWidth: 2)
+                    )
+                    .foregroundColor((entry.isOld ? Color(UIColor.systemGray3) : Color("insulin")).opacity(0.5))
+                    .frame(width: 44, height: 22)
+            }
+            
+            if let rate = rate,
+                let rateString = decimalFormatter.string(from: NSNumber(value: rate)) {
+                Text("\(rateString) U")
+                    .font(.caption2)
```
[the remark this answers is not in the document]
SwiftlyNoah: I updated this for each relevant text.

SmallStatusWidget/SmallStatusWidget.swift
```diff
@@ -0,0 +1,239 @@
...
+                    .fill(Color("WidgetSecondaryBackground"))
+            )
+            
+            HStack(spacing: 2) {
+                if entry.isOld {
+                    Image(systemName: "arrow.clockwise")
+                        .font(.system(size: 10))
+                }
+                
+                Text("\(entry.minsAgo) min\(entry.minsAgo != 1 ? "s" : "") ago")
```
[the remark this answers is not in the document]
SwiftlyNoah: Done. Refreshing now goes as follows: the Timeline has 3 entries, the current one, one for when the BG becomes stale, and for when the widget doesn't update on a loop (the whole widget becomes grayed).

SmallStatusWidget/SmallStatusWidget.swift
```diff
@@ -0,0 +1,239 @@
...
+    }
+    
+    func getTimeline(in context: Context, completion: @escaping (Timeline<Entry>) -> ()) {
+        update { newEntry in
+            var entries = [newEntry]
+            
+            let midnight = Calendar.current.startOfDay(for: Date())
+            let nextHour = midnight.addingTimeInterval(.hours(1))
+            
+            for minute in 1..<60 {
```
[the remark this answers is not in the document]
SwiftlyNoah: Done, see above. Rather than marking the loop circle as yellow or red when it could potentially be green, the whole widget is grayed if it doesn't refresh on a loop. The only rendered frames are the initial one, the stale glucose, and the stale widget.

SmallStatusWidget/SmallStatusWidget.swift
```diff
@@ -0,0 +1,239 @@
...
+//
+
+import WidgetKit
+import SwiftUI
+import LoopKit
+import LoopCore
+import HealthKit
+import CoreData
+
+class Provider: TimelineProvider {
```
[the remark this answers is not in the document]
SwiftlyNoah: I never bothered to change it, this is fixed :)

SmallStatusWidget/LoopCircleView.swift
SwiftlyNoah: Done, it now shows a red loop if nothing was stored in UserDefaults for lastLoopCompleted.

SmallStatusWidget/SmallStatusWidget.swift
```diff
@@ -0,0 +1,239 @@
...
+            }
+            
+            // Making sure that last glucose is not old
+            if let currGlucose = currentGlucose, currGlucose.startDate.addingTimeInterval(LoopCoreConstants.inputDataRecencyInterval) < Date() {
+                currentGlucose = nil
+                previousGlucose = nil
+            }
+            
+            // Making sure that previous glucose is within 15 mins of last glucose to avoid large deltas on sensor changes, missed readings, etc.
+            if let prevGlucose = previousGlucose, prevGlucose.startDate.addingTimeInterval(.minutes(30)) < Date() {
```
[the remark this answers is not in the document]
SwiftlyNoah: Done, now delta will only be shown if the readings are between 4 and 6 minutes apart.

SmallStatusWidget/SmallStatusWidget.swift
SwiftlyNoah: IMG_1337E8FF71B4-1
Is this better?

SmallStatusWidget/GlucoseView.swift
SwiftlyNoah: Stale glucose will now result it the glucose quadrant of the widget been grayed.

alisterburt:
@SwiftlyNoah really excited for this feature! FYI, I tried to build with this PR and ran into 
`/Users/alisterburt/programming/LoopWorkspace/Loop/StatusWidget/StatusWidget.swift:166:37: Variable binding in a condition requires an initializer`
image

SwiftlyNoah:
@alisterburt That happened was because I used some syntax that is only available in Swift 5.7; I just fixed it in the latest commit. Thanks for catching this!

SmallStatusWidget/SmallStatusWidget.swift
ps2: "Eventual" might be better, as that was a pre-existing term used in Loop. In any case, probably not all caps.

SwiftlyNoah: Good point, I updated it.

StatusWidget/StatusWidget.swift
```diff
@@ -0,0 +1,264 @@
...
+@main
+struct SmallStatusWidget: Widget {
+    let kind: String = "SmallStatusWidget"
+
+    var body: some WidgetConfiguration {
+        StaticConfiguration(kind: kind, provider: StatusWidgetProvider()) { entry in
+            SmallStatusWidgetEntryView(entry: entry)
+        }
+        .configurationDisplayName("Loop Status Widget")
+        .description("This widget displays your status.")
```
ps2, line 261: This should be more specific. Something like "See your current blood glucose and Loop's dosing status".

StatusWidget/StatusWidget.swift
```diff
@@ -0,0 +1,264 @@
...
+        observationEnabled: false,
+        provenanceIdentifier: HKSource.default().bundleIdentifier
+    )
+
+    func placeholder(in context: Context) -> StatusWidgetEntry {
+        return StatusWidgetEntry(date: Date(), originalDate: Date(), lastLoopCompleted: nil, closeLoop: true, currentGlucose: nil, previousGlucose: nil, unit: .milligramsPerDeciliter, sensor: nil, netBasal: nil, eventualGlucose: nil)
+    }
+
+    func getSnapshot(in context: Context, completion: @escaping (StatusWidgetEntry) -> ()) {
+        update { newEntry in
```
ps2, line 43: If `context.isPreview` is `true`, and we don't have current data from Loop, then we should display mock data, so that we give a good example of the widget for the Widget Gallery.

ps2:
Merged. Thanks @SwiftlyNoah!

PR #1807, opened 2022-09-30
Add "now" line to main charts

marionbarker:
Problem I'm trying to solve:
* In portrait mode, when reviewing the charts, the current glucose is pretty easy to see, but lining that up with Insulin and Carbs can be tricky
  * it sometimes takes me a couple of tries to get the readout (that lists values and times) to show and my finger often covers up the item of interest
*  In landscape mode, I am usually reviewing further back in time (so having now seems less important)
   * In practice, having the now indicator makes it a little easier to orient myself, I think because it "breaks" the horizontal display into history and prediction

Code changes taken from this [zulipchat post](https://loop.zulipchat.com/#narrow/stream/198661-ux/topic/Current.20time.20line/near/300675504) with thanks to Nils Evert for posting the modified files.

First graphic created from my real phone (no overnight carbs) using straight dev.

pr_no_timeline

Second graphic created from my real phone with this code change.

pr_with_timeline

### DIFF
--- a/LoopUI/Charts/COBChart.swift
+++ b/LoopUI/Charts/COBChart.swift
@@ -63,6 +63,10 @@ public extension COBChart {
         // Grid lines
         let gridLayer = ChartGuideLinesForValuesLayer(xAxis: xAxisLayer.axis, yAxis: yAxisLayer.axis, settings: guideLinesLayerSettings, axisValuesX: Array(xAxisValues.dropFirst().dropLast()), axisValuesY: yAxisValues)
 
+        let currentTimeValue = ChartAxisValueDate(date: Date(),  formatter: { _ in "" })
+        let currentTimeSettings = ChartGuideLinesLayerSettings(linesColor: UIColor.systemGray2, linesWidth: 0.5)
+        let currentTimeLayer = ChartGuideLinesForValuesLayer(xAxis: xAxisLayer.axis, yAxis: yAxisLayer.axis, settings: currentTimeSettings, axisValuesX: [currentTimeValue], axisValuesY: [])
+
         if gestureRecognizer != nil {
             cobChartCache = ChartPointsTouchHighlightLayerViewCache(
                 xAxisLayer: xAxisLayer,
@@ -76,6 +80,7 @@ public extension COBChart {
 
         let layers: [ChartLayer?] = [
             gridLayer,
+            currentTimeLayer,
             xAxisLayer,
             yAxisLayer,
             cobChartCache?.highlightLayer,
--- a/LoopUI/Charts/COBChart.swift
+++ b/LoopUI/Charts/COBChart.swift
@@ -64,7 +64,7 @@ public extension COBChart {
         let gridLayer = ChartGuideLinesForValuesLayer(xAxis: xAxisLayer.axis, yAxis: yAxisLayer.axis, settings: guideLinesLayerSettings, axisValuesX: Array(xAxisValues.dropFirst().dropLast()), axisValuesY: yAxisValues)
 
         let currentTimeValue = ChartAxisValueDate(date: Date(),  formatter: { _ in "" })
-        let currentTimeSettings = ChartGuideLinesLayerSettings(linesColor: UIColor.systemGray2, linesWidth: 0.5)
+        let currentTimeSettings = ChartGuideLinesLayerSettings(linesColor: UIColor.carbTintColor, linesWidth: 1.0)
         let currentTimeLayer = ChartGuideLinesForValuesLayer(xAxis: xAxisLayer.axis, yAxis: yAxisLayer.axis, settings: currentTimeSettings, axisValuesX: [currentTimeValue], axisValuesY: [])
 
         if gestureRecognizer != nil {
--- a/LoopUI/Charts/DoseChart.swift
+++ b/LoopUI/Charts/DoseChart.swift
@@ -107,7 +107,7 @@ public extension DoseChart {
         let gridLayer = ChartGuideLinesForValuesLayer(xAxis: xAxisLayer.axis, yAxis: yAxisLayer.axis, settings: guideLinesLayerSettings, axisValuesX: Array(xAxisValues.dropFirst().dropLast()), axisValuesY: yAxisValues)
 
         let currentTimeValue = ChartAxisValueDate(date: Date(), formatter: { _ in "" })
-        let currentTimeSettings = ChartGuideLinesLayerSettings(linesColor: UIColor.systemGray2, linesWidth: 0.5)
+        let currentTimeSettings = ChartGuideLinesLayerSettings(linesColor: colors.insulinTint, linesWidth: 1.0)
         let currentTimeLayer = ChartGuideLinesForValuesLayer(xAxis: xAxisLayer.axis, yAxis: yAxisLayer.axis, settings: currentTimeSettings, axisValuesX: [currentTimeValue], axisValuesY: [])
 
         // 0-line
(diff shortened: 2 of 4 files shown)


ps2:
There already *is* a distinction between past and future on the charts, that is easy to see. And all the current values for IOB, COB, and total delivery, and already displayed directly. Adding the line is ugly, imo.

You say “lining that up with Insulin and Carbs can be tricky”.  *Why* are you trying to line them up?

marionbarker:
If everyone found the difference between past and future on the charts something that is easy to see, there would not be a discussion.
Can you please tell me what I am missing in terms of the Insulin Delivery chart? Where is the distinction on that one example.

ps2:
The glucose chart shows glucose vs forecasted. The other charts show exact IOB/COB/etc at the current time.  I'm just stating what's there. My question about why you're trying to line them up is the main thing I'm asking.

marionbarker:
Use case - see at a glance how long Manual Temp Basal been running and when will it end

Details: Here's an example of the way I use Loop.
* I am a "lazy" looper in that I use pretty standard carb entries for lunch, dinner or snack regardless of the meal composition
   * I always pre-bolus, either using what Loop recommends, or if Loop recommends what I believe is too low an amount, a "fixed" pre-bolus minimum that I keep in my head
   * Because Loop often restricts basal following a meal bolus and before the glucose rise I know is coming, I use manual temp basal (MTB) to enforce a continued basal flow
      * Based on my experience pumping manually, keeping basal flowing makes a given bolus more effective
* I use fairly tight Dexcom alerts to warn me if I need to pay attention
   * No alerts - I ignore Loop 
   * Should my high alert sound, I'll look at Glucose, Insulin Delivery and Active Carbohydrates plots
      * With the "now" line in place, I know exactly where I am with respect to the MTB and don't need to tap on the pump icon to see how many minutes are left in MTB
      * For carb plot, especially if there was an activity that might cause a slope discontinuity, it's helpful to know when that discontinuity occurred with respect to "now". (In the case below, that was a hot shower following a bike ride).
      * I may still tap on the Carb plot to edit the carb entry to a higher number if I think it appropriate.

I created a composite graphic for portrait display by rebuilding with different options:
* Left: dev
* Middle: "now" line for Insulin Delivery and Active Carbohydrates only
* Right: "now" line for all 4 charts (per current commit for this PR)

pr_mtb_example

ps2:
Saying you want to see how much temp basal is left doesn't give me any insight into the importance of this. Please tell me what you're trying to do (end goal). I don't understand *why* you're trying to evaluate how much temp basal is left.

marionbarker:
NOTE: I am already overriding Loop's model at meals using more of a Sugar Surfing approach. (As I said earlier, I am a lazy looper who does not spend much time figuring out exact meal entries and this works well for me).
* I know, from experience, this is an easy way to handle post-prandial glucose with minimal effort on my part.
* By glancing at the slope and "stickiness" of the "high" from the glucose graph, I can make an assessment of "should I keep pushing higher MTB?"
* If yes, then I need an additional tap on pump icon to see how much time is left on the MTB and decide whether to cancel/restart MTB while phone is in my hand, or let it ride.

The other half of this is that I will sometimes add another bolus to help out what might not have been a big enough prebolus. Seeing the now line with respect to the last bolus triangle is helpful to assess if that is a good idea, especially with AB (when not using MTB) that may have been helping me silently.

jonfawcett:
I switch over to Loop Follow to make these types of decisions because it can't be quickly and easily visualized in Loop. Personally, I think this needs to go beyond just the now line change and the on-the-hour lines in Loop should be removed. Decision making isn't based on the actual hour, it is based on right now and intervals from now. This is made even more difficult without a now line because the large time period on screen makes the 5 min intervals incredibly small. The ways I use LF applies to Loop for this, so here are the details. LF has a now line and incrementing 1 hour lines prior to now back through the 6 hours of DIA. Each line diminishes in a more subtle dash to quickly visualize how far back in time it was. In addition, I usually display only 1 hour of prediction because it allows more of the prior time period to be on screen at once, which is more important to inform current decisions. I usually have 3 hours on screen at once. 2 hours past and 1 hour future which gives enough space between treatments to quickly gauge time from now and time between each treatment item. eg. you can quickly gauge that something is halfway between now and 1 hour in the past.

Examples/use cases:
- prebolus and low treatments aren't relevant to the actual hour. They are based on how long ago it was. 
- The "when" from IOB and COB is as important as the total. Current treatment decisions will be different if that was a single meal dose 3 hours ago, 2 smaller meal doses over time, a very recent meal dose, a correction, etc. This is even more important heading into sports practice or games.
- With a 17yo who often forgets to prebolus and frequently forgets to enter carbs until after she's eaten, it is incredibly important for us to gauge a change to the dose based on how much and when Loop has started issuing AB on the rise.

marionbarker:
Let me summarize when I believe the now line helps on the lower 3 charts:
* I want to evaluate what Loop has done and predicts for future (there are some many instances, living with T1D, where this is important - if you want I can list them in a different place - maybe that is your actual question)
   * Seeing where now is for the lower charts is hard; without the line - I'll drag my finger across the chart to see the readout - but it normally takes me a couple of tries to do this
* I know I did a prebolus, but I forgot to set timer on my watch and I want to ascertain in a glance if it's been long enough before starting to eat
   * without now line - requires a tap on an insulin chart to see times
* I set a manual temp basal (for many reasons) and may want to continue that MTB because of current status
   * without now line - requires a tap on the pump icon and scroll down to see how much MTB is left

Reasons other people like the "now" line (some duplication):
* how long since last meal or hypo treatment (COB), 
* how long since last bolus/increase in IOB
* hypo treatment situations (should I expect the rise to kick in yet?), 
* exercise situations (how long since last insulin, is IOB tapering off significantly now?)

Parent coming from FreeAPS to dev:
* I have to spend a long looking at the screen and trying to figure out where we "are". 
* I miss it [the "now" line] every single day. 
* I have shitty eyesight, and it was really helpful.

Person with diabetes:
"I am old and have older eyes. The vertical bar is a small nod to old timers like myself I’d say. Makes it much easier to line up current time in any single graph let alone all of them at once."

marionbarker:
I agree the lines as shown above could be improved; furthermore, they are hard to see in dark mode. Using graph colors and bigger line-width is easy. Now that I've taken the trouble to look at the code changes, I will be modifying my personal version in the future as shown in the graphic below.

I do not think the glucose "now" line is needed and I found it kind of annoying, so I removed that from this PR.

pr_modified_now_3-charts

ps2:
Thanks @jonfawcett You make some good points, and I could see a relative time axis being a better and cleaner solution than a stark line in the graph.  This might make sense as a setting. Generally Loop is trying to help people not have to worry about the kinds of questions that you all are providing as reasons for wanting this graph, but I do understand that a lot of you are going to be digging into the details all the time. If this was submitted as a full implementation of a relative time axis option (i.e. designed and not experiment-like), I think that could work.

jonfawcett:
I only had a few minutes this weekend to look at the PR and the code. It looks like you would need now-8 or 9 with 14-15 total values to handle landscape view. The first question is if the labels should remain on the hour even with the grid shifted. Loop Follow does not link the time labels with the hour lines and I have never had anyone comment either way about it.

To do it with even hour labels and a shifted grid, I think you may want to edit the existing grid layer x axis style to remove, hide, or give a blank line style to the grid lines for X. Then change your new layer to be created from the array starting now-8 to now +14 (or however many are needed). 

Alternatively, if the labels should be displayed based on the relative time, it looks like you would want to change the incoming xaxisvalues array for the grid layer. I didn't have time to trace where that is coming from but it should be easy enough to track down. This might get cluttered by needing the :xx in the time

Both of these would also need added to the main chart, which doesn't have anything from the PR right now.

ps2:
The labels on the time axis should reflect the division markers.

bjornoleh:
As an alternative to the “now” line, perhaps a dashed trace of the future part of the curves could be used, in a similar style as the glucose chart? Perhaps with a different fill of the areas under the curves too? Just an idea, perhaps it’s something that is easy to test? 

Agree with the relative time comments from Jon, this is what really matters. Not what time it is or time since one or two o’clock, but how long since treatment x or y.

marionbarker:
Closing this PR. Opened PR #1810.

PR #1810, opened 2022-10-03
Add Phone Setting for Loop to enable main charts now indicator

marionbarker:
This is an "ugly" implementation of using settings on the phone to enable the now indicator on the main Loop charts. Starting point for discussion.

Two issues:
* not sure if I did the settings in the best way
* I wasn't sure how to do the change suggested by Jon Fawcett in #1807 
   * I went with the design of the now indicator from that PR
   * I simply changed the line widths to 0 if not enabled in Settings->Loop->Add Now Marker, Main Charts

I did the 0 linewidth because I could not figure out the `let layers: . . .` code with currentTimeLayer not being present.


marionbarker:
To clarify - I'm looking at Jon's suggestion as a separate task - not ignoring it.
Is this method of making something an option acceptable for Loop design.

marionbarker:
Pete requests the setting switch be in Loop settings. Will revise and resubmit.

PR #1670, opened 2022-04-12
Added new strings and translations (Norwegian Bokmål)

jannyg:
Added some new strings and translations missing. There's still a lot missing, but it's a start I guess. I'm not a programmer and not familiar with Swift, but it seem to me like there are some strings not defined with "NSLocalizedString"?

PR #1813, opened 2022-10-14
Remove faulty string from nb Localizable.strings

bjornoleh:


bjornoleh:
The faulty string broke the localisation of the meal amounts and absorbed carbs from the carbs screen.

Broken:
Pasted Graphic 2

Corrected by removing the below:
/* Format string for combining localized numeric value and unit. (1: numeric value)(2: unit) */
"QUANTITY_VALUE_AND_UNIT" = "QUANTITY_VALUE_AND_UNIT";
Phone 14

The relevant string is already defined elsewhere, and should not be in Loop/nb.lproj/Localizable.strings:

LoopKit/Loop/Common/en.lproj/Localizable.strings:
/* Format string for combining localized numeric value and unit. (1: numeric value)(2: unit) */
"QUANTITY_VALUE_AND_UNIT" = "%1$@ %2$@";

The faulty string was introduced in https://github.com/LoopKit/Loop/pull/1670, together with a number of "Customs strings added from manually debugging". These strings are added without updating Base.lproject, making the nb translation quite different from the rest. Perhaps this is not the best way to do this, and hopefully someone can pick up the Crowdin Loop translation again. If so, a review of https://github.com/LoopKit/Loop/pull/1670 might be in order.

bjornoleh:
I added a small typo fix to this PR as suggested on Zulip (insuling —> insulin). 

@ps2 , this PR should be ready for merging now.

jannyg (approved):

PR #1825, opened 2022-11-08
Add missed meal notifications to Loop

novalegra:
This PR and the associated LoopKit PR (TODO: add PR link) add missed meal notifications to Loop. 

## Motivation
It's easy to forget to log a meal, and the Loop algorithm works best when it has information about all the carbs and insulin that could affect blood glucose levels. This PR uses the insulin dosing information in Loop to automatically detect when unexpected rise in BG is so large that the only likely explanation is an unannounced meal (UAM), then notify the user so they can log the meal and administer insulin if needed.

## Changes Made
The changes in Loop were primarily around managing the delivery and response to UAM notifications; the main UAM detection algorithm is in LoopKit. 

Main changes:
1. Add UAM notification generation call to the `update` function in `LoopDataManager` (since it's the main driver of the Loop algorithm updates, I'm open to putting it elsewhere if there's a better place)
2. Add switch in Alert Permissions to enable/disable missed meal notifications
3. Update the watch carb/bolus flow to allow it to be initialized with a carb entry quantity and time
4. Add tests for the changes made


Loop/View Models/AlertPermissionsViewModel.swift
```diff
@@ -0,0 +1,41 @@
...
+//  AlertPermissionsViewModel.swift
+//  Loop
+//
+//  Created by Anna Quinlan on 11/7/22.
+//  Copyright © 2022 LoopKit Authors. All rights reserved.
+//
+
+import SwiftUI
+
+public class AlertPermissionsViewModel: ObservableObject {
```
novalegra, line 11: I pulled this out into a view model since it seemed more in line with what we do for other settings views

LoopTests/Mock Stores/HKHealthStoreMock.swift
```diff
@@ -0,0 +1,82 @@
+//
+//  HKHealthStoreMock.swift
```
novalegra, line 2: I pulled this in so an actual (non-mock) carb store could be used to calculate carb effects for the missed meal detection unit tests instead of loading the carb effects via fixtures

Loop/Managers/Missed Meal Detection/MealDetectionManager.swift
```diff
@@ -0,0 +1,311 @@
...
+                    missedMealTimeline.append((pastTime, unexpectedDeviation, nil, nil))
+                    continue
+                }
+                
+                /// Find the threshold based on a minimum of `missedMealGlucoseRiseThreshold` of change per minute
+                let minutesAgo = now.timeIntervalSince(pastTime).minutes
+                let deviationChangeThreshold = MissedMealSettings.glucoseRiseThreshold * minutesAgo
+                
+                /// Find the total effect we'd expect to see for a meal with `carbThreshold`-worth of carbs that started at `pastTime`
+                guard let modeledMealEffectThreshold = self.effectThreshold(mealStart: pastTime, carbsInGrams: MissedMealSettings.minCarbThreshold) else {
```
ps2, line 162: Would it be simpler to just lookup CSF based on CR and ISF for `pastTime`, and use that to turn minCarbThreshold into an expected glucose change?

Loop/Managers/Missed Meal Detection/MealDetectionManager.swift
```diff
@@ -0,0 +1,311 @@
...
+                    continue
+                }
+                
+                missedMealTimeline.append((pastTime, unexpectedDeviation, modeledMealEffectThreshold, deviationChangeThreshold))
+                
+                /// Use the higher of the 2 thresholds to ensure noisy CGM data doesn't cause false-positives for more recent times
+                let effectThreshold = max(deviationChangeThreshold, modeledMealEffectThreshold)
```
ps2, line 169: It was a little difficult for me to understand why there were two thresholds. It looks like one is rate based (deviationChangeThreshold) and the other is meal size based (modeledMealEffectThreshold).  Maybe a naming change could make this more clear?

Loop/Managers/Missed Meal Detection/MealDetectionManager.swift
```diff
@@ -0,0 +1,311 @@
...
+
+                if unexpectedDeviation >= effectThreshold {
```
ps2, line 171: So I think I understand how the time detection is working. As you go farther back in time, `deviationChangeThreshold` grows, as it's based on `glucoseRiseThreshold * minutesAgo`, When `unexpectedDeviation` falls below this, that's what marks the start of the meal, correct?

Loop/Managers/Missed Meal Detection/MealDetectionManager.swift
```diff
@@ -0,0 +1,311 @@
...
+    
+    // MARK: Meal Detection
+    func hasMissedMeal(insulinCounteractionEffects: [GlucoseEffectVelocity], completion: @escaping (MissedMealStatus) -> Void) {
+        let delta = TimeInterval(minutes: 5)
+
+        let intervalStart = currentDate(timeIntervalSinceNow: -MissedMealSettings.maxRecency)
+        let intervalEnd = currentDate(timeIntervalSinceNow: -MissedMealSettings.minRecency)
+        let now = self.currentDate
+
+        carbStore.getGlucoseEffects(start: intervalStart, end: now, effectVelocities: insulinCounteractionEffects) {[weak self] result in
```
ps2, line 72: It might be a nice separation of concerns to pass in carbEffects, rather than asking carbstore for it, and creating a dependency on carbstore for this detector.

Loop/Managers/Missed Meal Detection/MealDetectionManager.swift
```diff
@@ -0,0 +1,311 @@
...
+                    /// Clamp starts & ends to `intervalStart...now` since our algorithm assumes all effects occur within that interval
+                    let start = max($0.startDate, intervalStart)
+                    let end = min($0.endDate, now)
+
+                    guard let effect = $0.effect(from: start, to: end) else {
+                        let item: GlucoseEffect? = nil // FIXME: we get a compiler error if we try to return `nil` directly
+                        return item
+                    }
+
+                    return GlucoseEffect(startDate: effect.startDate.dateFlooredToTimeInterval(delta),
```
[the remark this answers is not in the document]
ps2: It might not make a big impact in meal detection (might make meals appear to be detected earlier in time than the effects that are triggering the detection). Just thought I'd point out that it's not strictly correct. Dealing with this properly might be complex, since you're wanting to ultimately subtract entered carb effects from ICE, which are both on different timelines; carb effects are on a simulation timeline, and ICE is on the reported glucose timeline.

ps2 (commented):
I don't think there are any blockers here, but I did make several comments as I went through it. Thanks for all your work on this so far! I think we're close to merging it in!

Loop/Managers/LoopDataManager.swift
```diff
@@ -1702,6 +1709,14 @@ extension LoopDataManager {
     /// *This method should only be called from the `dataAccessQueue`*
     private func enactRecommendedAutomaticDose() -> LoopError? {
         dispatchPrecondition(condition: .onQueue(dataAccessQueue))
+        
+        mealDetectionManager.generateMissedMealNotificationIfNeeded(
```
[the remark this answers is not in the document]
novalegra: Sure! I've moved it into `finishLoop`

Loop/Managers/Missed Meal Detection/MealDetectionManager.swift
novalegra: The `deviationChangeThreshold` does grow as we go further back in time, but the `unexpectedDeviation` has to be higher than the `deviationChangeThreshold` to mark the start of the meal. 

Part of the reason this rate-based threshold is helpful is that it applies a higher required `unexpectedDeviation` (and this 'burden of proof') for meals that are further in the past, making it harder for a bunch of random CGM deviations to accumulate to trigger a meal start.

Loop/Managers/Missed Meal Detection/MealDetectionManager.swift
novalegra: I agree it would be good to remove this dependency. 

My only hesitation is that the `carbEffects` that are computed via `update()` in `LoopDataManager` go back ~30 minutes ("last BG time - retrospective correction interval"). Since the missed meal detection wants carb effects that go back 2 hours (`MissedMealSettings.maxRecency`), this would mean needing to call `getCarbEffects` again at the `LoopDataManager` level. This `getCarbEffects` call would need to know the `MissedMealSettings.maxRecency`, which feels like implementation details leaking out of the class. What do you think would be best?

Loop/Managers/Missed Meal Detection/MealDetectionManager.swift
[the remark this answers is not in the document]
novalegra: I agree that it's not ideal to have to translate the ICE into the type of effect that carb effects can be subtracted from. I just played with using a ceiled endDate to assign the GlucoseEffect `startDate` (so for your example, the startDate would be 1:55pm) and ran the unit tests. It actually seems that this change leads to an earlier meal start time being assigned. 

From a safety perspective, which behavior would be preferred - potentially having the missed meal time be too early or too late? I would think too early might be preferred (1) carb digestion & CGM lag already means we're lagging behind and (2) because there would be less remaining effect for Loop to dose for

Loop/Managers/Missed Meal Detection/MealDetectionManager.swift
novalegra: I thought about that and my concern with that approach was the potential for subtle bugs with in terms of handling overrides - I generally tried to use LoopKit's effects functions because then this is handled automatically.

Loop/Managers/Missed Meal Detection/MealDetectionManager.swift
ps2: I think the call again at LDM level is fine. As far as knowing how much data the MissedMeal algorithm needs, and what is considered "implementation details" it depends on if you want maxRecency to be an algorithm configuration option (in which case it is public, and the api documentation should state that the carb effects passed in should cover a range matching the configuration option), or make it a static amount (in which case the docs specify the needed time range that the passed in carbs need to cover).

Loop/Managers/Missed Meal Detection/MealDetectionManager.swift
ps2: LDM does have access to the schedules with recent overrides applied. I would accept this PR with a dependency on CarbStore, but if you're up for it, it does seem a little cleaner, and maybe easier to test and maintain, to keep it free of that dependency.

Loop/Managers/Missed Meal Detection/MissedMealSettings.swift
```diff
@@ -0,0 +1,25 @@
...
+
+public struct MissedMealSettings {
+    /// Minimum grams of unannounced carbs that must be detected for a notification to be delivered
+    public static let minCarbThreshold: Double = 15 // grams
+    /// Maximum grams of unannounced carbs that the algorithm will search for
+    public static let maxCarbThreshold: Double = 80 // grams
+    /// Minimum threshold for glucose rise over the detection window
+    static let glucoseRiseThreshold = 2.0 // mg/dL/m
+    /// Minimum time from now that must have passed for the meal to be detected
+    public static let minRecency = TimeInterval(minutes: 25)
```
ps2, line 19: Other places in Loop, like momentum calculation, we look at an average or regression over multiple samples to reduce noise, and use a shorter period, like 15 minutes.  And from my own personal experience, if I see 3 points all indicating a significant rise, that seems pretty trustworthy.  Does reducing `minRecency` to 15 minutes (or 20 minutes) result in too many false positives in some way?

Loop/Managers/Missed Meal Detection/MealDetectionManager.swift
novalegra: Yeah, I was hoping to make it a configuration option at build-time. I'll update it to use carb effects and make sure the documentation states this assumption

Loop/Managers/Missed Meal Detection/MissedMealSettings.swift
novalegra: I think it's probably OK to set the `minRecency` to 15 minutes. Since the algorithm now looks for direct carb absorption, and the carb threshold is set higher than most noisy CGM jumps, I think this shouldn't result in too many false positives. I'll make the change and would lean towards getting some live testing before merging this to check my assumptions

Loop/Managers/Missed Meal Detection/MealDetectionManager.swift
novalegra: I'm definitely up for it - I want to ensure this is maintainable 😁

Loop/Managers/Missed Meal Detection/MealDetectionManager.swift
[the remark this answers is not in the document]
ps2: Yes, I think making sure we cover more of the time that carbs could have been absorbing is the more conservative approach. This feature in general will turn off RC unless the user enters more carbs than detected, so I imagine it mostly acting as a nudge towards less delivery, even if the start time isn't detected precisely.

Again, I don't think this affects things too much.

ps2:
@novalegra I think this is ready for bringing into dev, if you agree.  Since it is optional and behind I switch, it shouldn't impact too many people, and we can keep refining it on dev if needed.

I was just looking at the enable switch, and think maybe it could be moved up to the "Alert Management" screen, and not be down in permissions, as it's not really a permission option.

ps2:
Once the conflicts with dev are fixed, I'll merge this in!

novalegra:
> @novalegra I think this is ready for bringing into dev, if you agree. Since it is optional and behind I switch, it shouldn't impact too many people, and we can keep refining it on dev if needed.
> 
> I was just looking at the enable switch, and think maybe it could be moved up to the "Alert Management" screen, and not be down in permissions, as it's not really a permission option.

That sounds good to me! I've merged in `dev` and moved up the switch

ps2:
Thanks @novalegra!

PR #1826, opened 2022-11-10
limit maxCarbs to 99g

marionbarker:
Reason for the change: Safety / prevent accidental entry of large number of carbs.

We discussed, earlier this year, adding a setting to Loop Therapy Settings to enable the user to adjust maxCarb for their app.

However, I checked my notes - I tried to work on this but did not succeed (my local branch name for both Loop and LoopKit is "dev_not_working_add_maxCarb_to_deliverylimits" with a date of April 30, 2022).

Details:
I think this is an important safety consideration (several mentors have helped people who accidentally got too much insulin by adding an extra digit to the carb entry, e.g., 115 instead of 15). I'd rather have the 2-digit protection for everyone and inconvenience those who eat 100 g or more into needing to do more than one entry.

In my mind it is similar to putting the 1 hour limit on future carb entries.  It annoys some folks, but there are work-arounds.

### DIFF
--- a/Loop/View Controllers/CarbEntryViewController.swift
+++ b/Loop/View Controllers/CarbEntryViewController.swift
@@ -33,7 +33,7 @@ final class CarbEntryViewController: LoopChartsTableViewController, Identifiable
         return deviceManager.displayGlucoseUnitObservable.displayGlucoseUnit
     }
 
-    var maxQuantity = HKQuantity(unit: .gram(), doubleValue: 250)
+    var maxQuantity = HKQuantity(unit: .gram(), doubleValue: 99)
 
     /// Entry configuration values. Must be set before presenting.
     var absorptionTimePickerInterval = TimeInterval(minutes: 30)


rteague-dr:
I vote no on this one. I enter 100+ carbs all the time for my meals. At minimum make it configurable.

scottf51:
+1 , the time thing is also something I have to do 3 steps for at least once a day  , I would much prefer that to be configurable too, but if I am the only one then so be it. I get the need for safety guards but could we have an expert mode or something because feels like some people would like to configure and do understand the implications. I could just poke around and edit the code each time , but when building often in dev is becomes a little painful having to keep applying patches. Sent from my iPhoneOn 10 Nov 2022, at 23:44, rteague-dr ***@***.***> wrote:﻿
I vote no on this one. I enter 100+ carbs all the time for my meals. At minimum make it configurable.

mcslader:
I also second the request to make the time thing more configurable. When my son eats pizza or pasta we currently have to set timers to remember to add the carbs for a late rise after the fact. We’d prefer to split the dose up front and just announce the carbs for a later time if possible.Thanks!Sent from my iPhoneOn Nov 10, 2022, at 6:03 PM, scottf51 ***@***.***> wrote:﻿
+1 , the time thing is also something I have to do 3 steps for at least once a day  , I would much prefer that to be configurable too, but if I am the only one then so be it. I get the need for safety guards but could we have an expert mode or something because feels like some people would like to configure and do understand the implications. I could just poke around and edit the code each time , but when building often in dev is becomes a little painful having to keep applying patches. Sent from my iPhoneOn 10 Nov 2022, at 23:44, rteague-dr ***@***.***> wrote:﻿
I vote no on this one. I enter 100+ carbs all the time for my meals. At minimum make it configurable.

scottf51:
You can actually set the carbs 1 hour on future , save and then go back in and move it forward an hour, save , repeat . That’s what I do, but would prefer to just add in 2-3 hour in future first time :) Sent from my iPhoneOn 11 Nov 2022, at 01:33, mcslader ***@***.***> wrote:﻿
I also second the request to make the time thing more configurable. When my son eats pizza or pasta we currently have to set timers to remember to add the carbs for a late rise after the fact. We’d prefer to split the dose up front and just announce the carbs for a later time if possible.Thanks!Sent from my iPhoneOn Nov 10, 2022, at 6:03 PM, scottf51 ***@***.***> wrote:﻿
+1 , the time thing is also something I have to do 3 steps for at least once a day  , I would much prefer that to be configurable too, but if I am the only one then so be it. I get the need for safety guards but could we have an expert mode or something because feels like some people would like to configure and do understand the implications. I could just poke around and edit the code each time , but when building often in dev is becomes a little painful having to keep applying patches. Sent from my iPhoneOn 10 Nov 2022, at 23:44, rteague-dr ***@***.***> wrote:﻿
I vote no on this one. I enter 100+ carbs all the time for my meals. At minimum make it configurable.

ps2:
2 vs 3 digit carb entry errors are dangerous, and do happen. I think a warning may be appropriate when entering 3 digits, but it *is* also common for people to enter > 100  grams appropriately, so it shouldn't be blocked.  A configurable limit might also be appropriate.

PR #1830, opened 2022-11-17
Add ability to estimate delivery duration of bolus

novalegra:
See https://github.com/LoopKit/LoopKit/pull/431/files for a description of the main changes. In `Loop`, I've added a delegate function to `LoopDataManagerDelegate` so the missed meal detection feature can compute the estimated autobolus duration when scheduling a missed meal notification.


ps2:
I think this can stay part of your missed meal  detection PR.

Issue #1817, opened 2022-10-24
Loop cycles complete just before new CGM values are received


damonbayer:
With Dash integration and autobolus, there are some scenarios where dosing decisions are issued >4 minutes after an old CGM value is received and <1 minute before a new CGP value is received. Thus, all dosing decisions are nearly 5 minutes "behind."

See [Zulip discussion](https://loop.zulipchat.com/#narrow/stream/312259-Omnipod-DASH/topic/Looping.20Interval.20with.20Dash) for issue report and proposed solutions.

aug0211:
This issue may limit Loop’s ability to proactively mitigate hypos.

Basal insulin delivery may end up being left on longer than appropriate by missing a proactive 0 temp basal by one Loop cycle.

PR #1835, opened 2022-11-20
Prevent pump driven updates from taking over 5 minute cgm update cycle

ps2:


ps2:
Fixes https://github.com/LoopKit/Loop/issues/1817

Issue #1841, opened 2022-11-27
Remote Carb Duration and Time not limited


marionbarker:
**Describe the bug**
Loop accepts a remote carb entry of any duration; it is not limited as when entering on the Loop phone.
Loop accepts a remote carb entry time that is not limited as when entering on the Loop phone.

**Attach an Issue Report**
N/A

**To Reproduce**
1. With Loop-dev and Nightscout configured for APN
3. Send a remote carb entry (I used 1 g with duration of 300 hours)
4. Observe the entry is accepted on the Looper's phone as sent
5. send an entry 15 hours in the past - it is accepted (Loop limits to -12 hours)
6. send an entry 3 hours in the future - it is accepted (Loop limits to +3 hours)

**Expected behavior**
I expect remote carb to match the limits set by the Loop app.
* The absorption duration should be no less than 30 min and no more the 8 hours, duration is limited to increments of 30 min.(Or whatever limits are set in Loop if these change.)
* The time should be in the range of 12 hours in the past to 1 hours in the future. (Or whatever limits are set in Loop if these change.)

**Additional context**
Perform these tests:
From Nightscout:
* 1g at 300 hr, accepted by Loop
* 1 g at -3.4 hr, accepted by Loop as 3 hour.
  * (invalid number is defaulted to 3 hour)

From LoopCaregiver app:
* 1g at 0 hr, accepted by Loop as 3 hour
  * (invalid number is defaulted to 3 hour)
* 1g at 0.15 hr, accepted by Loop as 9 min
* 1g at 0.5 hr, accepted by Loop as 30 min
* 1g at 3.145 hr, accepted by Loop as 3 hr 8 min
* 1g at 14.35 hr, accepted by Loop as 14 hr, 21 min

Testing with latest Loop-dev commit: 4a9305e and missed-meal-detection branch, commit: a4b1daa on 16 Nov 2022.

PR #1842, opened 2022-11-27
Handle invalid remote carb absorption values

gestrich:
A few changes in this PR:

* Enforce carb absorption range of 30 minutes to 8 hours.
* Use the user's default medium absorption if none supplied (A previous TODO in code).
* Add tests.

This issue was reported in https://github.com/LoopKit/Loop/issues/1841

Sending very large values also seems to make Loop crash or at least cause all carbs to disappear as I seemed to reproduce and also reported by a Loop user recently https://loop.zulipchat.com/#narrow/stream/144182-development/topic/Red.20loop.

### DIFF
--- a/Loop/Managers/DeviceDataManager.swift
+++ b/Loop/Managers/DeviceDataManager.swift
@@ -1337,7 +1337,7 @@ extension DeviceDataManager {
         let command: RemoteCommand
         
         do {
-            command = try RemoteCommand.createRemoteCommand(notification: notification, allowedPresets: loopManager.settings.overridePresets).get()
+            command = try RemoteCommand.createRemoteCommand(notification: notification, allowedPresets: loopManager.settings.overridePresets, defaultAbsorptionTime: carbStore.defaultAbsorptionTimes.medium).get()
         } catch {
             log.error("Remote Notification Error: %{public}@", String(describing: error))
             return
--- a/Loop/Models/RemoteCommand.swift
+++ b/Loop/Models/RemoteCommand.swift
@@ -66,7 +66,7 @@ extension RemoteCommand {
         } else if let carbsValue = notification["carbs-entry"] as? Double {
 
             let minAbsorptionTime = TimeInterval(hours: 0.5)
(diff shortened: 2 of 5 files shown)


Loop/Managers/DeviceDataManager.swift
```diff
@@ -1337,7 +1337,7 @@ extension DeviceDataManager {
         let command: RemoteCommand
         
         do {
```
gestrich, line 1339: Pass the defaultAbsorptionTime when no absorption given in remote command, rather than just using 3.0.

LoopTests/Models/RemoteCommandTests.swift
```diff
@@ -37,7 +37,7 @@ class RemoteCommandTests: XCTestCase {
         ]
```
gestrich, line 37: Added new tests for the new cases. I also needed to adjust a few unrelated due to the new `defaultAbsorptionTime` parameter.

Loop/Models/RemoteCommand.swift
```diff
@@ -64,11 +64,19 @@ extension RemoteCommand {
         }  else if let bolusValue = notification["bolus-entry"] as? Double {
             return .success(.bolusEntry(bolusValue))
         } else if let carbsValue = notification["carbs-entry"] as? Double {
-            // TODO: get default absorption value
-            var absorptionTime = TimeInterval(hours: 3.0)
-            if let absorptionOverride = notification["absorption-time"] as? Double {
-                absorptionTime = TimeInterval(hours: absorptionOverride)
+
+            let minAbsorptionTime = TimeInterval(hours: 0.5)
+            let maxAbsorptionTime = TimeInterval(hours: 8)
```
ps2, line 69: This constant is repeated in CarbEntryViewController; please extract into LoopConstants

ps2 (changes requested):
Looks good! Just a small request.

Loop/Models/RemoteCommand.swift
gestrich: Good idea - Done.

PR #1856, opened 2022-12-02
Improve remote notification logging

gestrich:
In an effort to better track remote bolus/carb failures, this will add additional logging. Also, I there were a few places that the `info` log type was used which apparently wasn't making its way to remote logging services -- switching to `error` type where appropriate.

### DIFF
--- a/Loop/Managers/DeviceDataManager.swift
+++ b/Loop/Managers/DeviceDataManager.swift
@@ -1313,10 +1313,11 @@ extension DeviceDataManager {
     func handleRemoteNotification(_ notification: [String: AnyObject]) {
 
         defer {
-            log.info("Finished handling remote notification")
+            log.default("Remote Notification: Finished handling")
         }
         
         guard FeatureFlags.remoteOverridesEnabled else {
+            log.error("Remote Notification: Overrides not enabled.")
             return
         }
 
@@ -1325,11 +1326,11 @@ extension DeviceDataManager {
             formatter.formatOptions =  [.withInternetDateTime, .withFractionalSeconds]
             if let expiration = formatter.date(from: expirationStr) {
                 guard expiration > Date() else {
-                    log.error("Expired notification: %{public}@", String(describing: notification))
+                    log.error("Remote Notification: Expired: %{public}@", String(describing: notification))
                     return
                 }
             } else {
-                log.error("Invalid expiration: %{public}@", expirationStr)
+                log.error("Remote Notification: Invalid expiration: %{public}@", expirationStr)
                 return
             }
         }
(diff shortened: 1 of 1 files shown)


Loop/Managers/DeviceDataManager.swift
```diff
@@ -1313,10 +1313,11 @@ extension DeviceDataManager {
     func handleRemoteNotification(_ notification: [String: AnyObject]) {
 
         defer {
```
gestrich, line 1315: Prefixing remote logs in this method to make them easy to search in logs.

Someday I'd like to refactor this method to throw an error (or return a Result<>) so that logging can happen in a second statement.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1339,69 +1340,70 @@ extension DeviceDataManager {
...
-            log.default("Canceling temporary override from remote command")
+            log.default("Remote Notification: Canceling temporary override")
             loopManager.mutateSettings { settings in settings.scheduleOverride = nil }
         case .bolusEntry(let bolusAmount):
-            log.default("Enacting remote bolus entry: %{public}@", String(describing: bolusAmount))
+            log.default("Remote Notification: Enacting bolus entry: %{public}@", String(describing: bolusAmount))
             
             //Remote bolus requires validation from its remote source
             guard remoteDataServicesManager.validatePushNotificationSource(notification) else {
                 NotificationManager.sendRemoteBolusFailureNotification(for: RemoteCommandError.invalidOTP, amount: bolusAmount)
```
gestrich, line 1360: This call above to `validatePushNotificationSource` would really benefit from throwing/returning an error rather than a bool result. That would let us report if the OTP was invalid vs the user doesn't have OTP setup on the Looper's device. Something to look at in a refactor.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1339,69 +1340,70 @@ extension DeviceDataManager {
...
             }
         case .carbsEntry(let candidateCarbEntry):
-            log.default("Adding carbs entry.")
+            log.default("Remote Notification: Adding carbs entry: %{public}@", String(describing: candidateCarbEntry))
             
             let candidateCarbsInGrams = candidateCarbEntry.quantity.doubleValue(for: .gram())
             
             //Remote carb entry requires validation from its remote source
             guard remoteDataServicesManager.validatePushNotificationSource(notification) else {
                 NotificationManager.sendRemoteCarbEntryFailureNotification(for: RemoteCommandError.invalidOTP, amountInGrams: candidateCarbsInGrams)
```
gestrich, line 1393: I really want to see how old the expired OTP codes are here. Add a log when this failure happens should let me infer in from user's logs. 

If OTPs are consistently 2-3 minutes old during a failure, we could consider extending the age of OTPs we accept to account for that. I think that would be a good interim step to making our remote capabilities more reliable until a command queue solution is developed.

PR #1857, opened 2022-12-03
Ensure remote treatments are uploaded to NS while in background

gestrich:
Nightscout remote treatments (bolus, carbs, overrides) are sometimes delayed in showing up in Nightscout. This seems related to the app being in the background.

The primary reason for the delays seems to be the various delivery/upload threads halting their progress while in the background -- iOS just stops giving CPU time. Opening the app will "wake" the threads back up and finish the upload process.

A mechanism for synchronously waiting on the GCD groupTask did not seem to working as intended from what could tell. It seems the reason was there was no groupTasks to wait on when the `wait` was invoked as the code that enacts treatments is highly asynchronous. There were often several dispatches to different queues to wait on before the groupTasks were added.

The change in this PR is to use backgroundTasks rather than waits. We _could_  instead use a wait with this same pattern I'm using for backgroundTasks... The reason I went with a background task is the wait is causing Xcode to show thread priority inversion warnings since several of the queues we would need to wait on are `.background` priority.

### DIFF
--- a/Loop/Managers/DeviceDataManager.swift
+++ b/Loop/Managers/DeviceDataManager.swift
@@ -1432,30 +1432,54 @@ extension DeviceDataManager {
     }
     
     func addRemoteCarbEntry(_ carbEntry: NewCarbEntry, completion: @escaping (_ result: CarbStoreResult<StoredCarbEntry>) -> Void) {
-        let backgroundTaskID = "Add Remote Carb Entry"
-        let backgroundTask = UIApplication.shared.beginBackgroundTask(withName: backgroundTaskID)
+
+        var backgroundTask: UIBackgroundTaskIdentifier?
+        backgroundTask = UIApplication.shared.beginBackgroundTask(withName: "Add Remote Carb Entry") {
+            guard let backgroundTask = backgroundTask else {return}
+            UIApplication.shared.endBackgroundTask(backgroundTask)
+            self.log.error("Add Remote Carb Entry background task expired")
+        }
+
         carbStore.addCarbEntry(carbEntry) { result in
             self.triggerBackgroundUpload(for: .carb)
-            UIApplication.shared.endBackgroundTask(backgroundTask)
+            if let backgroundTask = backgroundTask {
+                UIApplication.shared.endBackgroundTask(backgroundTask)
+            }
             completion(result)
         }
     }
     
     func enactRemoteBolus(units: Double, completion: @escaping (_ error: Error?) -> Void = { _ in }) {
-        let backgroundTaskID = "Add Remote Bolus Entry"
-        let backgroundTask = UIApplication.shared.beginBackgroundTask(withName: backgroundTaskID)
+        
+        var backgroundTask: UIBackgroundTaskIdentifier?
+        backgroundTask = UIApplication.shared.beginBackgroundTask(withName: "Enact Remote Bolus") {
+            guard let backgroundTask = backgroundTask else {return}
+            UIApplication.shared.endBackgroundTask(backgroundTask)
+            self.log.error("Enact Remote Bolus background task expired")
+        }
+        
         self.enactBolus(units: units, activationType: .manualNoRecommendation) { error in
             self.triggerBackgroundUpload(for: .dose)
-            UIApplication.shared.endBackgroundTask(backgroundTask)
+            if let backgroundTask = backgroundTask {
+                UIApplication.shared.endBackgroundTask(backgroundTask)
+            }
             completion(error)
         }
     }
     
     func triggerBackgroundUpload(for triggeringType: RemoteDataType) {
-        let backgroundTaskID = "Remote Data upload"
-        let backgroundTask = UIApplication.shared.beginBackgroundTask(withName: backgroundTaskID)
-        self.remoteDataServicesManager.triggerUpload(for: triggeringType) {
+        
+        var backgroundTask: UIBackgroundTaskIdentifier?
+        backgroundTask = UIApplication.shared.beginBackgroundTask(withName: "Remote Data Upload") {
+            guard let backgroundTask = backgroundTask else {return}
             UIApplication.shared.endBackgroundTask(backgroundTask)
+            self.log.error("Remote Data Upload background task expired")
+        }
+        
+        self.remoteDataServicesManager.triggerUpload(for: triggeringType) {
+            if let backgroundTask = backgroundTask {
+                UIApplication.shared.endBackgroundTask(backgroundTask)
+            }
         }
     }
 }
--- a/Loop/Managers/DeviceDataManager.swift
+++ b/Loop/Managers/DeviceDataManager.swift
@@ -763,27 +763,28 @@ extension DeviceDataManager {
             return
         }
 
-        self.loopManager.addRequestedBolus(DoseEntry(type: .bolus, startDate: Date(), value: units, unit: .units, isMutable: true), completion: nil)
-        pumpManager.enactBolus(units: units, activationType: activationType) { (error) in
-            if let error = error {
-                self.log.error("%{public}@", String(describing: error))
-                switch error {
-                case .uncertainDelivery:
-                    // Do not generate notification on uncertain delivery error
-                    break
-                default:
-                    // Do not generate notifications for automatic boluses that fail.
-                    if !activationType.isAutomatic {
-                        NotificationManager.sendBolusFailureNotification(for: error, units: units, at: Date(), activationType: activationType)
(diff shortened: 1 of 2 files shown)


Loop/Managers/DeviceDataManager.swift
```diff
@@ -763,27 +763,28 @@ extension DeviceDataManager {
             return
         }
```
gestrich, line 765: This method `enactBolus` was asynchronously calling `self.loopManager.addRequestedBolus` then enacting the bolus. This indeterminate order of operations seemed to often cause this method to call the completion block before the `addRequestedBolus` which means callers can't be sure things are done upon completion. This was an extraneous issue that was leading to experiments with waits and backgroundTasks not to work.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1352,12 +1353,12 @@ extension DeviceDataManager {
 
         switch command {
```
gestrich, line 1355: For overrides, bolus, carbs -- I moved the logic into dedicated methods given the expanded things that happen.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1421,8 +1421,39 @@ extension DeviceDataManager {
                 }
             }
         }
```
gestrich, line 1423: I removed this and am instead using the backgroundTasks.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1421,8 +1421,39 @@ extension DeviceDataManager {
...
-        // Wait up to 25 seconds for uploads triggered by these commands to finish
-        let _ = remoteDataServicesManager.waitForUploadsToFinish(timeout: .now() + TimeInterval(25))
+    }
+
```
gestrich, line 1425: The pattern for these 3 methods is similar:

1. Perform the treatment action.
2. Call `triggerBackgroundUpload` to force the upload to happen now so we have something to wait on. 

Without triggering the upload explicitly, we can't rely on the groupTasks having been started yet by this time.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1421,8 +1421,39 @@ extension DeviceDataManager {
...
+    func enactRemoteBolus(units: Double, completion: @escaping (_ error: Error?) -> Void = { _ in }) {
+        let backgroundTaskID = "Add Remote Bolus Entry"
+        let backgroundTask = UIApplication.shared.beginBackgroundTask(withName: backgroundTaskID)
+        self.enactBolus(units: units, activationType: .manualNoRecommendation) { error in
+            self.triggerBackgroundUpload(for: .dose)
+            UIApplication.shared.endBackgroundTask(backgroundTask)
+            completion(error)
+        }
+    }
+
```
gestrich, line 1450: A helper method to trigger an upload and attach a background task to it's completion.

Loop/Managers/RemoteDataServicesManager.swift
```diff
@@ -176,10 +176,6 @@ final class RemoteDataServicesManager {
         clearSettingsQueryAnchor(for: remoteDataService)
     }
```
gestrich, line 178: Remove this in favor of `triggerBackgroundUpload` above and `triggerUpload` which offers a callback on completion.

There is a chance the uploadGroup will include other things unrelated to what we triggered, but I think that is ok as we had this issue before with `waitForUploadsToFinish`

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1424,8 +1424,39 @@ extension DeviceDataManager {
...
         }
-        // Wait up to 25 seconds for uploads triggered by these commands to finish
-        let _ = remoteDataServicesManager.waitForUploadsToFinish(timeout: .now() + TimeInterval(25))
+    }
+    
+    func activateRemoteOverride(_ remoteOverride: TemporaryScheduleOverride?) {
+        loopManager.mutateSettings { settings in settings.scheduleOverride = remoteOverride }
+        self.triggerBackgroundUpload(for: .overrides)
+    }
+
```
gestrich, line 1433: For remote bolus and remote carbs -- there is an additional background task we do. We need to wait for the carb entry or bolus entry to complete first before we wait for the upload action. Without this, the app was sometimes hanging earlier in the process.

Loop/Managers/DeviceDataManager.swift
gestrich: I was worried these background task IDs needed to be unique per invocation. But from what I can tell they are labels for the debugger.

Loop/Managers/RemoteDataServicesManager.swift
ps2: My one concern about the background task strategy, rather than using the time that iOS has allocated us for processing the remote command (`didReceiveRemoteNotification`), is that we're depleting the limited resource of *background task* time.  We currently don't use background tasks for much else, so there's not currently a lot of contention. But it is a limited resource, and if we could take advantage of the time allocated for `didReceiveRemoteNotification`, that would be more aligned with the expectations of iOS.

ps2 (changes requested):
Generally looks good. There is one request to add logging, and another comment about the general strategy.

Loop/Managers/RemoteDataServicesManager.swift
gestrich: Just to confirm before I take this direction - I can easily switch to using the method that was used previously here for the remote uploads `waitForUploadsToFinish`.

To eliminate my additional background tasks for `self.enactBolus` `carbStore.addCarbEntry`, I assume you'd like a similar strategy followed -- using dispatch groups and waiting on their result rather than a background task. That will add some new Xcode thread inversion warnings, just FYI.

I can go  that direction as there are tradeoffs either way but I just want to clarify that's a change you'd like to see.

ps2:
Sorry for not being clear. I wouldn’t block this ticket on using the background tasks. If, after hearing my concerns you don’t agree or think it’d be a lot of work (the thread inversion would be something to consider), then I’m fine punting on it .Sent from my iPhoneOn Dec 4, 2022, at 1:35 PM, Bill ***@***.***> wrote:﻿
@gestrich commented on this pull request.



In Loop/Managers/RemoteDataServicesManager.swift:
> @@ -176,10 +176,6 @@ final class RemoteDataServicesManager {
         clearSettingsQueryAnchor(for: remoteDataService)
     }
 

Just to confirm before I take this direction - I can easily switch to using the method that was used previously here for the remote uploads waitForUploadsToFinish.
To eliminate my additional background tasks for self.enactBolus carbStore.addCarbEntry, I assume you'd like a similar strategy followed -- using dispatch groups and waiting on their result rather than a background task. That will add some new Xcode thread inversion warnings, just FYI.
I can go  that direction as there are tradeoffs either way but I just want to clarify that's a change you'd like to see.

gestrich:
@ps2  Thanks for the clarification. It is a tough call and I can see the tradeoffs - given the choice I'd lead towards the background task to avoid the thread inversion issues but I can see we will need to watch for background execution issues. Thanks for pointing that out. 

I updated the PR to log the background expirations and made sure to cancel the task inside the expiration. Just let me know if anything needed.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1424,8 +1424,63 @@ extension DeviceDataManager {
...
-        let _ = remoteDataServicesManager.waitForUploadsToFinish(timeout: .now() + TimeInterval(25))
+    }
+    
+    func activateRemoteOverride(_ remoteOverride: TemporaryScheduleOverride?) {
+        loopManager.mutateSettings { settings in settings.scheduleOverride = remoteOverride }
+        self.triggerBackgroundUpload(for: .overrides)
+    }
+    
+    func addRemoteCarbEntry(_ carbEntry: NewCarbEntry, completion: @escaping (_ result: CarbStoreResult<StoredCarbEntry>) -> Void) {
+
```
gestrich, line 1435: It may look weird at first to see local optional variable for the `UIBackgroundTaskIdentifier`. That was due to the block capturing the local variable. Rob Napier has a good [SO](https://stackoverflow.com/questions/28993628/how-to-pass-background-task-identifier-to-completion-handler) answer on someone who asked about this.

bill-foreflight:
@ps2 Thanks for checking. I'm actually going to do just a few days of testing if that works and will let you know when it looks good.

gestrich:
@ps2 We've had no issues with this branch and I think it is ready to merge now 👍

PR #1867, opened 2022-12-10
Improve Remote failure error messaging

gestrich:
We have had some reports of failed remote deliveries and the current error messaging isn't very helpful for understanding why a remote push failed. This information is captured in diagnostic logs but ideally this would be surfaced to a user. This PR enhances our user-facing errors messages.

Associated PRs:

LoopKit: https://github.com/LoopKit/Loop/pull/1867
NightscoutService: https://github.com/ps2/NightscoutService/pull/24

### DIFF
--- a/Loop/Managers/DeviceDataManager.swift
+++ b/Loop/Managers/DeviceDataManager.swift
@@ -1329,12 +1329,18 @@ extension DeviceDataManager {
             log.error("Remote Notification: Overrides not enabled.")
             return
         }
+        
+        let defaultServiceIdentifier = "NightscoutService"
+        let serviceIdentifer = notification["serviceIdentifier"] as? String ?? defaultServiceIdentifier
 
         if let expirationStr = notification["expiration"] as? String {
             let formatter = ISO8601DateFormatter()
             formatter.formatOptions =  [.withInternetDateTime, .withFractionalSeconds]
             if let expiration = formatter.date(from: expirationStr) {
-                guard expiration > Date() else {
+                let nowDate = Date()
+                guard nowDate < expiration else {
+                    let expiredInterval = nowDate.timeIntervalSince(expiration)
+                    NotificationManager.sendRemoteCommandExpiredNotification(timeExpired: expiredInterval)
                     log.error("Remote Notification: Expired: %{public}@", String(describing: notification))
                     return
                 }
@@ -1365,8 +1371,14 @@ extension DeviceDataManager {
             log.default("Remote Notification: Enacting bolus entry: %{public}@", String(describing: bolusAmount))
             
             //Remote bolus requires validation from its remote source
-            guard remoteDataServicesManager.validatePushNotificationSource(notification) else {
-                NotificationManager.sendRemoteBolusFailureNotification(for: RemoteCommandError.invalidOTP, amount: bolusAmount)
+            
+            let validationResult = remoteDataServicesManager.validatePushNotificationSource(notification, serviceIdentifier: serviceIdentifer)
+
+            switch validationResult {
+            case .success():
+                log.info("Remote Notification: Validation successful")
+            case .failure(let error):
+                NotificationManager.sendRemoteBolusFailureNotification(for: error, amount: bolusAmount)
                 log.error("Remote Notification: Could not validate notification: %{public}@", String(describing: notification))
                 return
             }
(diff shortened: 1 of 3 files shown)


Loop/Managers/DeviceDataManager.swift
```diff
@@ -1329,12 +1329,18 @@ extension DeviceDataManager {
             log.error("Remote Notification: Overrides not enabled.")
             return
         }
```
gestrich, line 1331: This is the start of routing a remote notification validation to the service that generated it (i.e. Nightscout being a service). Through now, we've  been allowing any service to validate remote notifications (i.e. OTP codes), even those that have nothing to do with it which could lead to weird issues. I think they should only be routed to the service that generated them. This is also useful for my work in this PR where I need to get the errors associated with remote validation so I only want to route them to the appropriate service.

Since we are not yet injecting the notification on the server-side (Nightscout), I'm making "NightscoutService" the fallback here.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1329,12 +1329,18 @@ extension DeviceDataManager {
...
+        
+        let defaultServiceIdentifier = "NightscoutService"
+        let serviceIdentifer = notification["serviceIdentifier"] as? String ?? defaultServiceIdentifier
 
         if let expirationStr = notification["expiration"] as? String {
             let formatter = ISO8601DateFormatter()
             formatter.formatOptions =  [.withInternetDateTime, .withFractionalSeconds]
             if let expiration = formatter.date(from: expirationStr) {
```
gestrich, line 1339: Show an error when the notification has expired.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1365,8 +1371,14 @@ extension DeviceDataManager {
             log.default("Remote Notification: Enacting bolus entry: %{public}@", String(describing: bolusAmount))
             
             //Remote bolus requires validation from its remote source
```
gestrich, line 1373: Capture the error from the validation service to give the user more information on why something failed and how to fix it.

Loop/Managers/NotificationManager.swift
```diff
@@ -130,6 +153,9 @@ extension NotificationManager {
 
         notification.body = body
         notification.sound = .default
```
gestrich, line 155: Some had these interruptionLevel and others did not - making consistent.

Loop/Managers/RemoteDataServicesManager.swift
```diff
@@ -586,15 +586,13 @@ extension RemoteDataServicesManager {
 
 extension RemoteDataServicesManager {
```
gestrich, line 588: As mentioned above, we will validate a push notification by the service that generated it.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1329,12 +1329,18 @@ extension DeviceDataManager {
             log.error("Remote Notification: Overrides not enabled.")
             return
         }
+        
+        let defaultServiceIdentifier = "NightscoutService"
+        let serviceIdentifer = notification["serviceIdentifier"] as? String ?? defaultServiceIdentifier
```
gestrich, line 1335: A side point as I was working through this handlePushNotification code...

I'm thinking an evolution of this code, someday, may be to send a raw notification to each service for handling. They could handle all the validation, like this expiration check below, which may vary by service.

There are some remote actions maybe only they can handle like activating a service setting or mode.

On that matter, maybe they could handle all notifications (i.e. deliver the bolus/carbs/etc). We could allow services to perform such via service delegation and add controls on the Loop side to restrict things for safety. 

Anyways, just a few thoughts.

Loop/Managers/DeviceDataManager.swift
ps2: Yes, agree 100%. We should only try to validate commands against the service that issued them. Right now, a service integration in Loop sends a device id to the remote service, so that it can route the commands to this device correctly. I think the service integration should also send a service id, so that Loop can know which service to use to validate the command.

Loop/Managers/DeviceDataManager.swift
ps2: I think validating the command is the service's responsibility. Not sure executing the command is, though. I think the "command set" is a Loop domain. Maybe we're talking about the same thing, though.

Loop/Managers/NotificationManager.swift
ps2: I'm not sure a failed remote command is of any urgency to the loop user; it's important to the caregiver.

gestrich: Good point - I've removed.

ps2 (approved):
This is polishing a dead-end use case, I hope. I'm counting on this being short term. It's not just the code that gets harder to change with time, but user behavior/expectations.

Issue #1863, opened 2022-12-07
Feature Request: Add 2 safety features for maxCOB and maxIOB


marionbarker:
**Is your feature request related to a problem? Please describe.**
1. There have been several instances where a bad CGM reading (or reading in the future) has caused Loop to deliver maxBolus (or 40% of that value) at 5 minute interval, leading to excessive insulin delivery.
1. There have been several instances where user accidentally entered a meal more than once which has caused Loop to deliver excessive insulin.

**Describe the solution you'd like**
Add a user adjustable fail-safe value to Loop.

One idea:

maxIOB:
* Automatic Delivery of insulin checks current IOB and if the recommended automatic bolus would exceed maxIOB, limit automatic bolus to not exceed maxIOB and issue a notification to user.
   * Manual Delivery would not be limited by maxIOB (or perhaps there could be 2 settings, one for automatic and one for manual)
   
maxCOB:
* When accepting a Carb entry or edit, check to see if the COB with that carb entry would exceed the user selected maxCOB and refuse to accept the entry
   * If user wants to proceed with the Carb entry, they could modify setting or loop might offer an accept button for one-time use without modifying maxCOB selection.

BiagioL:
The introduction of  these  two safety features are of great importance and will help avoid bad situations just like the one that  happen to me last month when without realizing it I entered 115 grs. of Carbs instead of the just 15 gr. intended and had my worse Hypoglycemia in 45 years as T1 and 3:1/2 years use of Loop.  Thank you for addressing it.

marionbarker:
With the addition of the Large Meal warning screen to the carb entry screen when entry exceeds 99 g, modify this feature request to limit the automatic dosing.
See PR #1871 and LoopKit PR #442

ps2:
Closing this; for now; the IOB mitigation is under development, and we should have a separate ticket (or tickets) to discuss any carb related mitigations that are still needed.

Issue #1862, opened 2022-12-07
Feature Request: Add Setting for maxCarbEntry


marionbarker:
**Is your feature request related to a problem? Please describe.**
It is possible to make an error typing in a carb entry in grams. For individuals who use small amounts of carbs, it would be helpful for them to select a maxCarbEntry in Settings.

**Describe the solution you'd like**
Add an option to configure the maximum carb entry for a single event.
For example, the user could select 99 g to avoid an accidental entry of 115 instead of 15.

**Describe alternatives you've considered**
1. A code customization is already provided in LoopDocs to modify the value.
1. This could be added as switch in iOS Settings->Loop with the LoopPatches add-on, but it is far better to incorporate into Loop.

marionbarker:
Two PR have been submitted which add a warning limit to the existing hard limit for carb quantity.
* New warning limit is 99 g.
* Hard (max) limit remains 250 g.

Values for warning and max quantity are set in LoopConstants.

[Loop PR 1873](https://github.com/LoopKit/Loop/pull/1873)

[LoopKit PR 438](https://github.com/LoopKit/LoopKit/pull/438)

Design:
* User enters any value up to 99 g, value is accepted
* User enters value between 100 g and 250 g, user is shown a modal screen that warns value is above 99 g
   * User then choses to continue with that value or can edit quantity
   * See first image for the warning message
* User enters value > 250 g, they cannot continue with that value
    * See second image for updated message

![IMG_0291](https://user-images.githubusercontent.com/19607791/208248514-178a24e0-c295-4b60-be5b-1dcf18844715.PNG)

![IMG_0290](https://user-images.githubusercontent.com/19607791/208248648-51b366ce-b25f-417d-a3d8-033e78612a56.PNG)

marionbarker:
Updated graphics when [LoopKit PR 468](https://github.com/LoopKit/LoopKit/pull/438) was updated with new messages.

marionbarker:
I'm closing this. We have added safety warnings with both large meals and when overrides are set.

PR #1870, opened 2022-12-14
DRAFT: add safety features: maxCarbEntry, maxAutoIOB

marionbarker:
This draft PR addresses
* maxCarbEntry Feature Request of #1862 
* maxAutoIOB portion of the Feature Request of #1863.

Safety Features added with this draft design:
* Prevents a single entry of carbs > maxCarbEntry amount
* Prevents automatic increase of insulin delivery above a user selected maxAutoIOB value

Rather than focusing on how to add new user adjustable settings to Loop and then implementing the functional code that uses them, I took the short-cut of adding them to a Settings Bundle for this Draft PR.

These two settings can be adjusted under iOS->Settings->Loop:
* maxCarbEntry
* maxAutoIOB

This code has been tested and does what I expect. 

For example:

1. If user selects maxCarbEntry of 99g, Loop provides the modal message that it cannot accept larger values when attempting to add or edit carbs of 100g or larger.
2. If user selects maxAutoIOB of 4.0 U, Loop:
    * Does not automatically increase delivery of insulin  above 4.0 U, but
    * Allows user to manually enter insulin delivery resulting in an IOB > maxAutoIOB, and
    * Recommended bolus shown to user is not limited by maxAutoIOB

I wasn't sure how to implement logging in DoseMath, so I added some print statements.

Example debug output when a correction would cause IOB to exceed maxAutoIOB:
```
*** maxAutoIOB Feature enacted in aboveRange
***  Initial minCorrectionUnits  3.73
***  maxAutoIOB, insulinOnBoard, checkMaxAutoIOB  4.0 1.68 -1.41
***  Adjusted minCorrectionUnits  2.32
```

Code Modifications
* CarbEntryViewController was modified to read maxCarbEntry from settings and apply it to carb entry/edit action
* LoopDataManager was modified to read the maxAutoIOB from settings and to pass new optional arguments to functions in DoseMath
* DoseMath private func insulinCorrection was modified to use new optional arguments
    * Check for maxAutoIOB was added to the aboveRange section
    * Tested for both automatic bolus and temp basal only dosing strategy
* The other code modifications, i.e., Loop.xcodeproj/project.pbxproj and Settings.bundle folder, will be discarded - these enable the shortcut of targeting the new settings inside the functions that need them


ps2 (changes requested):
These settings should be introduced to the user during onboarding, and available for the user to change in therapy settings along with the other limits.

marionbarker:
I may need help in doing this, but will give it a try.

ps2:
Discussed on zulip. I think this should be split into two different parts. The carb input warning level could be a new configuration item. Max IOB is very strongly related to Max Bolus & Max TB, and it might be possible to simplify there.

PR #1871, opened 2022-12-15
Use maxBolus to set automaticDosingIOBLimit

marionbarker:
This replaces PR #1870.
This PR adds one safety check so that automatic dosing of insulin cannot exceed a simple ratio multiplied by a constant set in LoopConstants. It addresses the more critical portion of Issue #1863.

If this is deemed a good direction, there are some items to do:
* add proper logging
* report in the Delivery Limits section that the maximum of automatic delivery of insulin will be capped by the ratio times maxBolus to make it clear to the user
* decide what the correct ratio should be (1.2 was entered as a placeholder) - subsequently changed to 2.0.

### DIFF
--- a/DoseMathTests/DoseMathTests.swift
+++ b/DoseMathTests/DoseMathTests.swift
@@ -55,16 +55,6 @@ class RecommendTempBasalTests: XCTestCase {
 
     fileprivate let maxBasalRate = 3.0
 
-    fileprivate let maxBolus = 12.5
-
-    var automaticDosingIOBLimit: Double {
-        return 2.0 * maxBolus
-    }
-
-    var maxAutomaticBolus: Double {
-        return maxBolus * 0.4
-    }
-
     fileprivate let fortyIncrementsPerUnitRounder = { round($0 * 40) / 40 }
 
     func loadGlucoseValueFixture(_ resourceName: String) -> [GlucoseFixtureValue] {
--- a/DoseMathTests/DoseMathTests.swift
+++ b/DoseMathTests/DoseMathTests.swift
@@ -55,7 +55,15 @@ class RecommendTempBasalTests: XCTestCase {
 
     fileprivate let maxBasalRate = 3.0
 
-    fileprivate let maxBolus = 5.0
+    fileprivate let maxBolus = 12.5
+
+    var automaticDosingIOBLimit: Double {
+        return 2.0 * maxBolus
+    }
+
+    var maxAutomaticBolus: Double {
+        return maxBolus * 0.4
+    }
 
     fileprivate let fortyIncrementsPerUnitRounder = { round($0 * 40) / 40 }
 
@@ -107,13 +115,8 @@ class RecommendTempBasalTests: XCTestCase {
         return TimeInterval(hours: 4)
     }
 
-    var automaticDosingIOBLimit: Double {
-        return 2.0 * maxBolus
-    }
-
     func testNoChange() {
         let glucose = loadGlucoseValueFixture("recommend_temp_basal_no_change_glucose")
-        let insulinOnBoard = 0.0
 
         let dose = glucose.recommendedTempBasal(
             to: glucoseTargetRange,
@@ -123,9 +126,7 @@ class RecommendTempBasalTests: XCTestCase {
             model: walshInsulinModel,
             basalRates: basalRateSchedule,
             maxBasalRate: maxBasalRate,
-            lastTempBasal: nil,
-            insulinOnBoard: insulinOnBoard,
-            automaticDosingIOBLimit: automaticDosingIOBLimit
+            lastTempBasal: nil
         )
 
         XCTAssertNil(dose)
@@ -133,7 +134,6 @@ class RecommendTempBasalTests: XCTestCase {
     
     func testNoChangeExponentialModel() {
         let glucose = loadGlucoseValueFixture("recommend_temp_basal_no_change_glucose")
-        let insulinOnBoard = 0.0
 
         let dose = glucose.recommendedTempBasal(
             to: glucoseTargetRange,
@@ -143,9 +143,7 @@ class RecommendTempBasalTests: XCTestCase {
             model: exponentialInsulinModel,
             basalRates: basalRateSchedule,
             maxBasalRate: maxBasalRate,
-            lastTempBasal: nil,
-            insulinOnBoard: insulinOnBoard,
-            automaticDosingIOBLimit: automaticDosingIOBLimit
+            lastTempBasal: nil
         )
 
         XCTAssertNil(dose)
@@ -153,7 +151,6 @@ class RecommendTempBasalTests: XCTestCase {
     
     func testNoChangeAutomaticBolusing() {
         let glucose = loadGlucoseValueFixture("recommend_temp_basal_no_change_glucose")
-        let insulinOnBoard = 0.0
 
         let dose = glucose.recommendedAutomaticDose(
             to: glucoseTargetRange,
@@ -164,9 +161,7 @@ class RecommendTempBasalTests: XCTestCase {
             basalRates: basalRateSchedule,
             maxAutomaticBolus: 5,
             partialApplicationFactor: 0.5,
-            lastTempBasal: nil,
-            insulinOnBoard: insulinOnBoard,
-            automaticDosingIOBLimit: automaticDosingIOBLimit
+            lastTempBasal: nil
         )
 
         XCTAssertNil(dose)
@@ -174,7 +169,6 @@ class RecommendTempBasalTests: XCTestCase {
 
     func testNoChangeOverrideActive() {
         let glucose = loadGlucoseValueFixture("recommend_temp_basal_no_change_glucose")
-        let insulinOnBoard = 7.69
 
         let dose = glucose.recommendedTempBasal(
             to: glucoseTargetRange,
@@ -185,9 +179,7 @@ class RecommendTempBasalTests: XCTestCase {
             basalRates: basalRateSchedule,
             maxBasalRate: maxBasalRate,
             lastTempBasal: nil,
-            isBasalRateScheduleOverrideActive: true,
-            insulinOnBoard: insulinOnBoard,
-            automaticDosingIOBLimit: automaticDosingIOBLimit
+            isBasalRateScheduleOverrideActive: true
         )
 
         XCTAssertEqual(0.8, dose!.unitsPerHour, accuracy: 1.0 / 40.0)
@@ -196,7 +188,6 @@ class RecommendTempBasalTests: XCTestCase {
     
     func testNoChangeOverrideActiveAutomaticBolusing() {
         let glucose = loadGlucoseValueFixture("recommend_temp_basal_no_change_glucose")
-        let insulinOnBoard = 7.69
 
         let dose = glucose.recommendedAutomaticDose(
             to: glucoseTargetRange,
@@ -208,9 +199,7 @@ class RecommendTempBasalTests: XCTestCase {
             maxAutomaticBolus: 5,
             partialApplicationFactor: 0.5,
             lastTempBasal: nil,
-            isBasalRateScheduleOverrideActive: true,
-            insulinOnBoard: insulinOnBoard,
-            automaticDosingIOBLimit: automaticDosingIOBLimit
+            isBasalRateScheduleOverrideActive: true
         )
 
         XCTAssertEqual(0.8, dose!.basalAdjustment!.unitsPerHour, accuracy: 1.0 / 40.0)
@@ -220,7 +209,6 @@ class RecommendTempBasalTests: XCTestCase {
 
     func testStartHighEndInRange() {
         let glucose = loadGlucoseValueFixture("recommend_temp_basal_start_high_end_in_range")
-        let insulinOnBoard = 7.69
 
         var dose = glucose.recommendedTempBasal(
             to: glucoseTargetRange,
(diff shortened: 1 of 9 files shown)


marionbarker:
I think we might want to revisit this concept and add another Setting for user to adjust. 

The reason is there were several people who commented on zulipchat in the last 24 hours that to minimize problems with large boluses, they set their maxBolus small, e.g., 5 U, and let autobolus take them up to the 20 U that their meal needs.

marionbarker:
Increase ratioMaxAutoInsulinOnBoardToMaxBolus to 2.0
Theory of operation:
* maxBolus is used to limit single doses, but max auto IOB can be higher
* with Loop in automatic mode (using either dosing strategy), subsequent loops will increase delivery up to ratioMaxAutoInsulinOnBoardToMaxBolus times maxBolus

For users who limit single dosing with smaller maxBolus but need more than 2 times maxBolus as active insulin on board, there may need to be a modification of their current method with this safety feature added.

For example, user who needs 20 U IOB but also needs to limit any given bolus to 5 units to avoid problems with their site:
* Set maxBolus to 10 (or up to 12.5 U)
* For meal entry, manually limit delivery to 5 U (or set no delivery and rely on autobolus)
* Each subsequent autobolus will be no more than 5 U and IOB can reach 20 U (or 25 U).

marionbarker:
Update some variable names.  

See also [LoopKit PR 442: Update max bolus information screen](https://github.com/LoopKit/LoopKit/pull/442)

ps2 (changes requested):
Looks like I didn't submit this review (it was marked pending).  I think this is mostly good, but could be cleaned up a bit (see comments), and this kind of change does deserve unit tests.

Loop/Managers/DoseMath.swift
```diff
@@ -328,8 +330,20 @@ extension Collection where Element: GlucoseValue {
                 units: units
             )
         } else if eventual.quantity > eventualGlucoseTargets.upperBound,
-            let minCorrectionUnits = minCorrectionUnits, let correctingGlucose = correctingGlucose
+            var minCorrectionUnits = minCorrectionUnits, let correctingGlucose = correctingGlucose
         {
+            // Limit automatic dosing to prevent insulinOnBoard > automaticDosingIOBLimit
+            if minCorrectionUnits > 0 &&
+                insulinOnBoard != nil &&
+                automaticDosingIOBLimit != nil &&
```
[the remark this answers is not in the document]
marionbarker: I updated the large conditional.

I need help figuring out how to add a unit test.

DoseMathTests/DoseMathTests.swift
```diff
@@ -1504,3 +1504,232 @@ class RecommendBolusTests: XCTestCase {
         XCTAssertEqual(1.96, dose.amount, accuracy: 1.0 / 40.0)
     }
 }
+
+class automaticIOBLimitTests: XCTestCase {
```
ps2, line 1508: Class names are always capitalized.  But I don't know that this should be a separate class. I see that we do have two classes in this file, with some duplication of methods, but I think these tests should fit into those existing classes.

Loop/Managers/DoseMath.swift
```diff
@@ -234,7 +234,9 @@ extension Collection where Element: GlucoseValue {
         at date: Date,
         suspendThreshold: HKQuantity,
         sensitivity: HKQuantity,
-        model: InsulinModel
+        model: InsulinModel,
+        insulinOnBoard: Double? = nil,
```
ps2, line 238: If Loop will be enforcing a cap on IOB in all cases, these parameters should not be optional, and the tests should be updated to pass values in.

ps2 (changes requested):

Loop/Managers/DoseMath.swift
ps2: This way we can see that existing tests pass, and if the results change, how they change.

DoseMathTests/DoseMathTests.swift
marionbarker: Moved all new tests into class RecommendTempBasalTests

Loop/Managers/DoseMath.swift
marionbarker: I updated the DoseMathTests to pass in the parameters.
Then I reversed the change I made yesterday that made the parameters optional. I hope this is what you meant.

Loop/Managers/LoopDataManager.swift
```diff
@@ -1647,7 +1650,9 @@ extension LoopDataManager {
                     lastTempBasal: lastTempBasal,
                     volumeRounder: volumeRounder,
                     rateRounder: rateRounder,
-                    isBasalRateScheduleOverrideActive: settings.scheduleOverride?.isBasalRateScheduleOverriden(at: startDate) == true
+                    isBasalRateScheduleOverrideActive: settings.scheduleOverride?.isBasalRateScheduleOverriden(at: startDate) == true,
+                    insulinOnBoard: dosingDecision.insulinOnBoard?.value,
```
ps2, line 1654: DosingDecision's responsibility is for recording the state of things during a dosing decision. I think use its value as an input here is mixing up responsibilities too much, and will make it hard to extract dosing decision if we change how that's generated.  The value should probably be passed to this method (preferred, to move more towards a functional implementation of the Loop algorithm, rather than the current stateful one), or the value should be assigned to a member variable on LDM.

Regardless of how this value gets here, failure to retrieve IOB should now be considered an error, as it is part of the Loop dosing algorithm. I think it was previously being treated as a warning.

ps2 (changes requested):
I reviewed this further, and I think handling of IOB needs some updates (see comments).

The changes for DoseMath tests are good, but do not cover LoopDataManager changes. It would be good to see some tests covering the changes in LoopDataManager.

Loop/Managers/LoopDataManager.swift
marionbarker: Preliminary modification made, awaiting review.
Working on LoopDataManagerTests.swift, nothing pushed on this yet.

marionbarker:
Brought the marionbarker wip/max-auto-iob branch in line with dev.
Asked @novalegra for help. She plans to supply this with a PR to my branch.

DoseMathTests/DoseMathTests.swift
```diff
@@ -105,6 +107,16 @@ class RecommendTempBasalTests: XCTestCase {
...
 
+    var automaticDosingIOBLimit: Double {
+        return 2.0 * maxBolus
+    }
+
+    // Tests that were in place before the addition of automaticDosingIOBLimit
+    // should all still succeed so long as
+    //      insulinOnBoard < automaticDosingIOBLimit - (dose returned by test)
+    //  The test with highest dose is a bolus of 2.3 U
+    var insulinOnBoard: Double = 7.69
```
ps2, line 118: This var is being modified locally in tests. It proves a challenge to the reader to know what this is set in a given test as it then depends on the order tests are run. Because actual IOB in real life is going to vary a lot, and we want to exercise a variety of situations, I don't think this should be a file scope variable; each test should just have its own value of IOB that it passes into the methods under test.  For tests that aren't actually testing particular behavior of IOB and IOB limit, passing in 0 is fine.

Loop/Managers/LoopDataManager.swift
```diff
@@ -1647,7 +1649,9 @@ extension LoopDataManager {
                     lastTempBasal: lastTempBasal,
                     volumeRounder: volumeRounder,
                     rateRounder: rateRounder,
-                    isBasalRateScheduleOverrideActive: settings.scheduleOverride?.isBasalRateScheduleOverriden(at: startDate) == true
+                    isBasalRateScheduleOverrideActive: settings.scheduleOverride?.isBasalRateScheduleOverriden(at: startDate) == true,
+                    insulinOnBoard: self.insulinOnBoard?.value,
```
ps2, line 1653: If we have a failure to get insulin on board, then dosing will not be limited. The `insulinOnBoard` function parameter here should probably be non-optional and the loop fail with an error if it can't be retrieved.

ps2 (changes requested):
IOB should be made a requirement for looping, if we're trying to ensure a clamp.  Also,  I recommend removal of the file scope IOB variable in tests, and rather specify what's being done in each test.

DoseMathTests/DoseMathTests.swift
marionbarker: I can certainly move this parameter internal to each of the test.

Loop/Managers/LoopDataManager.swift
marionbarker: I naively thought the code where self.insulinOnBoard is set:

https://github.com/marionbarker/Loop/blob/wip/max-auto-iob/Loop/Managers/LoopDataManager.swift#L1039-L1048

with the switch for result of .failure or .success handles the case where there is a failure to get insulin on board.

DoseMathTests/DoseMathTests.swift
marionbarker: I believe this is resolved with commit [01c775c](https://github.com/LoopKit/Loop/pull/1871/commits/01c775c8c0efa529a7853f9872e48990981289fd)

Loop/Managers/LoopDataManager.swift
marionbarker: Looking more closely (and finding your earlier comment) - it does seem to be an a warning.
Changing this is beyond what I can do.

DoseMathTests/DoseMathTests.swift
```diff
@@ -177,6 +174,7 @@ class RecommendTempBasalTests: XCTestCase {
 
     func testNoChangeOverrideActive() {
         let glucose = loadGlucoseValueFixture("recommend_temp_basal_no_change_glucose")
+        let insulinOnBoard = 7.69
```
ps2, line 177: Why are these and all the below 7.69? Seems an oddly specific number, and I don't think it's obvious why from reading the tests.  If you're testing some IOB based limiting or trying to test that it's not limiting, then I think setting specific values here, makes sense. But otherwise, maybe we just keep them all at 0 when we're not actually trying to trigger some IOB level specific logic.

ps2:
Ok, yes, I see that manual bolusing may still make sense to pass in the maxiob and current iob as nil. And I'm on the fence about pairing them up into a tuple. Will leave that up to you, it does seem maybe a little cleaner to pair them up.

DoseMathTests/DoseMathTests.swift
marionbarker: I am fine returning all prior tests back to 0.0 U IOB - I only changed that because I thought you asked me too. Sorry for the misunderstanding.

The tests where the value of the insulinOnBoard needs to vary are the 7 new tests.

Of all the previously existing tests, the highest dose was 2.3 U, so I chose 7.69 U as IOB.
With that value, the XCTAssert for all the existing tests would pass.
I got lazy pasting in the new let insulinOnBoard into each test and at one point 7.69 was in my paste buffer.
I think I added it starting here because we don't have any clamping tests with OverrideActive.

ps2:
Reviewing this more, I think that [GlucoseValue].insulinCorrection should be kept as is, and not modified.  There are three main users of this method, `recommendedAutomaticDose`, `recommendedTempBasal`, and `recommendedManualBolus`.

`recommendedManualBolus` does not use this limit, so we're passing in the iob limiting parameters as empty passthroughs (discussed above a bit).  The other two do apply IOB based limiting, and because of how they deliver, arguably should be applying the limit differently.  For fast boluses, it's probably ok to ignore time, and just consider it immediate, and limit boluses to maxiob - iob. For temp basals, though, we should probably be targeting the calculated IOB at t+30, as we are calculating the amount to deliver over the next 30 minutes.

I'll take a swing at reworking this, maybe not tackling the future IOB target for temp basals, but at least cleaning up the optionals, simplifying the manual basal path, and keeping `insulinCorrection` intact.

ps2:
Also encountered an error I didn't notice in my initial review: the bolus amount is being limited *before* the partial application factor is applied.  For example, if max bolus is 5, max automatic IOB is 10, IOB is 9.5 and the needed insulin is > 0.5 U, then the `insulinCorrection` in this PR was limiting to 0.5, and *then* applying partial application factor, which results in an automatic bolus of 0.2, rather than 0.5 U.

marionbarker:
I'm glad you are looking at this. Happy to do any testing you want.

ps2:
And a longstanding, but probably not too impactful bug: if max bolus, or the partial application max bolus are not a deliverable unit, Loop is not rounding to deliverable units, so we could be issuing unsupported boluses to pumpmanagers.

novalegra:
@ps2 as a heads up - I don't think my PR against @marionbarker's branch is merged in. I don't think it impacts the partial application factor issue you noted but does clean up the optionals

ps2:
Thanks @novalegra, I do appreciate you taking a swing at the optional issue.  I think moving the limit application into the dose calculations for each kind of dosing (rather than the shared insulinCorrection method)  obviates the need for that approach.  Sorry I didn't realize this earlier and save you that work.

marionbarker:
Regarding https://github.com/LoopKit/Loop/pull/1871#issuecomment-1426433583 - I did not know how to approach this a different way. 

That's why for all the testing I did, I noted that I artificially set partial bolus application factor to 100% to make is easier to evaluate the test results. I did the same thing in the tests added to the testing suite.

ps2 (approved):

alevar1975:
Dear Marion, dear Looper community, I  forwarding you this petition in hope you could spread the petition among the US diabetic community.Maybe, you could also put a banner on the Loop-Docs-website because receiving the medication you need to survive should be a human right.The petition was started by the #dedoc voices.Read more:https://www.change.org/p/urging-uninterrupted-insulin-access-in-humanitarian-crises?utm_content=cl_sharecopy_37882245_es-ES%3A4&recruiter=1218102373&recruited_by_id=18064c60-efe4-11eb-96ad-9d92c45e0609&utm_source=share_petition&utm_medium=copylink&utm_campaign=psf_combo_share_initialKind regards,AlenMessage ID: ***@***.***>

PR #1872, opened 2022-12-16
Updating watch extension logging

ps2:


ps2:
This logging was just work done while trying to understand a crash. Actual fix for crash here: https://github.com/LoopKit/Loop/pull/1874

PR #1873, opened 2022-12-17
Add warning if carb quantity exceeds warning limit

marionbarker:
One of two PR (see also LoopKit) in response to Feature Request #1862


marionbarker:
Moved new warning code to ValidateInput function.
Action is unchanged.

PR #1888, opened 2023-01-04
Adds sorting to AlertView-like dialog for adding a CGM

JamieKeene:
Currently the availableCGMs are unsorted when the user taps the 'Add CGM' pill in UIAlertController, where they're sorted by LocalizedTitle when triggered from settings. 

This PR corrects this difference, making the CGMs easier to parse for new users.

Before:

After:


ps2:
Thanks!

Issue #1875, opened 2022-12-18
Feature Request: Assist users with carb ratio effects during overrides


marionbarker:
**Is your feature request related to a problem? Please describe.**
There have been multiple posts on social media from Loop users who report difficulty understanding the effect on Loop predictions when they add carbs while running an override. The more extreme the override, the bigger the issue.
* The override properly affects basal rates, carb ratio and insulin sensitivity (because the ratio of ISF and CR is critical for predictions)
* However, sometimes people forget or do not understand the consequence of an override when entering or editing a carb entry.

**Describe the solution you'd like**
One solution is to alert users that an override is in effect when a carb entry is entered or edited.

**Describe alternatives you've considered**
There are a lot of posts on Facebook on this topic explaining what happened and why there is an unexpected bolus suggestion or unexpected prediction.
It is covered in documentation, but even experts sometimes forget.

marionbarker:
I am not able to figure out the swift programming required to modify the code to implement a possible change discussed in zulipchat. 

Below is a mock-up display for the concept of warning user upon entry to the carb screen if an override sensitivity is outside indicator limits.

![override-sensitivity-warning](https://user-images.githubusercontent.com/19607791/209588186-ee25c6b5-0df9-44b7-89bd-c0bf3c76cd08.jpeg)

I hacked the displayWarningSection to create the mock-up warning above.
To create this display, I modified 3 lines from the AccurateCarbEntry waning version and hacked the code to force top row to show warning. I changed secondaryLabel to label because the font was really light. It's also tiny.
```
-            cell.imageView?.tintColor = .destructive
+            cell.imageView?.tintColor = .warning
             cell.textLabel?.numberOfLines = 0
-            cell.textLabel?.text = NSLocalizedString("Your glucose is rapidly rising. Check that any carbs you've eaten were logged. If you logged carbs, check that the time you entered lines up with when you started eating.", comment: "Warning to ensure the carb entry is accurate")
+            cell.textLabel?.text = NSLocalizedString("Your override of \(overrideSensitivity)% modifies scheduled carb ratio and insulin senstivity by 25% or more. If this is not your intention, consider turning off the override.", comment: "Warning to ensure the carb entry is accurate")
             cell.textLabel?.font = UIFont.preferredFont(forTextStyle: .caption1)
-            cell.textLabel?.textColor = .secondaryLabel
+            cell.textLabel?.textColor = .label
```


[...]

PR #1892, opened 2023-01-08
Add carb entry override warning

billybooth:
Presents an appropriate warning cell via the CarbEntryViewController when an active override is modifying insulin needs. This safety risk mitigation was discussed in the [#development > Safety Feature Requests](https://loop.zulipchat.com/#narrow/stream/144182-development/topic/Safety.20Feature.20Requests/near/317748578) topic on Zulip. I opted not to display this on the basis of an arbitrary effectiveInsulinNeedsScaleFactor +/-, but instead display the warning if the effectiveInsulinNeedsScaleFactor modifies insulin needs at all.

As @ps2 suggested in the Zulip chat, optional UITableView rows are a bit tricky, so although an effort was made to properly support distinct warning row types (and allow for more to be added), additional refactor would be required to support more than two types of warning rows.
```swift
fileprivate enum WarningRow: Int, CaseIterable {
    case carbEntry
    case override
}
```

Here is a screen grab of the newly introduced override warning row:
IMG_3851

And naturally, support is present for stacking both the override and carb accuracy warning rows:
IMG_3850

Below is a reasonably full demonstration of the behavior added:
https://user-images.githubusercontent.com/20958170/211212450-9adcb837-ed61-4c32-9be5-b79d68496af8.mov

The changes here have been extensively tested and I have a lot of confidence in them, but due to critical importance of this view/view controller and the hazardous nature of dynamically manipulating the UITableView rows, I think this merits a rather thorough review by @ps2.

Let me know about any required changes or feel free to make them yourself if it's more expedient.

### DIFF
--- a/Loop/View Controllers/CarbEntryViewController.swift
+++ b/Loop/View Controllers/CarbEntryViewController.swift
@@ -111,7 +111,31 @@ final class CarbEntryViewController: LoopChartsTableViewController, Identifiable
     private var shouldDisplayAccurateCarbEntryWarning = false {
         didSet {
             if shouldDisplayAccurateCarbEntryWarning != oldValue {
-                self.displayAccuracyWarning()
+                if shouldDisplayOverrideEnabledWarning {
+                    self.displayWarningRow(rowType: WarningRow.carbEntry, isAddingRow: shouldDisplayAccurateCarbEntryWarning)
+                } else {
+                    self.shouldDisplayWarning = shouldDisplayAccurateCarbEntryWarning || shouldDisplayOverrideEnabledWarning
+                }
+            }
+        }
+    }
+    
+    private var shouldDisplayOverrideEnabledWarning = false {
+        didSet {
+            if shouldDisplayOverrideEnabledWarning != oldValue {
+                if shouldDisplayAccurateCarbEntryWarning {
+                    self.displayWarningRow(rowType: WarningRow.override, isAddingRow: shouldDisplayOverrideEnabledWarning)
+                } else {
+                    self.shouldDisplayWarning = shouldDisplayOverrideEnabledWarning || shouldDisplayAccurateCarbEntryWarning
+                }
+            }
+        }
+    }
+    
+    private var shouldDisplayWarning = false {
+        didSet {
+            if shouldDisplayWarning != oldValue {
+                self.displayWarning()
             }
         }
     }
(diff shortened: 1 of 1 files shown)


billybooth:
Failed to notice associated feature request #1875. This PR at least partially mitigates the safety concerns raised there.

@marionbarker [said](https://github.com/LoopKit/Loop/issues/1875#issue-1501980757):
> **Describe the solution you'd like**
> One solution is to alert users that an override is in effect when a carb entry is entered or edited.

ps2 (commented):
Thanks!  I like the wording on the warning. The logic is getting a bit tough to follow, and maybe a dynamic list (of sections and warning rows) approach would be simpler to read, but if you've tested this well, I think the approach is ok. At some point we should convert this to SwiftUI.

One edge case is making sure we don't generate any consistency violations (crashes) if the `shouldDisplayOverrideEnabledWarning` variable changes while on screen.  This could happen if a remote override was enacted while the screen was displayed, for example. If you're not set up to enact remote overrides easily, maybe just adding a timer to change the option after a couple seconds would exercise that part of the code, and you could remove the timer after you're convinced it's working.

marionbarker:
I tested this with remote override commands and reported in zulipchat: https://loop.zulipchat.com/#narrow/stream/144182-development/topic/Safety.20Feature.20Requests/near/320492954

billybooth:
@ps2 [said](https://github.com/LoopKit/Loop/pull/1892#pullrequestreview-1242500685):
> One edge case is making sure we don't generate any consistency violations (crashes) if the shouldDisplayOverrideEnabledWarning variable changes while on screen. This could happen if a remote override was enacted while the screen was displayed, for example. If you're not set up to enact remote overrides easily, maybe just adding a timer to change the option after a couple seconds would exercise that part of the code, and you could remove the timer after you're convinced it's working.

Sorry that the demo video doesn't appear to be rendering in conversation view; it plays back fine if you download it. Yes, as part of testing, I was flipping `shouldDisplayAccurateCarbEntryWarning` every other minute and turning overrides on and off with my watch to flip `shouldDisplayOverrideEnabledWarning` and force loop data updates. I am reasonably confident that no consistency violations can surface from these changes, but I apologize for the code not being super-readable. I tried to add comments where I thought they could be helpful, but it's clear that this would be a great candidate for migration to SwiftUI, since the effort required to accomplish this here is a bit much.

ps2:
Oh, sorry I missed the video! Thanks.

Issue #1890, opened 2023-01-04
Detect Future Glucose


marionbarker:
**Safety Issue Request**

**Is your feature request related to a problem? Please describe.**
When a future value of glucose is entered into Loop, Loop no longer provides a prediction curve but uses that value as the eventual glucose value, reports it on the main Loop screen and continues automatic dosing of insulin. When in closed loop, this can result in an increase or decrease of delivered insulin depending on the glucose value.

* This has happened when the Loop phone time was changed ahead manually (for example when playing a game) and then returned to automatic time later.
* This could happen if someone entered a glucose reading into Apple Health from a finger-stick reading and got the time wrong. 
  * For example, if it was 12:10 pm when entering a value taken at 11:45 am, and inadvertently entered 11:45 pm for the glucose time.

**Describe the solution(s) you've thought of**
* ignore any future glucose values
* make detection of future glucose a critical error - Loop actions will halt until the problem is resolved.

**Test Case**
Configuration: 
* use the version of Loop-dev with PR #1871 applied (this clamps the automatic insulin delivery to no more than 2 times maxBolus)
* select Dosing Strategy of Automatic Bolus
* modify the bolusPartialApplicationFactor to 1.0 (from 0.4) to make AB delivery faster
* test phone Therapy Settings: correction range 100 to 115, maxBolus 5 U, ISF 55, CR 10
* use simulated CGM set to constant 110 mg/dL, start with 0 IOB on test phone
* clamp will limit automatic dosing when IOB >= 10 U

Test High Glucose in the future:
* enter glucose of 250 mg/dL 18 hours in the future
* Loop begins to apply AB dosing up to 10 U IOB clamp

left graphic shows the initial AB doses to reach the IOB clamp
right graphic shows the test phone several hours later

![future-glucose](https://user-images.githubusercontent.com/19607791/210628691-b5e6df5f-35ba-405a-af43-dd4af8b000a3.jpg)

Software Configuration:
LoopWorkspace: dev, commit 84afcfb
Loop: wip/max-auto-iob, commit 97b7d8db

marionbarker:
I'm closing this as included in Loop 3.

PR #1894, opened 2023-01-09
Prevent Automatic Dosing with Future Glucose

marionbarker:
This ~~is a work-in-progress~~ PR is in response to Issue #1890.

Configuration
* Test phone configured with future glucose and dosing strategy set to automatic bolus.
* Running dev, phone automatically boluses every 5 minutes, creating undesirable levels of IOB.
* With commit (updated to) fffcf505 of this branch, tested that the future glucose is detected and automated dosing stops

Display to User:
* Loop goes red over next 15 minutes, but need to add a better alerting system
* Tapping on glucose leads to the CGM
* Main screen continues to show the future glucose value (not CGM value)
* Tapping on bolus shows no recommendation with the new "Glucose in Future: Check for data in the future on phone" warning
* Tapping on meal entry allows the meal entry, but meal bolus screen is set for Save without Bolusing (no recommended bolus) with Glucose in Future warning
* In both cases, can bolus successfully if bolus value is added

Edited to change figure to that displayed with commit: ffbdc469

IMG_F8750773B91E-1

Debug log message:
* 2023-01-09 08:01:34.624572-0800 Loop[47592:5278115] [LoopDataManager] glucoseInFuture(date: 2023-01-10 14:11:00 +0000)

### DIFF
--- a/Loop/Managers/LoopDataManager.swift
+++ b/Loop/Managers/LoopDataManager.swift
@@ -1130,7 +1130,7 @@ extension LoopDataManager {
     ///     - LoopError.missingDataError
     ///     - LoopError.configurationError
     ///     - LoopError.glucoseTooOld
-    ///     - LoopError.glucoseInFuture
+    ///     - LoopError.invalidFutureGlucose
     ///     - LoopError.pumpDataTooOld
     fileprivate func predictGlucose(
         startingAt startingGlucoseOverride: GlucoseValue? = nil,
@@ -1158,7 +1158,7 @@ extension LoopDataManager {
         }
 
         guard lastGlucoseDate.timeIntervalSince(now()) <= LoopCoreConstants.futureGlucoseDataInterval else {
-            throw LoopError.glucoseInFuture(date: lastGlucoseDate)
+            throw LoopError.invalidFutureGlucose(date: lastGlucoseDate)
         }
 
         guard now().timeIntervalSince(pumpStatusDate) <= LoopCoreConstants.inputDataRecencyInterval else {
@@ -1395,7 +1395,7 @@ extension LoopDataManager {
     /// - Throws:
     ///     - LoopError.missingDataError
     ///     - LoopError.glucoseTooOld
-    ///     - LoopError.glucoseInFuture
+    ///     - LoopError.invalidFutureGlucose
     ///     - LoopError.pumpDataTooOld
     ///     - LoopError.configurationError
     fileprivate func recommendBolusValidatingDataRecency<Sample: GlucoseValue>(forPrediction predictedGlucose: [Sample],
@@ -1412,7 +1412,7 @@ extension LoopDataManager {
         }
 
         guard lastGlucoseDate.timeIntervalSince(now()) <= LoopCoreConstants.inputDataRecencyInterval else {
-            throw LoopError.glucoseInFuture(date: lastGlucoseDate)
+            throw LoopError.invalidFutureGlucose(date: lastGlucoseDate)
         }
 
         guard now().timeIntervalSince(pumpStatusDate) <= LoopCoreConstants.inputDataRecencyInterval else {
@@ -1526,7 +1526,7 @@ extension LoopDataManager {
     /// - Throws:
     ///     - LoopError.configurationError
     ///     - LoopError.glucoseTooOld
-    ///     - LoopError.glucoseInFuture
+    ///     - LoopError.invalidFutureGlucose
     ///     - LoopError.missingDataError
     ///     - LoopError.pumpDataTooOld
     private func updatePredictedGlucoseAndRecommendedDose(with dosingDecision: StoredDosingDecision) -> (StoredDosingDecision, LoopError?) {
@@ -1551,7 +1551,7 @@ extension LoopDataManager {
         }
 
         if glucose.startDate.timeIntervalSince(startDate) > LoopCoreConstants.inputDataRecencyInterval {
-            errors.append(.glucoseInFuture(date: glucose.startDate))
+            errors.append(.invalidFutureGlucose(date: glucose.startDate))
         }
 
         let pumpStatusDate = doseStore.lastAddedPumpData
--- a/Loop/Managers/LoopDataManager.swift
+++ b/Loop/Managers/LoopDataManager.swift
@@ -1130,6 +1130,7 @@ extension LoopDataManager {
     ///     - LoopError.missingDataError
     ///     - LoopError.configurationError
     ///     - LoopError.glucoseTooOld
+    ///     - LoopError.glucoseInFuture
     ///     - LoopError.pumpDataTooOld
     fileprivate func predictGlucose(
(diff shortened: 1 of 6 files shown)


Loop/Managers/LoopDataManager.swift
```diff
@@ -1156,6 +1157,10 @@ extension LoopDataManager {
             throw LoopError.glucoseTooOld(date: glucose.startDate)
         }
 
+        guard lastGlucoseDate.timeIntervalSince(now()) <= LoopCoreConstants.inputDataRecencyInterval else {
```
ps2, line 1160: I think this should use a different constant, as it's not checking "recency", and I'm not sure 15 minutes is right cutoff. The only future values we should allow are probably due to device clock drift, and that's probably a smaller window that 15 minutes.

Loop/View Models/BolusEntryViewModel.swift
```diff
@@ -673,6 +674,8 @@ final class BolusEntryViewModel: ObservableObject {
             switch error {
             case LoopError.missingDataError(.glucose), LoopError.glucoseTooOld:
                 notice = .staleGlucoseData
+            case LoopError.missingDataError(.glucose), LoopError.glucoseInFuture:
```
ps2, line 677: Does this even compile? It's duplicating a case above.

Loop/Models/LoopError.swift
```diff
@@ -183,6 +190,9 @@ extension LoopError: LocalizedError {
         case .glucoseTooOld(let date):
             let minutes = formatter.string(from: -date.timeIntervalSinceNow) ?? ""
             return String(format: NSLocalizedString("Glucose data is %1$@ old", comment: "The error message when glucose data is too old to be used. (1: glucose data age in minutes)"), minutes)
+        case .glucoseInFuture(let date):
+            let minutes = formatter.string(from: -date.timeIntervalSinceNow) ?? ""
+            return String(format: NSLocalizedString("Glucose data is %1$@ in the future", comment: "The error message when glucose data is in the future to be used. (1: glucose data time in future in minutes)"), minutes)
```
ps2, line 195: I think it should say that the future glucose data is invalid, and refer to the specific reading, rather than just "data": "Invalid glucose reading with a timestamp that is %1$@ in the future"

ps2: The translation comment doesn't really make sense either.

Loop/Views/BolusEntryView.swift
```diff
@@ -334,6 +334,11 @@ struct BolusEntryView: View {
                 title: Text("No Recent Glucose Data", comment: "Title for bolus screen notice when glucose data is missing or stale"),
                 caption: Text("Enter a blood glucose from a meter for a recommended bolus amount.", comment: "Caption for bolus screen notice when glucose data is missing or stale")
             )
+        case .futureGlucoseData:
+            return WarningView(
+                title: Text("Glucose in Future", comment: "Title for bolus screen notice when glucose data is in the future"),
```
ps2, line 339: Title: "Invalid  Future Glucose"
Caption: "Check your device time and/or remove any invalid data from Apple Health"

ps2 (changes requested):

Loop/Managers/LoopDataManager.swift
marionbarker: I agree - I stuck it in as a placeholder.
Does LoopCoreConstants.futureGlucoseDataInterval match the naming protocol?
I'll put in 5 minutes.

Loop/Views/BolusEntryView.swift
marionbarker: Done. Will push changes soon.

Loop/Models/LoopError.swift
marionbarker: I fixed the first one but I'm not sure about where the translation comment is.

marionbarker:
Updated - not sure about translation request.
Edited original comment with updated figure.

marionbarker:
Tests all passed, but I wonder if one should be added to: LoopTests/ViewModels/BolusEntryViewModelTests.swift

marionbarker:
Added two tests to LoopTests/ViewModels/BolusEntryViewModelTests.swift.

Loop/Managers/LoopDataManager.swift
```diff
@@ -1130,6 +1130,7 @@ extension LoopDataManager {
     ///     - LoopError.missingDataError
     ///     - LoopError.configurationError
     ///     - LoopError.glucoseTooOld
+    ///     - LoopError.glucoseInFuture
```
ps2, line 1133: nit: maybe `invalidFutureGlucose`?  Not a blocker, but "glucose in future" doesn't seem to convey any kind of problem.

Loop/Views/BolusEntryView.swift
```diff
@@ -334,6 +334,11 @@ struct BolusEntryView: View {
                 title: Text("No Recent Glucose Data", comment: "Title for bolus screen notice when glucose data is missing or stale"),
                 caption: Text("Enter a blood glucose from a meter for a recommended bolus amount.", comment: "Caption for bolus screen notice when glucose data is missing or stale")
             )
+        case .futureGlucoseData:
+            return WarningView(
+                title: Text("Invalid Future Glucose", comment: "Title for bolus screen notice when glucose data is in the future"),
+                caption: Text("Check for device time and/or remove any invalid data from Apple Health.", comment: "Caption for bolus screen notice when glucose data is in the future")
```
ps2, line 340: "Check for" -> "Check your"

ps2 (commented):
Just noticed a typo, and made an optional name change recommendation.

Loop/Models/LoopError.swift
ps2: The translation comment had a phrase "is in the future to be used", which I can't parse, but it looks like you fixed it. Thanks.

Loop/Managers/LoopDataManager.swift
marionbarker: I agree. Done.

Loop/Views/BolusEntryView.swift
marionbarker: Thanks. Done.

marionbarker:
All test cases succeeded in Xcode.
Built and tested on test phone (after last changes), commit 3038bc7e.

* Saved carbs with no bolus
  * It was not planned, but glucose was below low end of correction range, therefore no autobolus, which is expected behavior
  * Lowered low end of correction range and modified simulated CGM entry so Loop should give an autobolus and it did

* At 15:57, add single glucose 13 minutes in the future (16:10)
  * Main screen shows the glucose, no trend is visible
  * Got automatic dose at 15:58, might have been 15:58 by the time I hit save on glucose in health
  * Invalid Future Glucose when tap on Manual Bolus until time is 16:05
  * Next automatic dose given when next CGM reading comes in 16:06

ps2:
Thanks!

PR #1897, opened 2023-01-14
Open Loop with "Loop://" URLs from other applications

elnjensen:
This PR defines a general URL scheme for opening Loop from other applications.  There is a config variable that defines the URL to use, which by default is the app's display name.  I'll submit a separate PR to LoopWorkspace with a placeholder override variable showing how to override that. 

There is skeleton code that extracts the rest of the URL so that URLs like "Loop://bolus" could be used to link to the Bolus view directly, but right now I'm unsure of the right place/method for doing that.  I thought I'd start by PR'ing this part to get started, and perhaps you could advise on next steps for the deeper links.

### DIFF
--- a/Loop/AppDelegate.swift
+++ b/Loop/AppDelegate.swift
@@ -91,34 +91,4 @@ final class AppDelegate: UIResponder, UIApplicationDelegate, WindowProvider {
     func application(_ application: UIApplication, supportedInterfaceOrientationsFor window: UIWindow?) -> UIInterfaceOrientationMask {
         return loopAppManager.supportedInterfaceOrientations
     }
-    
-    // MARK: - UIApplicationDelegate - handle external URLs
-
-    func application(_ application: UIApplication,
-                     open url: URL,
-                     options: [UIApplication.OpenURLOptionsKey : Any] = [:] ) -> Bool {
-
-        // Process the URL.
-        guard let components = NSURLComponents(url: url, resolvingAgainstBaseURL: true), let linkTarget = components.host else {
-            // No sub-path specified
-            return true
-        }
-        
-        // Below examples would call something like the function presentCarbEntryScreen defined in StatusTableViewController; not sure if that type of code has to all be copied here, or if those various functions (e.g. carb entry, bolus view, settings) should be put elsewhere, or could be repurposed from existing files.
-
-        switch linkTarget.lowercased() {
-        case "carbs" :
-            // open carb entry view controller
-            return true
-        case "bolus" :
-            // open bolus view controller
-            return true
-        case "settings" :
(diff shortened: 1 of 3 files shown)


ps2:
Probably best to rebase this on the current dev; it has a lot of branch history atm from the master->dev sync, that doesn't need to be here.

elnjensen:
> Probably best to rebase this on the current dev; it has a lot of branch history atm from the master->dev sync, that doesn't need to be here.

Sorry about that!  Should be better now.

Loop/AppDelegate.swift
```diff
@@ -91,4 +91,34 @@ final class AppDelegate: UIResponder, UIApplicationDelegate, WindowProvider {
...
     }
+    
+    // MARK: - UIApplicationDelegate - handle external URLs
+
+    func application(_ application: UIApplication,
+                     open url: URL,
+                     options: [UIApplication.OpenURLOptionsKey : Any] = [:] ) -> Bool {
+
+        // Process the URL.
+        guard let components = NSURLComponents(url: url, resolvingAgainstBaseURL: true), let linkTarget = components.host else {
```
ps2, line 102: I'd prefer to not include any of this example/aspirational code yet. It'd be better to be able to evaluate it in the context of an actual feature that was using it.

ps2 (changes requested):

elnjensen:
OK, done.  When you have a chance (either here or in Zulip), I'd appreciate any thoughts on how to go a step further, i.e. best practice for switching view controllers to handle deeper links.

ps2:
Thanks!  I think for discussing navigating to specific places in Loop via the url scheme, why don't you open another PR with what you're thinking, and we can continue the discussion there?  As for some starting guidance, I think it would make more sense to have the URL paths be aligned more around use cases, like "/add_meal?carbs=35g&duration=3hours" rather than just navigation commands, like "/show_carb_entry_view_controller?..."

jgefaell:
When building archive to TestFlight, upload fails with below error if the MAIN_APP_DISPLAY_NAME does not conform to RFC1738. 

Changing the value from "Loop Dev" to something conformant like "Loop-Dev" avoids the error and allows successful upload to TestFlight but imposes on the app display name in iOS and Nightscout Loop pill.  

Building to a simulator device directly does not provoke an error. 

```
Asset Validation Failed

The following URL schemes found in your app are not in the correct format: 
[Loop Dev]. URL schemes need to begin with an alphabetic character, and be 
comprised of alphanumeric characters, the period, the hyphen or the plus sign only. Please 
see RFC1738 for more detail. (ID: 323a6d62-ab9d-4044-855e-2f598a6bc7b9)
```

PR #1903, opened 2023-01-18
Add X-Large Watch Complications

IsThisPaul:
Add support for X-Large watch complications


ps2:
Thanks! Fixed with #1901

PR #1917, opened 2023-01-31
Add profile expiration time to settings page

Chris-Almond:
Adding the ability for users to see when their profile expires without waiting for the pop-up warning modal message. 

This small enhancement gives users more visibility into the remaining life of their Loop app, to help them plan to rebuild and load fresh profiles.

### UI Before 

beforeLight 

### After changes

\> 3 Days Dark

 greaterThan3DaysDark 

\> 3 Days Light 

greaterThan3DaysLight 

< 3 Days Dark 

lessThan3DaysDark 

< 3 Days Light 

lessThan3DaysLight

### DIFF
--- a/Common/FeatureFlags.swift
+++ b/Common/FeatureFlags.swift
@@ -37,6 +37,7 @@ struct FeatureFlagConfiguration: Decodable {
     let usePositiveMomentumAndRCForManualBoluses: Bool
     let dynamicCarbAbsorptionEnabled: Bool
     let adultChildInsulinModelSelectionEnabled: Bool
+    let profileExpirationSettingsViewEnabled: Bool
 
 
     fileprivate init() {
@@ -212,6 +213,12 @@ struct FeatureFlagConfiguration: Decodable {
         #endif
 
         self.dynamicCarbAbsorptionEnabled = true
+        
+        #if PROFILE_EXPIRATION_SETTINGS_VIEW_ENABLED
+        self.profileExpirationSettingsViewEnabled = true
+        #else
+        self.profileExpirationSettingsViewEnabled = false
+        #endif
     }
 }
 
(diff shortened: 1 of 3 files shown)


ps2:
Nice! Profile Expiration and LoopDocs are DIY Loop specific things, so it should probably live in LoopSupport, which is a DIY Plugin. If you can move it there, I'd be happy to accept this PR. Another option would be to put it behind a feature flag.

ps2 (changes requested):
Nice! Profile Expiration and LoopDocs are DIY Loop specific things, so it should probably live in LoopSupport, which is a DIY Plugin. If you can move it there, I'd be happy to accept this PR. Another option would be to put it behind a feature flag.

Chris-Almond:
Added a feature flag `PROFILE_EXPIRATION_SETTINGS_VIEW_ENABLED` in the latest commit with the default as disabled. 

1. How does this look?
2. Do you prefer I squash these commits together myself (or will GitHub handle that during the merge)?
3. Do you want me to create a PR adding this flag in a commented out state to `LoopWorkspace`'s `LoopConfigOverride.xcconfig`? If so, any specific branch I should target? (sample change below)
```
 // Features
-SWIFT_ACTIVE_COMPILATION_CONDITIONS = $(inherited) SIMULATORS_ENABLED //DEBUG_FEATURES_ENABLED
+SWIFT_ACTIVE_COMPILATION_CONDITIONS = $(inherited) SIMULATORS_ENABLED //PROFILE_EXPIRATION_SETTINGS_VIEW_ENABLED //DEBUG_FEATURES_ENABLED
```

ps2:
I can squash commits during merge, thanks.  Generally the default state is the DIY state, so I think the flag should be such that without any extra compilation conditions, the profile UI is shown. And we can add the flag to disable it in builds where it shouldn't be shown, so you would not need to submit a Workspace PR.

Chris-Almond:
Sounds great. Thanks. 

Ok, so my action item is to invert the flag to let the default be enabled. 

Anything else or good to go after that?

Chris-Almond (commented):
Inverted the feature flag so the default is enabled. 

Also squashed a tiny rendering bug that shows hours when days remaining > 3 and hours == 0.

PR #1930, opened 2023-02-19
Remote 1: Action Models

gestrich:
This will migrate away from the RemoteCommand type to the RemoteAction type. I'd like to separate actions, things that get enacted in Loop, from the RemoteCommand that delivered the action. This should help with unit testing and Remote Command 2.0 where this distinction is more meaningful. For now, I've remote the RemoteCommand type in favor of RemoteAction, however I plan to reintroduce it when Remote Commands 2.0 will bring the queuing capabilities (and responsibilities for a command)

Note that the related LoopKit PR will need merged with this https://github.com/LoopKit/LoopKit/pull/453


Loop/Managers/DeviceDataManager.swift
```diff
@@ -812,6 +812,18 @@ extension DeviceDataManager {
             self.loopManager.updateRemoteRecommendation()
         }
     }
+
```
gestrich, line 815: Async wrapper

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1341,8 +1353,15 @@ extension Notification.Name {
 
 // MARK: - Remote Notification Handling
 extension DeviceDataManager {
```
gestrich, line 1355: High-level changes in this class:

1. Make an async variant of `handleRemoteNotification`. This will be especially helpful for remote commands 2 later.
2. Use the new RemoteAction models for enacting a command. 
3. Add methods for handling async background uploads

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1372,156 +1388,144 @@ extension DeviceDataManager {
...
         
-        let command: RemoteCommand
+        let action: RemoteAction
         
         do {
-            command = try RemoteCommand.createRemoteCommand(notification: notification, allowedPresets: loopManager.settings.overridePresets, defaultAbsorptionTime: carbStore.defaultAbsorptionTimes.medium).get()
+            action = try RemoteAction.createRemoteAction(notification: notification).get()
         } catch {
             log.error("Remote Notification: Parse Error: %{public}@", String(describing: error))
             return
```
gestrich, line 1397: There were all sorts of validation logic for specific command types in this method previously. The new paradigm here is to move the validation logic into the RemoteAction. That helps hide this complexity and makes it more easy to unit test.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1372,156 +1388,144 @@ extension DeviceDataManager {
             }
         }
         
-        let command: RemoteCommand
+        let action: RemoteAction
```
gestrich, line 1392: RemoteCommand was renamed to RemoteAction. Later on I plan to reintroduce a RemoteCommand, but it will instead _have_ a RemoteAction. The distinction is a command will include the action and all the ancillary things related to commands (security validation, status reporting, etc). 

Separating these concepts should make more sense with Remote 2.0.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1372,156 +1388,144 @@ extension DeviceDataManager {
...
+        case .carbsEntry(let carbAction):
+            do {
+                try validatePushNotificationSource(notification: notification)
+                try await handleRemoteCarbAction(carbAction)
+            } catch {
+                await NotificationManager.sendRemoteCarbEntryFailureNotification(for: error, amountInGrams: carbAction.amountInGrams)
+                log.error("Remote Notification: Carb Action Error: %{public}@", String(describing: notification))
             }
         }
     }
```
gestrich, line 1433: These "handle" methods below all take an action and attempt to enact it.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1372,156 +1388,144 @@ extension DeviceDataManager {
...
+        switch validationResult {
+        case .success():
+            log.info("Remote Notification: Validation successful")
+        case .failure(let error):
+            throw error
+        }
+    }
+    
+    //Remote Overrides
+
```
gestrich, line 1449: The paradigm here is each RemoteAction is asked for its data required to enact it (i.e. TemporaryOverride, CarbEntry, etc). 

That call allows the action to validate that its data is normalized wrt to Loop's expectations (ex: carbs not exceeding reasonable max)

Loop/Managers/RemoteDataServicesManager.swift
```diff
@@ -209,6 +209,14 @@ final class RemoteDataServicesManager {
             completion()
         }
     }
```
gestrich, line 211: Async wrapper

Loop/Models/RemoteCommand.swift
```diff
@@ -8,78 +8,34 @@
 
 import Foundation
 import LoopKit
-import HealthKit
-
```
gestrich, line 12: Not needed as each remote action type will own its own errors.

Loop/Models/RemoteCommand.swift
```diff
@@ -8,78 +8,34 @@
...
-enum RemoteCommand {
-    case temporaryScheduleOverride(TemporaryScheduleOverride)
-    case cancelTemporaryOverride
-    case bolusEntry(Double)
-    case carbsEntry(NewCarbEntry)
-}
-
 
 // Push Notifications
```
gestrich, line 12: Generally like before, this method parses a notification and extracts a remote action (previously command).

The difference is Loop valid action checks are not done here (ex: Carb amount within reasonable limits). That will instead be managed by the client. 

Separating notification payload parsing and validation seems like a good idea here. This will go away when I use Codable to do all this parsing in Remote 2.0.

Loop/Models/RemoteCommand.swift
```diff
@@ -8,78 +8,34 @@
...
         } else if let carbsValue = notification["carbs-entry"] as? Double {
-
-            let minAbsorptionTime = TimeInterval(hours: 0.5)
-            let maxAbsorptionTime = LoopConstants.maxCarbAbsorptionTime
             
-            var absorptionTime = defaultAbsorptionTime
+            var absorptionTime: TimeInterval? = nil
             if let absorptionOverrideInHours = notification["absorption-time"] as? Double {
                 absorptionTime = TimeInterval(hours: absorptionOverrideInHours)
             }
```
gestrich, line 35: Carb food type was missing from the APIs. This will require a change to the Nightscout codebase in order to support this but should be ok when empty too.

LoopTests/Models/RemoteActionTests.swift
```diff
@@ -0,0 +1,108 @@
+//
```
gestrich, line 1: I needed to update the RemoteCommandTests with these and change a few that are unnecessary.

I added more tests to LoopKit, which test the individual remote action types themselves which should cover all the cases I know of.

gestrich:
Updates done - I think I responded to all the feedback but let me know if other questions.

ps2 (approved):
LGTM! Thanks! I think at some point it'd be good to move more of the remote command processing out of DeviceDataManager. Maybe we have a RemoteCommandManager, or some other high level Loop manager responsible for processing remote commands.

gestrich:
@ps2 The last commit in this PR didn't seem to make it in the merge from what I can tell db2ef8d7c9b6ef75358e735557fad25743eec9ce. Just some info on what that commit does...

Since the changes earlier in the PR to using Apple async apis, I found I needed to request background processing time before starting an async task. Otherwise, Loop would hang in the background before it makes it into the async block (doing in the block is too late). My testing showed that change seemed to work as well as Loop 3.0.

PR #1934, opened 2023-02-25
Remote PR Set #2: Introduce RemoteCommands

gestrich:
This PR is part of the below set that should be merged together. This one will use "RemoteCommands", a protocol for interacting with the remote services that is the source of the RemoteCommand. This will lay the groundwork for remote 2.0 commands.

* Loop: https://github.com/LoopKit/Loop/pull/1934
* LoopKit: https://github.com/LoopKit/LoopKit/pull/455
* NightscoutService: https://github.com/ps2/NightscoutService/pull/33
* TidepoolService: https://github.com/LoopKit/TidepoolService/pull/4


Loop.xcodeproj/project.pbxproj
```diff
@@ -383,7 +383,6 @@
 		A9DF02CB24F72B9E00B7C988 /* CriticalEventLogTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A9DF02CA24F72B9E00B7C988 /* CriticalEventLogTests.swift */; };
 		A9DFAFB324F0415E00950D1E /* CarbBackfillRequestUserInfoTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A9DFAFB224F0415E00950D1E /* CarbBackfillRequestUserInfoTests.swift */; };
 		A9DFAFB524F048A000950D1E /* WatchHistoricalCarbsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A9DFAFB424F048A000950D1E /* WatchHistoricalCarbsTests.swift */; };
```
gestrich, line 385: The RemoteCommand and RemoteCommandTests will now be in LoopKit. See the LoopKit PR for more info on its new role.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1376,81 +1376,56 @@ extension DeviceDataManager {
             log.error("Remote Notification: Remote Commands not enabled.")
             return
         }
-
```
gestrich, line 1379: I consider the validation of remote commands to be owned by the domain of the provider (plugin) that produced the command (ex: OTP and Expiration). This allows the providers to have their own validation criteria and Loop to not be constrained by older provider versions.

There is an argument that certain things happen in the provider (maybe OTP) and certain things in Loop (i.e. Expiration and later dose hashing).

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1376,81 +1376,56 @@ extension DeviceDataManager {
...
-                    await NotificationManager.sendRemoteCommandExpiredNotification(timeExpired: expiredInterval)
-                    log.error("Remote Notification: Expired: %{public}@", String(describing: notification))
-                    return
-                }
-            } else {
-                log.error("Remote Notification: Invalid expiration: %{public}@", expirationStr)
-                return
-            }
-        }
-
```
gestrich, line 1396: The remote provider will now be responsible for push notification parsing which may vary across providers and provider versions.

Loop/Managers/DeviceDataManager.swift
```diff
@@ -1376,81 +1376,56 @@ extension DeviceDataManager {
...
             return
         }
         
-        log.default("Remote Notification: Handling action %{public}@", String(describing: action))
+        await handleRemoteCommand(command)
+    }
+    
+    func handleRemoteCommand(_ command: RemoteCommand) async {
+        
+        log.default("Remote Notification: Handling command %{public}@", String(describing: command))
```
gestrich, line 1394: The RemoteCommand can communicate with its remote service. This will be helpful in the future where command status will be reported back to the service. 

I'm leveraging that capability here where validate() is now called directly on the command without needing to look up the originating source in this code, as was previously done in validatePushNotificationSource.

Loop/Managers/RemoteDataServicesManager.swift
```diff
@@ -597,13 +597,11 @@ extension RemoteDataServicesManager {
 
 extension RemoteDataServicesManager {
```
gestrich, line 599: Getting rid of validatePushNotificationSource as mentioned above. 

The method serviceForPushNotification will find the appropriate service for a push notification. We will use that service then to ask for a RemoteCommand from the push notification.

Loop/Managers/RemoteDataServicesManager.swift
```diff
@@ -644,18 +642,22 @@ extension RemoteDataServicesManager {
             UserDefaults.appGroup?.deleteQueryAnchor(for: remoteDataService, withRemoteDataType: .overrides)
         }
     }
+}
+
+extension RemoteDataServicesManager: RemoteCommandSource {
```
gestrich, line 648: This is async as someday the service will make an API request to put a remote lock on the command state (i.e. Set to in-progress state)

Loop/Managers/DeviceDataManager.swift
gestrich: The only reason I validate the command in each switch case below, rather than in just one place above: I need to use the appropriate error handler which vary for remote carbs vs remote bolus. The overall error handling of commands below is something to consider refactoring in the future but I wanted to keep this PR scoped.

Loop/Models/RemoteCommand.swift
```diff
@@ -1,59 +0,0 @@
-//
```
gestrich, line 1: These were replaced by Codable models in rileylink_ios

LoopTests/Models/Remote/RemoteCommandTests.swift
```diff
@@ -1,108 +0,0 @@
-//
```
gestrich, line 1: These tests were refactored/expanded and moved to rileylink_ios

PR #1948, opened 2023-03-14
Add vertical time bars to charts

Chris-Almond:
### Add a red vertical "current time line" to indicate the current time on charts. 

I have found this useful when looking at the **IOB, iDelivery, and COB** charts to see "where are we?" at a quick glance. 

Rationale: Since the time windows included on the charts roll forward on an hourly(?) cadence in chunks, it becomes difficult for a user to distinguish where they are within a basal adjustment window or active carb period.

1. This PR adds the utility function and adds this line exclusively to the loop charts that do **not** include a _past_ vs _future_ delineation. The historical vs predicted glucose chart distinguishes past vs future using dots and dashes, which I personally find effective, so I did not add this layer to that chart.
2. The "current time line" stays in the appropriate screen position regardless of if the phone is in portrait or landscape orientation. Screenshots included

This is one of a handful of small UI enhancements I have locally, so offering it up to see if there's interest.

Screenshots: 
Simulator Screen Shot - iPhone 14 - 2023-03-14 at 00 13 15
FullPortraitDark
GlucoseLandscapeDark
InsulinAndCarbsDark
InsulinDark
Simulator Screen Shot - iPhone 14 - 2023-03-13 at 23 56 33
Simulator Screen Shot - iPhone 14 - 2023-03-13 at 23 56 59
Simulator Screen Shot - iPhone 14 - 2023-03-13 at 23 57 09

### DIFF
--- a/LoopUI/Charts/COBChart.swift
+++ b/LoopUI/Charts/COBChart.swift
@@ -74,13 +74,16 @@ public extension COBChart {
             )
         }
 
+        let currentTimeLineLayer = VerticalTimeChartLine.create(xAxis: xAxisLayer.axis, yAxis: yAxisLayer.axis)
+        
         let layers: [ChartLayer?] = [
             gridLayer,
             xAxisLayer,
             yAxisLayer,
             cobChartCache?.highlightLayer,
             cobArea,
-            cobLine
+            cobLine,
+            currentTimeLineLayer,
         ]
 
         return Chart(frame: frame, innerFrame: innerFrame, settings: chartSettings, layers: layers.compactMap { $0 })
--- a/LoopUI/Charts/DoseChart.swift
+++ b/LoopUI/Charts/DoseChart.swift
@@ -128,6 +128,8 @@ public extension DoseChart {
             )
         }
 
+        let currentTimeLineLayer = VerticalTimeChartLine.create(xAxis: xAxisLayer.axis, yAxis: yAxisLayer.axis)
+
         let layers: [ChartLayer?] = [
             gridLayer,
             xAxisLayer,
@@ -136,7 +138,8 @@ public extension DoseChart {
             doseChartCache?.highlightLayer,
             doseArea,
             doseLine,
-            bolusLayer
+            bolusLayer,
+            currentTimeLineLayer,
         ]
 
         let chart = Chart(frame: frame, innerFrame: innerFrame, settings: chartSettings, layers: layers.compactMap { $0 })
--- a/LoopUI/Charts/IOBChart.swift
+++ b/LoopUI/Charts/IOBChart.swift
@@ -88,6 +88,8 @@ public extension IOBChart {
             )
         }
 
+        let currentTimeLineLayer = VerticalTimeChartLine.create(xAxis: xAxisLayer.axis, yAxis: yAxisLayer.axis)
+        
         let layers: [ChartLayer?] = [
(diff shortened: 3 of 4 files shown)


ps2:
I'm not a fan of this line. Mainly aesthetics, but also for UX simplification. These charts (at least in their default form) are attempting to convey a rough idea of state at a glance, and not more fine grained things like "where they are within a basal adjustment window or active carb period". Those can be found by inspection if needed. If those are high-frequency questions, it seems like maybe something else is off? These are not considered primary uses. Maybe we should be considering them for some reason. Why are you wanting to know where you are within a basal adjustment window, or active carb period?

In lieu of adding additional use cases to support for this view, other things that might make this more tenable to merge, are:
* Making this an optional part of the display, with a setting that can turn it on or off. The default should be disabled.
* Making the line only appear on inspection.
* Making the line aesthetically not so jarring.

Chris-Almond:
Hi Pete, I appreciate the thorough response! 

Q: When you say “inspection” are you referring to the landscape rotated view of these charts, or is an inspection something else?

Aesthetics wise, I think the **dots and dashes** of **historical vs predicted** shown by the top glucose chart is a nice subtle but clear way to distinguish the past, present, and future. 

I drafted this PR because my brain wants a similar delineation in the bottom 3 charts and I figured a thin vertical line was a simple way to visualize the same 😄. I’m not married to the vertical line so I'm interested to hear your thoughts on a more minimal (less abrasive?) way to show the same.

> Why are you wanting to know where you are within a basal adjustment window, or active carb period?

Good question! I _personally_ like the added relativity in the “at a glance” view of how long ago the preceding bolus/basal adjustment event occurred. That offers added context and confidence to help me make future decisions, especially when trusting auto-bolus/basal to course-correct. Similarly, I interpret the RHS of the (_currently red_) line of the COB/IOB as “gas already in the tank that still may yet still drive BG up/down”, respectively. 

> - Making this an optional part of the display, with a setting that can turn it on or off. The default should be disabled.
      -  No problem. I can look into adding a `Chart Settings` section under `Settings`, but to your 3rd bullet below, perhaps this needs some further refining before slapping something together. I'd prefer a good UI view that makes the "on" option less "jarring". I audibly laughed at "The default should be disabled" 😆. 
> - Making the line only appear on inspection.
      - See above Q on inspection(s). 
> - Making the line aesthetically not so jarring.
      - Interested to hear your thoughts on alternative views that would offer a be less abrasive UX. 

Cheers & thanks

ps2:
As for design ideas, reducing the value or tone contrast.  Maybe something like a vertical edge blur. I'm not sure. Some experimentation would be called for.

Chris-Almond:
I’ll look through the codebase for examples of correctly persisting user preferences. I’m thinking an on/off toggle and a color picker combo. 

Does that sound like a reasonable path forward?

ps2:
We do not have an example of a setting like this, so it will take some work. I think we should have a new "Display Options" row in the "Configuration" section of Settings. And in there it should be an on/off toggle.  Do not have a color picker. It'd be better to pick a set of colors that are visible, but at a visual priority equivalent with the axis grid marks, and that work well in both light and dark modes.

ps2:
Closing this for inactivity. If you are able to add a menu for this option to be enabled, and can present a design with appropriate visual priority and dark mode support, then I think this can be accepted.

PR #1958, opened 2023-03-20
feat(basal): adding in manual basal entry

bassrock:
Allows manual entry screen to support multiple insulin types. Works in coordination with https://github.com/LoopKit/LoopKit/pull/464 which adds Lantus as a supported manual entry type.


ps2:
I think this open loop use case needs to be thought through more, and a coherent design created, if we want to support this "manual dosing - open loop" use case in Loop. Just adding a lantus option to the existing manual dose interface makes the interface more confusing to closed loop users who are using the manual dose entry, and still many other parts of the Loop UI don't make any sense for manual dose open loop users.

I wonder about creating a "MDI" PumpManager, and Loop would know that it's a pump that cannot support automation, but would rather present recommendations. And this could allow other alerts, like a "you haven't entered your basal dose" reminder, or generate a notification when the amount of insulin that Loop recommends crosses a threshold.

PR #1979, opened 2023-04-13
Fixed numeral localization in insulin delivery table view

### LINKED ISSUE #1968: Some Arabic numbers appear in the form of symbols
**Describe the bug**
By using the Arabic language, some numbers appear in the form of symbols, as shown in the pictures
From Insulin Delivery page (Event History)

**Screenshots**

**Phone**
 - Hardware: iPhone 11
 - OS Version:  iOS 16.3.1

**Loop Version**
 - Version Number: 3.2.1

**CGM**
 - Device: Dexcom G6
 - Manager app: Dexcom App

**Pump**
 - Manufacturer: Omnipod Dash

**Additional context**
I think if English numbers are used in the App, the problem could be solved.

hannkm:
Fixes #1968 
This fixes the issue of symbols appearing instead of numerals. Changes mainly affect Arabic but all numbers should be displayed correctly according to their language now.

### DIFF
--- a/Loop/View Controllers/InsulinDeliveryTableViewController.swift
+++ b/Loop/View Controllers/InsulinDeliveryTableViewController.swift
@@ -609,7 +609,7 @@ fileprivate var numberFormatter: NumberFormatter {
 fileprivate func createAttributedDescription(from description: String, with font: UIFont) -> NSAttributedString? {
     let descriptionWithFont = String(format:"<style>body{font-family: '-apple-system', '\(font.fontName)'; font-size: \(font.pointSize);}</style>%@", description)
 
-    guard let attributedDescription = try? NSMutableAttributedString(data: Data(descriptionWithFont.utf8), options: [.documentType: NSAttributedString.DocumentType.html], documentAttributes: nil) else {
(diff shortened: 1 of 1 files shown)


hannkm:
This is the issue replicated before the change:
before
This is what it looks now after the fix:
after

ps2:
Thanks, can you describe *why* this is an appropriate change? I don't see it documented that this method expects utf16, and that would be surprising, as usually utf8 is appropriate for these.  See this for a possible alternate solution: https://stackoverflow.com/questions/61411944/nsattributedstring-from-utf16-characters-in-uitextview

hannkm:
Here are some articles I referenced:
https://www.objc.io/issues/9-strings/unicode/ (https://www.objc.io/issues/9-strings/unicode/)
This is the official Apple documents:
https://developer.apple.com/documentation/foundation/nsstring (https://developer.apple.com/documentation/foundation/nsstring)
NSMutableAttributedString requires non-ASCII values to be encoded in UTF-16, and I believe this is the only function (as submitted in this PR) that utilized utf8 encoding for the data entries previously. 

Could you clarify where I should document this?

PR #1988, opened 2023-04-28
Glucose Based Partial Application Factor

marionbarker:
Please see zulipchat topic [Dosing Strategy Linear Ramp](https://loop.zulipchat.com/#narrow/stream/144182-development/topic/Dosing.20Strategy.20Linear.20Ramp)

Please see associated [LoopKit PR 477](https://github.com/LoopKit/LoopKit/pull/477)


marionbarker:
Notes:
* If this moves forward, I will add test cases.
* This draft has a print statement that should either be removed or made into proper log statement.

marionbarker:
Graphic showing the slider on the Dosing Screen:

image

Definition of the linear ramp:
* minGlucoseSlidingScale = correction range lower bound + 10 mg/dL (minGlucoseDeltaSlidingScale)
* maxGlucoseSlidingScale = 250 mg/dL
* minPartialApplicationFactor = 0.15
* maxPartialApplicationFactor = 1.00

when current glucose is A, effectiveBolusApplicationFactor is B:
* at or below min glucose, 0.15
* between min and max glucose, linear ramp from minFactor to maxFactor 
* at or above max glucose, 1.00

Note - started out with terminology sliding scale. Not all variables were renamed linear ramp - but I expect many changes will occur, so waiting to update the names.

marionbarker:
Note - this is a DRAFT and subject to change.
This code works as designed, but there's a lot of room for discussion on the design and implementation.

I've started a google sheet [Evaluate AB, TB and Fractions](https://docs.google.com/spreadsheets/d/1DUzLTDRpVj41o8adzXZbtf_ZaPatIRbQaGvLHVSl2cA/edit?usp=sharing) for doing calculations. (This is a work-in-progress document that any one with a link can view, but only I can edit.)

Question 1: How do these fractions compare to default AB of 40%

I've captured 2 screenshots from the sheet for particular values:

PR-1988-01

PR-1988-02

marionbarker:
Question 2: What is the difference between Temp Basal and Automatic Bolus (various Fractions) in terms of dosing over time.

In [Evaluate AB, TB and Fractions](https://docs.google.com/spreadsheets/d/1DUzLTDRpVj41o8adzXZbtf_ZaPatIRbQaGvLHVSl2cA/edit?usp=sharing), the temp basal indicated is added to scheduled basal; I'm not explicitly showing the scheduled basal - only TB above that.

This one was fun and I learned some things. After playing around with this, I think we should modify min / maxPartialApplicationFactor to 0.2 / 0.8.

* Reminder to people who want more aggressive calculation for correction
    * That is NOT the purpose of this PR
    * This PR is to assist people who prefer Loop be slower to respond at low glucose and faster to respond as glucose increases
    * Any modifications to recommended correction model must be in a separate PR
    * This PR needs to "play nice" with any other PR that might change the modeling

With the idealization (in the sheet) that the recommended correction at the start doesn't change over time, there are some recommended correction values for which AB never reaches the full amount. (The last 0.05 U times the AB fraction puts it below the pump rounder.)

In reality, if AB or TB deliver less insulin than is actually needed, rising glucose levels will update the prediction and thus an increase in the correction required.

The graphic below compares correction delivery over half an hour for Temp Basal and Automatic Bolus with 3 values for AB Fraction. For all of these, I used the pod delivery model (temp basal as late in an interval as possible and 0.05 U minimum dose).

**Graphic: 1.00 U correction, Basal Not Limited, AB Fractions = 0.20, 0.40, 0.80**


PF-1988-03-not-basal-limited

marionbarker:
I know this PR requires a lot of changes before it can be accepted. However, I am using it as a patch with dev on my own device. 

With the updates to dev on May 6, 2023 (LoopWorkspace commit 00f7b05), the patch required an update to apply.

After updating the Loop and LoopKit PR for dev, I then did a second update to modify the constants. I made these changes after spending some time looking at the calculations for different lower bound settings and constant values.

The plot compares the PAF for an "aggressive" looper with a lower bound correction range of 87 and a "conservative" looper with a lower bound correction range of 120. The dashed lines are the original constants from April 28. The solid lines are the updated constants from May 6. See the chart below the plot for the constants before and after. That chart also shows values at which the PAF hits 40% (original AB setting).

PR-1988-05_new-constants-plot

PR-1988-05_new-constants-table

marionbarker:
I rephrased the words seen by the user. There are other requests from Pete via DM that have not yet been addressed.


PR-1988_update_wording

marionbarker:
Thanks to @bjorkert. Pushed changes to enable swapping of original and replacement protocols and to minimize LoopKit changes

Loop/Managers/DosingStrategy.swift
```diff
@@ -0,0 +1,20 @@
...
+//  Created by Jonas Björkert on 2023-06-03.
+//  Copyright © 2023 LoopKit Authors. All rights reserved.
+//
+
+import Foundation
+import HealthKit
+import LoopKit
+import LoopCore
+
+protocol DosingStrategy {
```
ps2, line 14: "Dosing Strategy" is a term already used to describe temp basal vs automatic bolus.  Maybe call this `ApplicationFactorStrategy`?

Loop/Managers/LinearRampDosingStrategy.swift
```diff
@@ -0,0 +1,36 @@
...
+//  Created by Jonas Björkert on 2023-06-03.
+//  Copyright © 2023 LoopKit Authors. All rights reserved.
+//
+
+import Foundation
+import HealthKit
+import LoopKit
+import LoopCore
+
+struct LinearRampDosingStrategy: DosingStrategy {
```
ps2, line 14: "Linear Ramp" is more a detail about how it's doing the calculation. The fact that it's adjusting the application based on glucose is a more primary feature of the strategy. Maybe "GlucoseBasedApplicationFactor" would be better.

Loop/Managers/LoopDataManager.swift
```diff
@@ -67,6 +67,8 @@ final class LoopDataManager {
 
     private var insulinOnBoard: InsulinValue?
 
+    var dosingStrategy: DosingStrategy = OriginalDosingStrategy() // default to original strategy
```
ps2, line 70: Rather than "Original", maybe call it "Constant"

Loop/Managers/LoopDataManager.swift
```diff
@@ -1698,7 +1700,23 @@ extension LoopDataManager {
                     return self.delegate?.roundBolusVolume(units: units) ?? units
                 }
 
-                let maxAutomaticBolus = min(iobHeadroom, maxBolus! * LoopConstants.bolusPartialApplicationFactor)
+                // Create dosing strategy based on user setting
+                let dosingStrategy: DosingStrategy = UserDefaults.standard.bool(forKey: "applyLinearRampToBolusApplicationFactor")
+                ? LinearRampDosingStrategy()
```
ps2, line 1705: Indentation?

Loop/Managers/LoopDataManager.swift
```diff
@@ -1698,7 +1700,23 @@ extension LoopDataManager {
                     return self.delegate?.roundBolusVolume(units: units) ?? units
                 }
 
-                let maxAutomaticBolus = min(iobHeadroom, maxBolus! * LoopConstants.bolusPartialApplicationFactor)
+                // Create dosing strategy based on user setting
+                let dosingStrategy: DosingStrategy = UserDefaults.standard.bool(forKey: "applyLinearRampToBolusApplicationFactor")
```
ps2, line 1704: Storing this in user defaults is fine for algorithm experiments, but please put "experimental" in the key name. If this ever makes it as a standard options of the Loop algorithm, it will go into LoopSettings

Loop/Models/LoopConstants.swift
```diff
@@ -75,4 +75,12 @@ enum LoopConstants {
     static let simpleBolusCalculatorMinGlucoseBolusRecommendation = HKQuantity(unit: .milligramsPerDeciliter, doubleValue: 70)
     static let simpleBolusCalculatorMinGlucoseMealBolusRecommendation = HKQuantity(unit: .milligramsPerDeciliter, doubleValue: 55)
     static let simpleBolusCalculatorGlucoseWarningLimit = HKQuantity(unit: .milligramsPerDeciliter, doubleValue: 70)
+
+    // Dosing Stategy: Automatic Bolus: Sliding Scale
+    static let minPartialApplicationFactor = 0.20 // min fraction of correction when glucose > minGlucoseSlidingScale
```
ps2, line 80: Keep these constants in the strategy implementation file for now, please.

Loop/Views/DosingStrategySelectionView.swift
```diff
@@ -45,27 +47,52 @@ public struct DosingStrategySelectionView: View {
...
+                ),
+                trailingView: strategy.isAutomaticBolus ? linearRampVolusApplicationFactorSection : nil
             )
             .padding(.vertical, 4)
         }
     }
 }
 
+extension DosingStrategySelectionView {
+    var linearRampVolusApplicationFactorSection: AnyView {
```
ps2, line 59: Would prefer that this go in an entirely new section entitled "Algorithm Experiments", separate from bolus/basal dosing strategy, and be a single line with the title and an "on/off" *label*, and a disclosure arrow.  Tapping on the line would take you a full page description of the option, and the ability to enable/disable it.

I understand that it's convenient to make it part of the dosing strategy, since it doesn't apply to Temp Basal, but this should be kept separate in the UI for now, as it's an experimental option.

ps2 (changes requested):
Heading in the right direction. Some naming change requests, and a UI request. Also, whether or not this option is enabled should show up in an issue report.

marionbarker:
Comments above addressed with commit 8b3f61e.

Will post updated graphics soon.

marionbarker:
Graphic showing new User Interface.

pr1988-ui-proposed

marionbarker:
pr1988-graph-bolus-application

ps2:
Thanks Marion.  Haven't reviewed the code changes, but as for UI (thanks for posting screenshots), the experiment name should stick to a neutral description of what it does, and not try to have positive connotations (enhanced, improved, etc...). To me, I see this as Glucose Based Partial Application. It's not short, but it's descriptive. This is adjusting partial application based on glucose level.

ps2:
A couple more tweaks:
* The experiment name in the algorithm experiments section does not need "Enable", imo.  Just the title would do. It looks like an action to say "Enable x"
* The switch on the description page should be moved to the bottom of the text, or even on a second page, to encourage reading of the descriptive text. It's important to understand an experiment before trying it.
* I think a little warning triangle at the end of the "Algorithm Experiments" section title would be appropriate.

marionbarker:
Updated graphic for User Interface:

pr1988-ui-proposed-2

diggabyte:
Would it be possible to allow for also choosing dosing amount > 80% in the upper bound? Perhaps even to 100%? Or is this a serious safety issue?

I often find myself manually triggering a bolus to fill in the 20%, particularly during periods of elevated BG.

diggabyte:
Also wondering if the "high" threshold could be variable as well? 200mg/dL is nice default. But for some dialed-in / expert Loop users, "high" may be more like 160mg/dL, for example.

marionbarker:
Thanks to @bjorkert (repo: cgm-simulator), I can run automated tests with a fixed input glucose profile uploaded at 5 minute intervals to Nightscout, which I can use that Nightscout URL as both a CGM source and an upload site for Loop. I can then extract the data for AB and TB from Nightscout for each test.

I have also switched from using an rPi DASH simulator to using the in-the-app pump simulator configured as a pod. (I don't know if it was my network or the pi but I was always getting at least one communication dropout during tests - this messed up the Loop cycle during the tests.)

The following plot is the integrated amount of insulin delivered (increases from AB and decreases from TB) for two tests using the same input glucose profile. 
* Test labeled RC: nominal Loop-dev version running with standard retrospective correction and automatic bolus using a constant 40% factor
* Test labeled RC+GBAF: running with standard retrospective correction and automatic bolus using the Glucose Based Application Factor

As expected the RC+GBAF is a little slower at the beginning of the test (lower glucose) to provide extra insulin, but as glucose increases, it is faster. By the time the test is over the total insulin delivered is the same.

pr1988-compare-GBAF-unannounced-meal

ps2 (approved):

PR #1989, opened 2023-05-01
Disable missed meal detection if there are user-entered or calibration BGs

novalegra:
This PR disables missed meal detection if there are BG points in the missed meal detection window that are calibration or user-entered, since these can cause large jumps and cause false-positive detections. Associated [LoopKit PR](https://github.com/LoopKit/LoopKit/pull/478)


ps2:
Updated version here: https://github.com/LoopKit/Loop/pull/2056

PR #1997, opened 2023-05-19
Use maximumDateFutureInterval for Carb Entry

marionbarker:
The value for maximumDateFutureInterval is defined as 4 hours. But the usage in picker is hard-coded 1 hour.
Use the defined value in the picker.


marionbarker:
Base code modified. Delete this PR and make a new one.

Issue #2004, opened 2023-05-29
Bug: dev only; Override Active Toolbar Icon inconsistent


marionbarker:
**Describe the bug**
When an override is set, it should show both:
1. Override icon is active indication in the Toolbar 
2. Override name in the HUD Status Row.

Initially, this is true. But with recent dev builds, the active indication in the Toolbar can stop showing.

Not sure exactly when this started.

I reported this on zulipchat in this conversation.
* https://loop.zulipchat.com/#narrow/stream/144182-development/topic/Dev.20updates/near/357016848

In that same zulipchat stream, Damon Bayer said he could reproduce by restarting the app.
* When I did that on a test phone running 3.2.2, the toolbar icon remained set
* Built dev on that same phone with override active, toolbar icon is not set (build is essentially a quit/restart)
    * tapping on the override icon (even though not showing as active) properly disables override
    * set the override again
* Built 3.2.2 back on the same phone with override active, toolbar icon remains active as desired

Now that I know how to reproduce, I can do this on my real phone.
* set an override
    * quit and restart app
    * active icon shows on watch but not phone toolbar
    * disabling from watch works
    
**To Reproduce**
* Restart app
* Unknown - intermittent (not entirely sure if app was restarted)

marionbarker:
Found it but only by trial and error looking at diffs. I have no idea why this fix works.

Comment out one line:
https://github.com/LoopKit/Loop/blob/dev/Loop/View%20Controllers/StatusTableViewController.swift#L167

I'll prepare the PR next.

TLDR - 

Examined the code and I decided the problem must be in:
* Loop/View Controllers/StatusTableViewController.swift

I was not able to find the problem to fix it - so I took a different approach.
I identified the LoopWorkspace commit when this occurred.

These are the 2 relevant Workspace commits:
* after this commit, the icon does not work as expected
   * 201ab79 Changes from Tidepool (#55)
* At this commit level, the icon works as expected
   * 1803e01 Always upload artifacts

I then looked at Loop commits between those two LoopWorkspace commits and examined the diffs for "Loop/View Controllers/StatusTableViewController.swift"

marionbarker:
Merged into dev. Closing this issue.

PR #2005, opened 2023-05-31
restore workout icon active status

marionbarker:
Fix for Issue #2004


Trpl7ca:
I changed the line of code and rebuilt while I had and active override in progress. The blue heart was not illuminated before but after rebuilding it came on. Looks like that may have fixed it.

PR #2008, opened 2023-06-10
Add Integral Retrospective Correction Option to Algorithm Experiments

marionbarker:
The adds the Integral Retrospective Correction (IRC) option to Algorithm Experiments in Loop Settings.
The underlying IRC code is from @dm61 and has been wrapped in the structure to match the configuration used for PR #1988.


marionbarker:
The controls are found on the Loop Settings screen in the Algorithm Experiments section.
The default value is that Integral Retrospective Correct is off.

Loop-Settings-with-IRC-Algorithm-Experiments

marionbarker:
The explanation screen is currently as shown. I will update this comment with a new graphic when this is revised.

IRC-explanation-screen

PR #2013, opened 2023-06-26
Widget Updates

Camji55:
### Medium Widget
Simulator Screenshot - iPhone 14 Pro - 2023-06-26 at 16 38 48

### 3x Lockscreen Widgets
Simulator Screenshot - iPhone 14 Pro - 2023-06-26 at 16 38 52


bfeinzimer:
thank you for doing this, hope this gets approved soon

biesbjerg:
Am I right in assuming this requires Xcode 15 (beta) to build?

image

image

biesbjerg:
While testing on iOS 16 I noticed the lock screen widgets have a background color - guess that's unintentional, right?

IMG_8194

Camji55:
@biesbjerg thank you for the feedback! Yes, it currently requires building with Xcode 15, but I’ll drop that requirement and open a separate Pull Request with the iOS 17/Xcode 15 changes to be merged after Xcode 15 releases.

As for the lockscreen appearance on iOS 16, that’s definitely not ideal — I will look into this.

Thanks again for testing and providing feedback!

ps2:
Thanks Cameron. Let me know when you think this is ready. Excited to see this in Loop!

bfeinzimer:
had a side question for Cameron - 
back on apple watch we can get "x min" to show the last time loop was updated, is that implementable here on the lock screen as well?

having a colored loop icon (red/yellow/green) would have the same effect but I think iOS prevents coloring of the LS widgets...right?

biesbjerg:
@Camji55 My pleasure — Thank you for working on it!

My initial experience is that the widgets show stale/no data quite frequently, which haven't really been an issue previously.

I theorise that it's because of the lock screen widget using up the update budget ([daily budget typically includes 40-70 refreshes](https://developer.apple.com/documentation/widgetkit/keeping-a-widget-up-to-date#Plan-reloads-within-a-budget)).

Without this PR, to check glucose etc, I would typically tap the screen to wake up device, swipe right to reveal the widget, which I guess uses less of the budget than needing to keep the lock screen widget up to date all the time — does that sound plausible?

Camji55:
@bfeinzimer thank you for the suggestion! The problem with having last updated text is finding a place to put it — On the Accessory Inline and Accessory Circular widgets there is no room for this info.

As for the Accessory Rectangular widget, you’re correct that the color of the Loop status circle would not change on the lockscreen. I may however be able to use a broken circle instead of changing the color.

Camji55:
@biesbjerg this is a problem I’ve noticed as well. The 40-70 refreshes is about 4-6x less than the number of new CGM values reported to loop by a standard 5min CGM interval. This is less than ideal and as far as I can tell there isn’t a way around it.

A couple ideas I have to possibly help this scenario though neither are great:
1. Reload the timelimes of the lockscreen widgets separately than the regular homesceen widgets. This will insure that the homescreen widget’s wont waste any of the maximum reloads while they aren’t even shown.
2. Change the reload frequency of the lockscreen widgets from 5mins to 30mins. This will prolong reloads to hopefully the whole day but we lose some of the up-to-date nature.

Happy to hear any suggestions!

Camji55:
@ps2 given some of the feedback I wonder if it’s worth separating the lockscreen widgets and the larger homescreen widget into separate PRs.

ps2:
@Camji55 Yeah, if you want to land the homescreen widget while you see if we can find a solution for the lockscreen widget using up more refresh cycles, that's fine.  Do we end up using more refresh budget with this PR even if the user doesn't add the lockscreen widgets to their lockscreen? If not, then I don't see a problem with landing it, and then continuing to look for improvements.

Camji55:
@ps2 I opened a separate PR for the homescreen widget here: https://github.com/LoopKit/Loop/pull/2027

biesbjerg:
@bfeinzimer I'm not sure why this PR status shows as merged —  https://github.com/LoopKit/Loop/pull/2027 is the one that has been merged (which does not contain lockscreen widgets)

bfeinzimer:
gotcha. still working out the visual bugs or the refresh on the the lock screen widget? eagerly awaiting this :P

IsThisPaul:
> @biesbjerg thank you for the feedback! Yes, it currently requires building with Xcode 15, but I’ll drop that requirement and open a separate Pull Request with the iOS 17/Xcode 15 changes to be merged after Xcode 15 releases.
> 
> As for the lockscreen appearance on iOS 16, that’s definitely not ideal — I will look into this.
> 
> Thanks again for testing and providing feedback!

Looks like it might be time for these updates. I was going to try to tackle it myself but looks like you're aware of it already.

bfeinzimer:
any update for the lock screen widget, esp now with ios 17 out?

bfeinzimer:
> @ps2 given some of the feedback I wonder if it’s worth separating the lockscreen widgets and the larger homescreen widget into separate PRs.

was there ever an update to this PR? thanks, eagerly awaiting this!

Camji55:
@IsThisPaul & @bfeinzimer, I've opened a PR for the Xcode 15 updates to the homescreen widgets https://github.com/LoopKit/Loop/pull/2078.

@bfeinzimer, as for lockscreen widget there still needs to be some experimentation and a larger discussion around safety. We're limited to refreshes by the system to between [40 and 70 per day](https://developer.apple.com/documentation/widgetkit/keeping-a-widget-up-to-date#Plan-reloads-within-a-budget). With a lockscreen widget, this refresh budget gets eaten up very quickly and then the entire rest of the day you're left with expired/invalid data as it can no longer refresh with current data. This problem is not exclusive to the lockscreen widget; the homescreen widget also has the same limited number of refreshes per day which is why it can often be not as up-to-date as the legacy iOS system widgets as noted [here](https://github.com/LoopKit/Loop/pull/2027#issuecomment-1650043529) by @biesbjerg.

bfeinzimer:
thanks for detailing it out…it does make me curious as how the dexcom in house app widgets handle this limitation. any ideas?
  
  
  
  

        
  
  
>

loudestnoise:
@bfeinzimer are you talking about the Dexcom home screen widgets? I don't think they have a a Lock Screen one. In my experience Dexcom's Home Screen widgets have the same issue with stale data.

bfeinzimer:
I agree the data refreshes always seemed stale. How many are allowed on apple watch faces as I "feel" they perform better   
  
  
  
  
  
  
  

  
  
>

loudestnoise:
From [Apple Developer - Complications](https://developer.apple.com/design/human-interface-guidelines/complications)
> You can update the timeline a limited number of times each day, and the system stores a limited number of timeline entries for each app. For developer guidance, see [Keeping your complications up to date](https://developer.apple.com/documentation/clockkit/deprecated_articles_and_symbols/keeping_your_complications_up_to_date).

So a limited number of times ¯\_(ツ)_/¯

PR #2014, opened 2023-06-27
Refactor RetrospectiveCorrection Initialization

bjorkert:
This pull request solves the issue with toggling IRC On/Off without re-starting the app.

I did not want to re-create the RetrospectiveCorrection class for each evaluation or loop, therefore a new computed property retrospectiveCorrection is added in LoopDataManager. This property checks the integralRetrospectiveCorrectionEnabled flag in UserDefaults, and based on its state, it creates an instance of IntegralRetrospectiveCorrection or StandardRetrospectiveCorrection. This instance is then cached and reused unless the flag's value changes. 

Two new private variables, lastIntegralRetrospectiveCorrectionEnabled and cachedRetrospectiveCorrection, have been added to support the caching mechanism.

By removing the instance creation logic from the "Static configuration" block in LoopSettings, it recognizes that this process is, in fact, not static. Moreover, the LoopSettings class adheres to the Equatable protocol, which makes it unsuitable for holding a cache of the RetrospectiveCorrection instance.

### DIFF
--- a/Loop/Managers/LoopDataManager.swift
+++ b/Loop/Managers/LoopDataManager.swift
@@ -433,15 +433,15 @@ final class LoopDataManager {
     }
 
     // Confined to dataAccessQueue
-    private var lastRetrospectiveCorrectionEnabled: Bool?
+    private var lastIntegralRetrospectiveCorrectionEnabled: Bool?
     private var cachedRetrospectiveCorrection: RetrospectiveCorrection?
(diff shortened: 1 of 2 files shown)


marionbarker:
I merged recent dev commits into this branch and proved it builds.
Next I will do more testing that IRC toggle changes state without quit/restart required.

PR #2026, opened 2023-07-18
add effect of suspending insulin delivery

dm61:
Allows users to view the predicted effect of suspending insulin delivery on the Predicted Glucose screen. This is for information purposes only and does not affect dosing in any manner. An immediate use case is to assist users decide on their own if a carb correction might be needed to mitigate a potential low in the future. 

The Precited Glucose screen starts as usual, with only carbs, insulin, momentum, and retrospection enabled. 
IMG_2859

In this example, an eventual glucose below the correction range is predicted. Manually selecting the Suspension of Insulin Delivery shows the effect and, in this case, shows that Loop should be able to easily correct the predicted glucose without the need for any carb corrections. 
IMG_2860

Future (more ambitious) use cases for the effect of suspending insulin delivery may include:
- Assessments of dosing risks, potentially in deciding how large the partial application factor should be;
- Development of an automated carb correction notification feature.

### DIFF
--- a/Loop/Managers/LoopDataManager.swift
+++ b/Loop/Managers/LoopDataManager.swift
@@ -372,6 +372,8 @@ final class LoopDataManager {
     }
 
     private var retrospectiveGlucoseDiscrepanciesSummed: [GlucoseChange]?
+    
+    private var suspendInsulinDeliveryEffect: [GlucoseEffect] = []
 
     fileprivate var predictedGlucose: [PredictedGlucoseValue]? {
         didSet {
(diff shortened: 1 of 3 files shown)


ps2:
Nice! Thanks!

ps2:
Oh, this is a little different than I originally assumed; I was out and about and saw it on my phone quickly. I first assumed it was showing the original/full forecast compared to enabling/disabling various effects.

I do like the idea of being able to explore what full suspension would do in a given situation, but not sure if the forecast details page is the right place. One idea would be to make it a preview for a suspend command, like how we show the effects of a potential bolus visually.

For the use case of giving the user awareness of a likelihood for a need of carbs when bg is trending low, I think that would be yet a different design, and maybe an alert.

For the use case of just wanting to get a quantification of how much "braking power" Loop has at any given time, that seems more like a "Learn" app type feature, or maybe a whole separate area in Loop for exploratory/review/learning lessons.

ps2:
Haha, ignore my (now deleted) comment. Will check this out in the context of the other effects. I do still think this is a good visualization for a number of use cases; just thinking about where it ultimately belongs.

PR #2028, opened 2023-07-20
mitigate irc and momentum stacking

dm61:
This update mitigates the upward stacking of IRC and momentum effects, which may occur when retrospective correction is active and glucose is increasing rapidly, as [reported on Zulip](https://loop.zulipchat.com/#narrow/stream/144182-development/topic/Algorithm.20Experiments/near/376540452).

### DIFF
--- a/Loop/Models/RetrospectiveCorrection/IntegralRetrospectiveCorrection.swift
+++ b/Loop/Models/RetrospectiveCorrection/IntegralRetrospectiveCorrection.swift
@@ -179,7 +179,7 @@ class IntegralRetrospectiveCorrection: RetrospectiveCorrection {
             
             // Overall glucose effect calculated as a sum of propotional, integral and differential effects
             proportionalCorrection = IntegralRetrospectiveCorrection.proportionalGain * currentDiscrepancyValue
-            // Differential effect added only when negative, to avoid upward stacking with momentum, while still mitigating slugeshness of retrospective correction when discrepancies start decreasing
+            // Differential effect added only when negative, to avoid upward stacking with momentum, while still mitigating sluggishness of retrospective correction when discrepancies start decreasing
             if differentialDiscrepancy < 0 {
                 differentialCorrection = IntegralRetrospectiveCorrection.differentialGain * differentialDiscrepancy
             }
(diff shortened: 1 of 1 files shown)


Loop/Models/RetrospectiveCorrection/IntegralRetrospectiveCorrection.swift
```diff
@@ -179,7 +179,10 @@ class IntegralRetrospectiveCorrection: RetrospectiveCorrection {
             
             // Overall glucose effect calculated as a sum of propotional, integral and differential effects
             proportionalCorrection = IntegralRetrospectiveCorrection.proportionalGain * currentDiscrepancyValue
-            differentialCorrection = IntegralRetrospectiveCorrection.differentialGain * differentialDiscrepancy
+            // Differential effect added only when negative, to avoid upward stacking with momentum, while still mitigating slugeshness of retrospective correction when discrepancies start decreasing
```
ps2, line 182: With this change, IRC is still accumulating increasing positive effects as time goes on, right?  It's just eliminating this "gain" on the positive side, right?

Also, small typo: "slugeshness" -> sluggishness.

dm61: Yes, IRC is still accumulating increasing positive effects as time goes on; this is just eliminating an additional positive effect when the most recent discrepancy is higher than the one before. This is likely to occur when glucose is increasing quickly (while there are no carbs on board, so RC effects are present), which is why stacking with momentum can happen. In the opposite direction, this effect is safe and beneficial - reduces the retrospective effect more quickly when discrepancies are decreasing.

dm61:
Corrected the calculation, and have observed operation some more. I think this is a good safety improvement to irc.

Loop/Models/RetrospectiveCorrection/IntegralRetrospectiveCorrection.swift
bedtime4bonzos: @dm61 I'm interested in your thoughts on leaving in the differential term for negative differentialDiscrepancy and what you want it to do, both to understand it and to help evaluate its effectiveness.

dm61: @bedtime4bonzos consider a scenario when discrepancies remain positive for extended periods of time, for example, due to temporarily reduced insulin sensitivity. or some other unmodeled factor pushing glucose upwards. IRC accumulates such effects, which results in a larger (positive) retrospective correction. This is all fine as long as the unmodeled factor persists. However, once the unmodeled factor goes away, for example, insulin sensitivity goes back up, IRC tends to be sluggish as it takes some time for the accumulated discrepancies to decay. That's where the differential term comes into play - reduces the retrospective correction in response to discrepancies that are _decreasing_.  As a result, IRC with the differential term (in the negative direction) is less sluggish (and thereby safer) when the unmodeled effects go away. In the positive direction, the differential term, especially in combination with the. momentum, can result in too large positive corrections, which is why this has been removed in this IRC update.

PR #2054, opened 2023-08-23
Override warning fix

SwiftlyNoah:
After an override is complete, the warning that an override is enabled is still present in the carb entry screen (described in Zulip [here](https://loop.zulipchat.com/#narrow/stream/144182-development/topic/Active.20Override.20warning.20on.20Carb.20Entry)). I was not able to reproduce this issue, but still suspect that this change will fix it, as the new code now matches the [old UIKit code](https://github.com/LoopKit/Loop/blob/befcbcbe6733e52b315cb29904f7893e39da3514/Loop/View%20Controllers/CarbEntryViewController.swift#L242).

### DIFF
--- a/Loop/View Models/CarbEntryViewModel.swift
+++ b/Loop/View Models/CarbEntryViewModel.swift
@@ -290,13 +290,14 @@ final class CarbEntryViewModel: ObservableObject {
     }
     
     private func checkIfOverrideEnabled() {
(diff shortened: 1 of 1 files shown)


billybooth:
I hadn't actually looked at the changes to the `CarbEntryViewModel.swift` yet. This is so much easier to reason about. Nicely done! 👍🏻

PR #2056, opened 2023-08-23
Disable missed meal detection if there are user-entered or calibration BGs

ps2:


ps2:
@novalegra I believe this is updated to dev appropriately; just resolving conflicts and updating to new function name for checking if an array of glucose values contains a calibration.  Would appreciate your review. Thanks!

novalegra:
I reviewed and everything looks good to me! Note that I can't currently build since I'm away from the computer for the next few weeks

Issue #1854, opened 2022-11-30
ShareClient fails to update


marionbarker:
**Describe the bug**
Issue#1: If Dexcom Share is added as a CGM before a Pump is added to Loop, the CGMManager:ShareClientManager does not update after Pump is added.
Issue#2: Walked away from my phone, so DexcomShare stopped providing new readings, Loop stopped updating even after readings resumed with sensor.
* Tested ONLY with a rPi DASH simulator
* I suspect this has been there for a while
   * I remember needing to delete and add back Dexcom Share when testing
    * Not a typical use case

**Expected behavior**
Expect
* to be able to add CGM and pump to Loop in either order
* if dexcom share source is interrupted and then resumes that Loop begin to work without intervention

The intervention required is to delete Dexcom Share as a CGM and then add it back again.

**To Reproduce**
* I built Loop-dev (commit ad4e6b8) to a blank phone, pulled in Therapy from NS, confirmed and saved
   * Added Dexcom Share as CGM
   * Added rPi DASH simulator
   * Left phone in open loop
   * Loop never turned green (waited more than 5 minutes)
* Deleted CGM and added Dexcom Share as CGM
   * Loop turned green immediately

Repeated to confirm, pod insertion completed at 08:39 PST.
Report ShareClientManager statements in debug log up through 08:49 PST.
```
2022-11-30 08:36:58.066920-0800 Loop[1372:280813] [DeviceDataManager] CGMManager:ShareClientManager did update with 2 values
2022-11-30 08:37:06.017885-0800 Loop[1372:280814] [DeviceDataManager] CGMManager:ShareClientManager did update with no data
```

I see the no recent glucose message on the HUD.
Delete CGM and then add it back and get an immediate green loop, and loop stays green
```
2022-11-30 08:51:52.336748-0800 Loop[1372:284698] [DeviceDataManager] CGMManager:ShareClientManager did update with 4 values
2022-11-30 08:52:11.353423-0800 Loop[1372:284777] [DeviceDataManager] CGMManager:ShareClientManager did update with no data
```

Now delete both pump and CGM.
This time add Pump first and then CGM.
Insert completed at 09:23 am PST
Add Dexcom Share as CGM at 09:24 am PST and get an immediate green loop.
```
2022-11-30 09:24:14.159099-0800 Loop[1388:294020] [DeviceDataManager] CGMManager:ShareClientManager did update with 2 values
```

[...]

PR #2058, opened 2023-09-01
Enable pump heartbeat if needed when adding a pump

ps2:
May fix https://github.com/LoopKit/Loop/issues/1854


marionbarker:
Test Scenario:

iPhone (iOS 16.6) add Dexcom Share as CGM and Pump Simulator:
* does not update share with Loop at dev branch (commit 3367e606):
   * phone unlocked and app open (starts green) turns yellow - don't wait for red
* update to Loop branch pump-heartbeat-fix (commit 9b9012f6):
   * Loop updates and turns green on rebuild (but that's the same as a quit/restart)
   * phone unlocked and app open - turns yellow and new share value is not populated
   * Confirmed the new line is in DeviceDataManager.swift

ps2:
This change *only* affects the situation described in the issue: add CGM, then pump, and no reboot, or even no loop-background to foreground.  It only works with a BLE pump. And only with a ble pump with regular BLE events. 

There *was* an issue where we weren't enabling pump BLE heartbeats when the pump was added after the CGM. That's what this fixes.

ps2:
@marionbarker I'm not sure if your test is testing the same problem you described in the ticket, as the test report doesn't say anything about adding devices in a particular order.  If you're having general issues with a lack of heartbeat, that would be a separate issue that we'd need to look into.

marionbarker:
Latest LoopWorkspace (22 Sep 2023), commit ea23351.
* add CGM = dexcom share, add Pump = simulator - Loop went red
* Added the one-line change to DeviceDataManager.swift and rebuilt
   * Loop went green with rebuild
   * Turns yellow at 6 min

If you want a different test - happy to provide one - let me know the desired configuration.

ps2:
This fix, like the situation described in https://github.com/LoopKit/Loop/issues/1854, is a fix for having a non-ble CGM, and then adding a BLE pump that supports a heartbeat.  To test it, you will need to make and install a loop version *with* the fix, and *then* add the pump.

ps2:
The test for this would be:1) build & install loop *with* this change. 2) add non-ble CGM, 2)

marionbarker:
I will be happy to test when I return home in a few days.

marionbarker:
Summary: Tested in the order specified by Pete. 

Results: This fix makes the app behave correctly when non-heartbeat Dexcom Share CGM is added before heartbeat pump.

* Nightscout as a Remote CGM works fine even when added before the pump with heartbeat
* Dexcom Share has the problem before the fix, even with app in foreground
* Add fix and Dexcom Share works as expected

## Test Details
First reproduce the problem, then test the fix:
1. Start with no app on phone, (iPhone 8, iOS 16.6)
2. Build dev (without the fix), commit a7cc83c
    * Onboard
    * Add Nightscout CGM (my sensor is warming up, cannot test with Dexcom Share at this time)
    * Add DASH rPi simulator (has a heartbeat)
    * Wait for green loop
    * Lock phone
    * Observe test nightscout site while phone is locked, loop continues to update
3. Continue with same version of dev, but now my sensor is warmed up
    * Delete CGM and Pump
    * Add Dexcom Share as CGM
    * Add DASH rPi simulator (has a heartbeat)
    * Loop goes yellow even with phone unlocked and app in foreground
4. Modify Loop/Managers/DeviceDataManager.swift with one line change (PR 2058)
    * Delete CGM and Pump
    * Add Dexcom Share as CGM
    * Add DASH rPi simulator (has a heartbeat)
    * Wait for green loop
    * Lock phone
    * Observe test nightscout site while phone is locked, loop continues to update

PR #2090, opened 2023-10-30
Clarify Unable to Reach Pump message

marionbarker:
Propose new wording for the Unable to Reach Pump message.
I believe this one change would help clarify this situation.

The graphic below shows the original wording on the left and the proposed wording on the right.

update-for-unable-to-reach-pump

### DIFF
--- a/Loop/Managers/DeliveryUncertaintyAlertManager.swift
+++ b/Loop/Managers/DeliveryUncertaintyAlertManager.swift
@@ -30,7 +30,7 @@ class DeliveryUncertaintyAlertManager {
         if self.uncertainDeliveryAlert == nil {
             let alert = UIAlertController(
                 title: NSLocalizedString("Unable To Reach Pump", comment: "Title for alert shown when delivery status is uncertain"),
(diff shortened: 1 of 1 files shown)


ps2:
Thanks, I think this is heading in the right direction, but the changes will need to be made on the pump specific views (the pages you go to when you tap Learn More). Also, we probably want to avoid vagueness like "loop hardware", and "usual red loop correction methods".

ps2:
Closing, as this will need to be done at the pumpmanager level.

marionbarker:
I was trying to avoid the 3 pump manager messages - but that was just me being lazy.

I still think this first message needs to be reworded because people assume it's because they walked away from the rileylink (for example). They do not understand Loop was in the middle of something that would affect IOB when the communication issue happened.

I will work with the three Learn More messages; but I might push another commit to this branch for your review.

motinis:
> I still think this first message needs to be reworded because people assume it's because they walked away from the rileylink (for example). They do not understand Loop was in the middle of something that would affect IOB when the communication issue happened.

Yes - I can confirm that this has happened to me (I assumed it's just that the phone was too far away or something) and didn't understand how the current situation is different.

marionbarker:
Associated PR for OmniBLE has been prepared for review. Once a decision is reached on wording here, I'll prepare for the other pumps.

* [OmniBLE PR 104: Clarify wording for Learn More screen when Unable to Reach Pod](https://github.com/LoopKit/OmniBLE/pull/104)

PR #2097, opened 2023-11-21
Fix logic error for TestFlight expiration

marionbarker:
Now that some users of Browser Build method have profile expirations that are shorter than 90 days after the build, we see that the TestFlight expiration continues to be 90 days after build date.

In other words, TestFlight expiration is independent of profile expiration.

Remove the logic that considers profile expiration for a TestFlight build.

This has been tested with a Mac-Xcode build.


marionbarker:
This has been tested successfully using Browser Build.

PR #2125, opened 2024-02-06
Modify maxCarbEntryFutureTime to 4 hours

marionbarker:
Many people use future carb entry to handle fat and protein rise.
I believe increasing this value from 1 hour to 4 hours is a reasonable compromise between safety and convenience.

Note:
* This is a personal opinion
* I use 4 hours when dosing for my evening meal
   * I enter a prompt meal entry and future fat/protein rise entry at the same time
   * I am able to do this because I customized my code

### DIFF
--- a/Loop/Models/LoopConstants.swift
+++ b/Loop/Models/LoopConstants.swift
@@ -25,7 +25,7 @@ enum LoopConstants {
     static let maxCarbAbsorptionTime = TimeInterval(hours: 8)
     
     static let maxCarbEntryPastTime = TimeInterval(hours: (-12))
-    static let maxCarbEntryFutureTime = TimeInterval(hours: 1)
+    static let maxCarbEntryFutureTime = TimeInterval(hours: 4)
 
     static let maxOverrideDurationTime = TimeInterval(hours: 24)
     


ksilive:
Hi, im against this change as its more easy to forget (and maybe make a mistake as a novice user) and can lead to dangerous lows in the night. Lows are more dangerous overall then temp highs.

elnjensen:
@ksilive This change would have no effect on you if you don't enter future carbs. Simply accept the current time stamp for any carb entry and there is no safety issue.

ksilive:
Users will try and experiment with this and I state again that this can be seriously dangerous.

Related point I will make is this, this safety limit I’m annoyed with. This is set way to high at 3,7 / 67 and I run weekly into it. Would I suggest to change this, probably not. I understand children and novice users also use loop. As stated before lows can be very serious and deadly.

scottf51:
I feel like there are a lot more dangerous levers already available than this , like all the insulin settings. If someone is going to randomly start tweaking things I think this is a way down the list. Given, the breadth of abilities of people that use it , do we need an “expert mode” , even if that requires a one line code change (rather than lots of little ones) to enable the more dangerous stuff. Feels like a button with big warning should be enough though TBHJust my $0.02 of course Sent from my iPhoneOn 17 Feb 2024, at 06:43, ksilive ***@***.***> wrote:﻿
Users will try and experiment with this and I state again that this can be seriously dangerous.
Related point I will make is this, this safety limit I’m annoyed with. This is set way to high at 3,7 / 67 and I run weekly into it. Would I suggest to change this, probably not. I understand children and novice users also use loop. As stated before lows can be very serious and deadly.

marionbarker:
Closing because I do not think this will be accepted.

scottf51:
Shame, compared to the other settings this does not seem super dangerous to me and I use this often with fatty foods. Surely things like ISF are much more likely to cause issues than this  Sent from my iPhoneOn 20 Sep 2024, at 21:39, Marion Barker ***@***.***> wrote:﻿
Closing because I do not think this will be accepted.

marionbarker:
close this. I don't think this will ever be accepted. Available with the customization select script

PR #2131, opened 2024-02-13
LOOP-4781 Loop to use LoopAlgorithm swift package

ps2:


sraka1:
@ps2 will this not land in Loop?

ps2:
This is being worked on in Tidepool Loop, but I will merge it into DIY dev after the 3.4 release.

Issue #2135, opened 2024-02-23
Feature request: Make it easier to save Favorite Food from Carb Entry screen


marionbarker:
**Is your feature request related to a problem? Please describe.**

Many people get frustrated when creating Favorite Foods. 
* If they happen to start from the Settings, Favorite Food screen, it is really straightforward
* If they try to create one from the Add Carb Entry screen, they are not sure how to get the "Save as favorite food" button to be active
   * In order to make it active, they must tap on the plate and then select at least one emoji

**Describe the solution you'd like**
The selection of a food emoji should not be required on the Carb Menu to activate that button.

I made a code change and this changed the behavior.

See PR #2136

marionbarker:
Bump

carolinecreamer:
@marionbarker Hi!! I am not sure how to direct message on Git, but I am a computer science student at the University of Michigan. I am currently taking a software engineering course and our final project is to contribute to an open source project on Github. I am a type one diabetic and am hoping to work on diabetes technology in the future, so I was hoping to contribute to the Loop Repo, but wanted to make sure this is okay with you first? I am thinking about working on this feature along with #2127.

marionbarker:
@carolinecreamer 

Here are some links for you to review.
* [Development History](https://loopkit.github.io/loopdocs/intro/overview-intro/#development-history)
* [Loop Development](https://loopkit.github.io/loopdocs/version/development/)
* [What is LoopWorkspace](https://loopkit.github.io/loopdocs/version/loopworkspace/#what-is-loopworkspace)

Please join zulipchat and join all the streams but especially the development stream.
* Use [Zulipchat](https://loop.zulipchat.com/) forums for Loop.

marionbarker:
Fixed in Loop 3.4.x

PR #2136, opened 2024-02-23
make it easier to save a Favorite Food from Carb Entry screen

marionbarker:
This is a proposed solution to Issue #2135


marionbarker:
This was tested. You can now create a Favorite Foods from the carb entry screen without needing to tap the plate icon to enter a special food emoji.

PR #2149, opened 2024-03-28
remove shaking phone as trigger for presentDebugMenu

marionbarker:
It is too easy to trigger the debug menu during daily phone use now that allowDebugFeatures is enabled for DIY loop.
Remove the shaking option, but keep the rotate phone between landscape and portrait 6 times with no more than 2 sec between rotations.

### DIFF
--- a/Loop/View Controllers/StatusTableViewController.swift
+++ b/Loop/View Controllers/StatusTableViewController.swift
@@ -1865,15 +1865,6 @@ final class StatusTableViewController: LoopChartsTableViewController {
         lastOrientation = UIDevice.current.orientation
     }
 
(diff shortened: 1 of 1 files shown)


elnjensen:
Thanks for flagging this.  I'll confirm that I accidentally trigger this almost daily, in normal use, so I'm in favor of making it harder to trigger accidentally.

marionbarker:
I tested this with a test phone.  
* Shake and see the menu every time
* Apply patch, shake and do not see the menu
   * However, what I call the tick-tock approach - rotating my phone back and forth to landscape and portrait 3 times reliably brings up the menu

Trpl7ca:
Thanks hope it gets approved, that has been a nuisance

Trpl7ca:
It worked, thank you

PR #2163, opened 2024-05-02
Use full ICE history for displaying estimated carbohydrate effects

### LINKED ISSUE #2159: Glucose change graph 'predicted' effects are inconsistent at different times or between orientations
There are some inconsistencies in the glucose change graphs on the carbohydrates screen that make it harder to interpret.

I split this into two parts:
1) Inconsistency between views and times - which appears to be a bug affecting the glucose change graph
- dynamic carbohydrate absorption is sometimes shown in the landscape view but not portrait view
- dynamic carbohydrate absorption is sometimes not shown on the left edge of the graph in both views

2) a preference for a static prediction matching the carb entries - which from zulipchat discussion I understand is not expected behaviour, and from reading the code looks like a harder thing to achieve. I guess similar information can be had by reading through each of the carb entries and how absorption is tracking.

**To Reproduce**
On the home screen tap on carbs on board chart to get to the carbohydrates screen.

Look at the 'glucose change' graph, in particular the 'predicted' glucose change plotted on the graph.

Switch to landscape view. Look carefully at the 'glucose change' graph. In this view the grey glucose change graph is often different and almost completely follows the 'observed glucose change' graph - making the whole graph appear in dark green.

Issue is intermittent and can also appear in the portrait view.

**Expected behavior**
I would expect:
1) the predicted glucose change to be the same when shown in portrait and landscape mode

I would prefer:
2) a static prediction to be shown in the 'predicted' dataset - i.e. the 'predicted' glucose change to stay the same over time unless the meal entries are updated.
 From zulipchat discussion I understand this is not expected behaviour at this point in time - the dataset labelled 'predicted' is in fact Loop's current estimate of the absorption rate of each meal up to the current point in time, which may be either minimum rate absorption or dynamic carb absorption. Note that as dynamic carb absorption equates to the observed glucose change, it gives you limited additional information compared to the observed glucose change dataset. You can see dynamic carb absorption is in effect because the graph looks messy, but can't easily see how this compares to an 'ideal'.  A comparison of static prediction with ICE could be quite useful to support reviewing meal entry estimates (size and absorption time).

**Screenshots**
In the screenshot below the grey/dark green shows a static minimum absorption rate prediction derived from the carb entries. Overlaid on top in light green is the observed glucose effect or Insulin Counteraction Effects. 

[BC714572-ED9F-446D-A551-53AB5DF35D3E.png](https://loop.zulipchat.com/user_uploads/8196/EOwD3vnDEcTVtDJdAVAeDz9f/BC714572-ED9F-446D-A551-53AB5DF35D3E.png)

The landscape mode screenshot below, taken seconds later, shows things differently.

Here, the minimum rate absorption estimate does not appear from about 12:20 to 19:38 (the time of the screenshot). It is present in the future. For example, the part of the two charts around 16:00 clearly has different values for the the 'predicted' dataset. 

[89BFB94D-443B-48AE-ABA9-75CD927154D8.png](https://loop.zulipchat.com/user_uploads/8196/8caRHCFC1Gqj2a2aemeoWEFl/89BFB94D-443B-48AE-ABA9-75CD927154D8.png)

**Loop Version**
 - Version Number: v3.2.3
 - Repo: Loopkit/Loop
- issue also appears in dev although the code has changed

**CGM**
 - Device: Dexcom G6
 - Manager app: Dexcom App

**Pump**
 - Manufacturer: Omnipod Dash

**Additional context**
Discussion on Zulipchat: https://loop.zulipchat.com/#narrow/stream/144111-general/topic/ICE.20.2F.20glucose.20change.20graph.20display.20inconsistencies

Looks like this graph was originally added along with dynamic carb absorption: https://github.com/LoopKit/Loop/pull/507
But has been modified since, including when non-linear carb absorption (with dynamic effects) was added in Loop 2.0 . This screen is described for that and subsequent versions here: https://loopkit.github.io/loopdocs/operation/features/ice/#glucose-change-display

There are slightly different explanations of what is supposed to be shown on the graph on the two pages linked above.

One describes the 'predicted' dataset as being derived from the meal entries, which I had interpreted as a static prediction which is then compared with the observed glucose change. Elsewhere e.g. on the Open Loop documentation page the graph as a whole is described as displaying 'how the app thinks your meal is being absorbed'.

oliverstory:
The 'predicted' dataset on the 'glucose change' graph was calculated using insulin counteraction effects that had been truncated to the width of the chart. This meant that carb entries early in the timeline may not have dynamic carb absorption applied as the implied ICE was zero for the first part of the carb entry's history.

Addresses the different behaviour in landscape and portrait view of the glucose change graph in the carbohydrate screen. so that dynamic carb effects are always shown when they are in effect.

For discussion (but not included in this pull request), for clarity the graph dataset label could be changed from 'Predicted' to 'Estimated and predicted' given that it already incorporates dynamic glucose effects (when these are in effect), so is partly based on observation.

Does not address what I described as 'desired' behaviour of showing a static prediction for the 'predicted' dataset; I am not sure how you would achieve that.

Closes #2159

### DIFF
--- a/Loop/View Controllers/CarbAbsorptionViewController.swift
+++ b/Loop/View Controllers/CarbAbsorptionViewController.swift
@@ -173,7 +173,7 @@ final class CarbAbsorptionViewController: LoopChartsTableViewController, Identif
                 }
 
                 reloadGroup.enter()
-                self.deviceManager.carbStore.getGlucoseEffects(start: chartStartDate, end: nil, effectVelocities: insulinCounteractionEffects!) { (result) in
+                self.deviceManager.carbStore.getGlucoseEffects(start: chartStartDate, end: nil, effectVelocities: allInsulinCounteractionEffects) { (result) in
                     switch result {
                     case .success((_, let effects)):
                         carbEffects = effects


marionbarker:
I cannot reproduce the issue reported in #2159 with my test phone.
I can apply the patch to dev, commit 47450c1.
It applies cleanly, builds with no problem and the ICE screens are unchanged.

If there's an additional test that would be helpful, I could try again.
(I ran for about 6 hours with 3 carb entries.)
Do you know of specific configuration I could try to reproduce to make the inconsistency show up?

oliverstory:
Based on my understanding of the maths, to reproduce the issue prior to the patch you need:
1. A carb entry where ICE is slightly larger than minimum rate absorption but not hugely larger.
2.  To set this up on a test phone you could try adding  an insulin entry and have glucose flat. Add a matching carb entry at a similar time.   You could enter this at a time close to the earliest time on the portrait view glucose change graph to make subsequent steps shorter. 
3. Carb entry should initially show dynamic carb absorption - ie ‘predicted’ curve will follow ‘observed’ curve - on both graphs. 
3. Wait until the start time of the carb entry moves off the left of the portrait view graph. At a certain point, the sum of minimum rate absorption over time for the entry will be more than the sum of ICE (because prior to the patch, ICE is truncated at the left edge of the graph). Now the predicted curve will follow minimum rate absorption and be below ICE on the portrait view. But on landscape view, the it should still show dynamic carb absorption (because the landscape view uses a longer series of ICE values).

Trpl7ca:
I would like to try this, if I use your commt number and git hub info and add line to my yum file, would that work for browser build?

oliverstory:
Yes it would work (for dev branch, not for main branch). You can see the
lines you need in my browser build yum file.

itsmojo (approved):
Did not try to reproduce, but this simple change seems reasonable and safe to do with my limited understanding of this code.

oliverstory:
This code will change with the tidepool-merge (and I _think_ this display bug may not exist in that branch, from reading the code - I have not tested tidepool-merge)

ps2 (approved):
Yes, this is already fixed in the tidepool merge. We can merge this into dev for now, since it looks like it should fix the issue there, and I'll just have to resolve the conflicts before tidepool-merge lands.

PR #2191, opened 2024-07-15
feature: Live Activity

bastiaanv:
This PR is an initial/basic version of this Live Activity, without any feature which are spoken of in the Zulipchat.

### Lock Screen (all iPhone's):
| History only                                                                                 | History and predictive line                                                                  | Live activity stale (8h limit i.e.)                                                                 |
|----------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------|
| IMG_2536 | IMG_2534 | IMG_2535 |

### Dynamic Island (iPhone 14 pro max and higher):

> Screenshots are from a iPhone 15 pro max simulator

| Dynamic Island minimal                                                                                                                                | Dynamic Island expanded                                                                                                                               |
|-------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------|
| Simulator Screenshot - iPhone 15 Pro Max - 2024-07-15 at 12 49 36 | Simulator Screenshot - iPhone 15 Pro Max - 2024-07-15 at 12 51 14 |

### Settings

> Can be found at: Settings -> Alert management -> Live Activity

| Root settings                                                                                | Bottom row settings                                                                          | Available datapoints for bottom row                                                          |
|----------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------|
| IMG_2537 | IMG_2538 | IMG_2539 |

## Things to know:

- The bottom row has a max of 4 items, but less items can be configured for small screen sizes (like the iPhone 13 mini)
- The bottom row can be configured to the user's wishes. Swap items around, different order etc. The default is: `CurrentBG`, `IOB`, `COB`, `UpdatedAt` (like in the screenshots)
- The following events trigger a re-render of the Live Activity:
  - LoopDataManager -> `GlucoseStore.glucoseSamplesDidChange`
  - LoopDataManager -> `CarbStore.carbEntriesDidChange`
  - LoopDataManager -> `dosing changing`
  - NotificationCenter -> `UIApplication.willEnterForegroundNotification`


A9qx:
Love the draft/prototype! Would love to see this make it into the project down the line I think it looks amazing!

marionbarker:
Built onto my real phone. Thanks so much for making this available. 👍

bastienmoulia:
Awesome work !
Is it possible to change the y-axis of the graph in order to be more center on the data like in the app?
image
image

marionbarker:
I have been using this for several days - I really enjoy having it attached to Loop.

### Bug report:
* Dynamic Island expanded (long press on dynamic island)
   - truncates the display when glucose is in mg/dL and above 99 mg/dL
   - graphic below with 24 hour display disabled (top) and enabled (bottom)

loop-pr-2191-dynamic-island-expanded

### Feature requests: 
* see the target range shown on the Look Screen Widget graph
* see if an override is currently enabled, definitely added to target display on graph, possibly with the override name displayed

marionbarker:
Nice upgrade except the override appears confused.

* This override was active from 11:05 to 1:05 pm (with 2 hours duration), but it shows as being currently active (1:45 pm).
* I see the same pattern when long-pressing the dynamic island.

old-override-shows

* I added a new override at 1:47 pm that has a duration of 3 hours.
* I will report back in 20 minutes or so. Long enough to see if the start point is correct
* And then I will turn it off and see what the plots look like.

marionbarker:
The walking override started at 1:47 with a duration of 3 hours and was cancelled 19 minutes later.
* upper plot: While override is active. The bad news is the start point for the plot for both lock screen and dynamic island expanded plot is "now" instead of when override started.
* lower plot: After disabling the override. The good news is that plotting a cancelled override was not repeated - perhaps that was related to the new build, or perhaps because it was cancelled instead of allowed to run the full duration.

override-status

bastienmoulia:
Thanks for the blue area!

Same here for the bug with an override label shown even none were triggered
IMG_B15E4FD11EDD-1

marionbarker:
I did a bit more testing. 
* If an override is active, it is shown as starting "now"
* If an override expires after the planned duration, it continues to be show starting "now"
* If an override is cancelled early, no extraneous override is displayed

bastienmoulia:
In the last update, I didn't see the target area for the next hour by default:
IMG_598D1AB784AD-1

If I activate the premeal, then remove it, it's ok.

bastiaanv:
Aight, it has been a crazy few days, but I made several fixes and restructured the code a bit. This should fix most of the weird behaviour found by @bastienmoulia and @marionbarker.

@marionbarker regarding the extraneous override. With this new version, I made it so that the preMeal override is "preferred" above the scheduled overrides, since that is only active for a maximum between now and 1hour. Do you agree with this? I dont think there is enough space to show both texts, tho it is possible to show both in the graph

bastiaanv:
Other question, do we want to keep the upper and lower limit in the settings (and with that the green, orange and red coloring in the graph)?

bastienmoulia:
> Other question, do we want to keep the upper and lower limit in the settings (and with that the green, orange and red coloring in the graph)?

Personally I prefer to not have multiple color because it causing more stress when it's orange or red. So no limit.

marionbarker:
I have no strong opinion on the color. But with the ability to modify the glucose levels at which colors change allows people who want colors to have them and folks who do not want colors to turn them off.

I just set mine to 40 and 400 mg/dL to give it a try.

bastiaanv:
I added the setting to enable or disable the colouring based on the BG

| Chart view without colouring                                                                                                                          | New setting (with BG colouring enabled)                                                                                                                                          |
|-------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------|
| Simulator Screenshot - iPhone 15 Pro Max - 2024-08-01 at 20 43 44 | Simulator Screenshot - iPhone 15 Pro Max - 2024-08-01 at 20 45 36 |

marionbarker:
## Bug reports:
1. I tested with the latest and I am still having the Live Activity show overrides that were allowed to end at the planned duration. It does clear by setting and unsetting any override on the Loop app.
2. Adjust the values for BG coloring, there does not appear to be a way to save these values

## Comments
I would prefer if there was an explicit Save button to save new settings on the Live Activities screen with the Back button leaving settings alone.

## Successes
The things that work as expected:
* Choose Large or Small
* Available for Large or Small:
   * Add / Delete / Rearrange items for bottom row
* Available for Large only:
   * Enable / Disable Prediction Line
   * Enable / Disable Use BG coloring
   * Adjust the values for the BG coloring (see bug report)

rfmurphy81:
Dumb question, but could anyone point me to documentation that could show me how to try out this branch? Downloading the ZIP from https://github.com/bastiaanv/Loop/tree/feat/live-activity doesn't seem to be buildable in Xcode but perhaps I am missing a step.

bastienmoulia:
> Dumb question, but could anyone point me to documentation that could show me how to try out this branch? Downloading the ZIP from https://github.com/bastiaanv/Loop/tree/feat/live-activity doesn't seem to be buildable in Xcode but perhaps I am missing a step.

I don't know how to do this with Xcode, but with the build from github I change the .gitmodules file like this :
https://github.com/bastienmoulia/LoopWorkspace/blob/dev/.gitmodules
And when there is an update I run `git submodule update --remote` in the LoopWorkspace before pushing it.

bastiaanv:
There are several ways to do this:
1. Open the terminal in the Loop folder. Do 'git remote set-url origin https//github.com/bastiaanv/loop', then do 'git pull', then do 'git checkout feat/live-activity'

2. Or download the patch file from https://github.com/loopandlearn/customization, open the terminal in the Loop folder, then do 'git apply <path-to-patch-file>

Please note, that I typed this on my phone, so there might be typos in my commands

partcoffee:
## Bug report
When using Nightscout as CGM, it appears that the data will always be reported as stale. For additional context on this issue, Nightscout is reading from LibreLinkUp, and thus reporting blood glucose values on a 1-minute interval, which may affect how this is handled.

Using v3.4.1 with the [latest customization](https://github.com/loopandlearn/customization/pull/41) as of today.

bastiaanv:
Hi @partcoffee ,

This behaviour is correct. The Live Activity doesnt keep iOS from suspending the app (also called Heartbeath). This also means that Loop isn't doing anything while the app is suspended. Jumping to conclusion, I assume you use Loop as a following app (just watching the BG values from NightScout).

There are a few ways to correct this:
1. Connect a BLE-enabled pump to Loop
2. Customize Loop to include background sounds (like [LoopFollow](https://github.com/loopandlearn/LoopFollow/blob/main/LoopFollow/Helpers/BackgroundTaskAudio.swift)) or use Location services to keep Loop from being suspended
3. Or switch to something like [xDrip4ios](https://github.com/JohanDegraeve/xdripswift) which is a dedicated app to follow BG values from different sources like [NightScout](https://xdrip4ios.readthedocs.io/en/latest/connect/follower/) or [LibreLinkUp](https://github.com/JohanDegraeve/xdripswift/tree/master/xdrip/Managers/LibreLinkUp)

partcoffee:
> Hi @partcoffee ,
> 
> 
> 
> This behaviour is correct. The Live Activity doesnt keep iOS from suspending the app (also called Heartbeath). This also means that Loop isn't doing anything while the app is suspended. Jumping to conclusion, I assume you use Loop as a following app (just watching the BG values from NightScout).
> 
> 
> 
> There are a few ways to correct this:
> 
> 1. Connect a BLE-enabled pump to Loop
> 
> 2. Customize Loop to include background sounds (like [LoopFollow](https://github.com/loopandlearn/LoopFollow/blob/main/LoopFollow/Helpers/BackgroundTaskAudio.swift)) or use Location services to keep Loop from being suspended
> 
> 3. Or switch to something like [xDrip4ios](https://github.com/JohanDegraeve/xdripswift) which is a dedicated app to follow BG values from different sources like [NightScout](https://xdrip4ios.readthedocs.io/en/latest/connect/follower/) or [LibreLinkUp](https://github.com/JohanDegraeve/xdripswift/tree/master/xdrip/Managers/LibreLinkUp)

Thanks for the clarification, but I'm using Nightscout as a CGM with Omnipod Dash, which functions as a heartbeat as well from my understanding. Loop is running its normal closed loop activities in the background like bolusing and suspending basal based on the readings from Nightscout.

My CGM is not compatible unfortunately, which is why Nightscout is being used.

The Live Activity is reporting the data as stale immediately, and not after a few minutes, when the app in theory still would be "awake".

bastiaanv:
@partcoffee My bad. Thank you for the clarification!

I think you are spot on with the 1 minute time interval. The LA has a limit for the amount of data I can send to the widget, too much data leads to the Widget always being stale. So I can limit the amount of samples being send to the widget even further, but that would limit the usages of the LA. You could try to limit the glucoseSamples yourself for the LA maybe to 70: https://github.com/bastiaanv/Loop/blob/feat/live-activity/Loop/Managers/Live%20Activity/GlucoseActivityManager.swift#L183

But the best solution would be to limit the amount of Glucose samples in Nightscout. Maybe the import script you are running allows for something like that?

bastiaanv:
@marionbarker I fixed your reported bugs! I also made that when you open Loop and the LA isn't showing but is activated, that Loop will recreate the LA, without the need to re-enable LA or kill Loop

partcoffee:
@bastiaanv Thanks for looking in to it! I'll do some tinkering with it myself, I know it's a very niche issue.

I think the best approach would be to alter Nightscout, as you suggested. Loop runs a lot better with 5 minute intervals.

bastiaanv:
Good luck! If you need help, you can always contact me via DM on Zulipchat

bastiaanv:
@marionbarker Update the PR to include a save button and minor bg coloring fix

marionbarker:
Not published yet in CustomizationSelect. 
Testing prior to publishing.

## Test result 1
The stale override that expired at end of duration is fixed.

## Save is almost ok
The save button is there but does not seem to work quite right.
1. If I go to the Live Activities screen
   * make a modification 
   * save it and then use the Back button and then return to the Live Activities screen - it shows previous value (before I hit save)
2. If I subsequently view the LA widget on the Lock Screen, the widget does show the update
3. If I then go back to the Live Activities display, the modifications shows as a requested in Step 1

I rebooted the phone and got the same behavior.

rfmurphy81:
@bastiaanv, thanks so much for putting this together. I put this together last night and had a few notes that I was hoping to pass along for you to consider incorporating into the next commit:

1. On the lock screen, the lime green color of the current glucose reading is difficult to read on my light blue background during the daytime (using the "weather" background). See attached. The contrast needs to be sharper. If it were black or a darker color, it would be easier to read. Is this something that is fixable on your end?
IMG_0482
2. Also on the lock screen, I'd like to see the "eventual glucose" number somewhere as I use that reference.
3. Also on the lock screen, is it possible to incorporate the basal display somewhere? I'd like to know whether I am getting my scheduled basal or if Loop has modified it. 
4. In the Settings > Notifications > Live activity screen, the Save button doesn't actually do anything so I am unable to actually change any of the default values on this screen. 
I built this last night with the code.

And for those who are not as technically savvy as me, I followed the instructions at https://www.loopandlearn.org/custom-code/ which easily allowed me to pull in a set list of customizations, including this one. :)

bastiaanv:
@rfmurphy81 First of all, thank you for your feedback! Really appreciate it.

I will try to find some time to fix the save button soon!

About point 1: The only option currently is to disable BG colouring, but since the save button doesnt work correctly it is hard to try that option. Other idea: would less opacity in the background be a solution for you? Because I think changing the color would be a solution for your background color, but not for someone else. Having less opacity in the widget would leave the widget less vulnerable to this issue.

About point 2: I found the code for the y-axis calculation for the Loop chart, will use that code as soon as I try to fix the save button!

bastiaanv:
@marionbarker I updated the LA with the following:
- Fix for the Save button
- Removed the auto-scaling and ported the Loop yAxis generator

Do you have time for a update of the customisation?

marionbarker:
The most recent updates to Live Activity have been incorporated in the customization script. Run the script again and you will be offered an update to Live Activity if you already have it installed.

```
/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/loopandlearn/lnl-scripts/main/CustomizationSelect.sh)"
```

Documentation overview for the script: https://www.loopandlearn.org/custom-code/#overview

cachrisman:
Just wanted to share something I just noticed - that with the live activity running, if my phone is in StandBy (mine is on a MagSafe desk charger, it shows my BG as a small bubble in the upper right of the clock and I can tap that and see the live activity full screen. Pretty neat!!

Keep up the good work everyone! Thank you for all you do!

IMG_0108
IMG_0109

bastiaanv:
## Progress update:
Allow users to always show the LA or to show the live activity on specific triggers (for now: High or low prediction)

| Settings                                                                                     | Live Activity trigger warning                                                                |
|----------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------|
| IMG_0031 | IMG_0030 |

marionbarker:
Sorry to be slow testing this.
I configured it on an SE.
I modified the Live Activity setting to disable the Always Enabled, and disabled the Show when Low is predicted, enabled the Show when High only.
However, the LA on Lock screen still showed up (there was a low at the time.)

Also, IMO, the Save button should be inactive unless there is a change to save. Then when it is tapped to save a change, the Save button should be inactive again.

marionbarker:
I did more testing - this is with an SE so I can only look at the lock screen - there is no dynamic island. 

When `Always enabled` is off and glucose is between low and high settings, the lock screen shows the note on the left of the graphic below. (Tapping to update the widget does nothing unless you modify the setting to enable the `Always enabled` slider.)

Also, the two sliders for `Show when low is predicted` and `Show when high is predicted` have no effect. The live activity on the lock screen shows when glucose is below low or above high values regardless of whether those sliders are selected or not.

Live_Activity_pr2191

bastiaanv:
Reverted always on feature, since Apple doesn't allow a live activity to be created in the background. Only in the foreground or via push notifications.

Also added the change to have the save button disabled while there are no changes

marionbarker:
## Summary:

* I have the latest version of this PR on my Looping phone. (commit `cc8ee11`).
* I have been using it since last summer.
* I strongly recommend this be added to Loop as a feature; but there are some issues that could be addressed now or later.
* This feature is enabled by entering Loop, Alert Management and using the new row labeled Live Activity.

## Tests:

I have not tested this in a while. It seems to have the same issue we see in other features where you make a change, and the change is actually implemented, but it doesn't show up on the view. By going to another screen (or quit/restart the app) you can force the refresh for that screen with current values.

This happens for some Therapy Settings and Algorithm Experiment selections as well. It appears to be new with iOS 18.

This is an Issue reported for one display glitch and a PR proposed for another one:
* Issue #2267 
* PR #2279 

I think we need a uniform fix for this for iOS 18.

### Screens can show old settings

I did some testing with the screen where you can choose which 4 blocks you want on the bottom row on the lockscreen. The changes showed up on the Live Activity, but if I returned to the Bottom Row design in Loop, they showed the old selection. A quit/restart updated the view so my current selections were shown.

### Save Button always active

The save button is still always active whether a change was made or not.

### Features worked

Except for the display glitches mention above, all features and changes worked as expected.

### Graphics of Controls

Some of the controls are shown on the screenshots below.

pr-2191-controls

### Graphics on Lockscreen

The screenshots below show the lock screen
* left side is when Mode: Small is selected (only shows the blocks selected for the Bottom Row)
* right side when Mode: Large  is selected with prediction lines enabled.

pr-2191-lockscreen

marionbarker (approved):
I approve this PR, but acknowledge it is subject to some UI glitches on some of the control views.

In practice, most people will configure the Live Activity layout one time and leave it alone. I have not modified the layout for mine since I last tested this PR.

However, if the UI glitches can be fixed easily, they should be.

mns6754:
Hi,
I didn’t build my own loop app, installed it through my health care provider (BC Diabetes). I’m not really a tech savvy; however, have a good handle on the loop app features (I think). I’m using iPhone 12 mini. I couldn’t find the Live Activity option in Alert Management. Is there any way to get this feature on my loop app/ phone even if I haven’t built my own loop app? After upgrading to IOS 18, the loop widget has become useless.

bastiaanv:
Hi @mns6754 ,

I am afraid not. Your health care provider needs to changes their Loop build to include LA.

Docs: https://www.loopandlearn.org/loop-features-in-development/#live-activity

mns6754:
> Hi @mns6754 ,
> 
> I am afraid not. Your health care provider needs to changes their Loop build to include LA.
> 
> Docs: https://www.loopandlearn.org/loop-features-in-development/#live-activity

Thank you, Bastiaan, for the clarification

marionbarker:
## Status

Live Activity controls work as expected.
LGTM.

## Test

Test again with the latest updates. Focus on the whether the configuration displays the just-selected selections after a change. (Before they were enacted, but the screen did not update right away.)

### Configuration

IPhone SE 3rd gen, running iOS 18.3.2
* LoopWorkspace dev, commit `3d48a5c`
* update Loop branch from this PR plus commit from PR 2298 (fix error accidentally introduced with PR 2295)

### Test Save Button and Configuration Display

The Save button acts appropriately again (see this [comment](https://github.com/LoopKit/Loop/pull/2191#issuecomment-2741968011) for details). This behavior is related to iOS 18 - previous tests were with iOS 17.

Initially, the Live Activity screen Save button was inactive
* Modify the Live Activity Mode from Large to Small, Save button becomes active
* Tap save and save button becomes inactive
* Tap back and return, Mode of Small is correctly displayed
* Modify the Live Activity Mode from Small to Large, Save button becomes active
* Tap back (without saving) and then return
   * Success: Mode still shows as Small and the Save button is inactive

### Comprehensive Tests for Save Button and Configuration Display

Tests below include looking at the Live Activity on the Lock Screen. For "success", the LA must match the displayed configuration.

Repeat test for all rows on the Live Activity screen:
* success: Enabled
* success: Mode
* success: Add predictive line (Large mode only)
* success: Use BG coloring
* success: Upper limit
* success: Lower limit

### Test Bottom Row configuration screen

Tests below include looking at the Live Activity on the Lock Screen. For "success", the LA must match the displayed configuration.
* success: delete one item, save button active - back and return, no change
* success: delete one item, save button active- save, back and return, change is displayed on screen
* success: add one item, save button active - back and return, no change
* success: add one item, save button active- save, back and return, change is displayed on screen
* success: rearrange order, save button active - back and return, no change
* success: rearrange order, save button active- save, back and return, change is displayed on screen

Loop.xcodeproj/project.pbxproj
```diff
@@ -4832,6 +4921,7 @@
 				INFOPLIST_FILE = "Loop Widget Extension/Bootstrap/Info.plist";
 				INFOPLIST_KEY_CFBundleDisplayName = "Loop Widgets";
 				INFOPLIST_KEY_NSHumanReadableCopyright = "Copyright © 2022 LoopKit Authors. All rights reserved.";
+				IPHONEOS_DEPLOYMENT_TARGET = 16.2;
```
marionbarker, line 4924: Please remove this change (found 8 times) from the pbxproj file.
Instead update `Loop.xcconfig` and change this section:

from:
```
// Min iOS Version [DEFAULT]                                    
IPHONEOS_DEPLOYMENT_TARGET = 15.1
```

to:
```
// Min iOS Version [DEFAULT]                                    
IPHONEOS_DEPLOYMENT_TARGET = 16.2
```

marionbarker (changes requested):
I commented on the need to use Loop.xcconfig to change IPHONEOS_DEPLOYMENT_TARGET. Withdrawing approval until that is updated.

bastiaanv:
@marionbarker Change request is pushed!

marionbarker:
## Resolution Needed

This code does not build as is because of 2 problems.

Once I resolved both of these in my local copy of your feat/live-activity branch, the code built and live-activity is displayed on my test phone (SE 3rd generation).

1. I made a mistake and merged a fix meant for tidepool-merge branch into dev for the Bolus Progress display glitch. I later fixed it with another PR that merged to dev at a later time - you picked up the first half of the fix (PR 2295) and not the second when you merged dev into your branch. Please update with latest Loop dev that includes PR 2298.

2. Turns out a couple of hard coded values for IPHONEOS_DEPLOYMENT_TARGET are in the `Loop.xcodeproj/project.pbxproj` file (blame says about 2 years ago).

This is from dev branch, and those lines are also there in feat/live-activity branch (at lines 5085 and 5195).  
```
Loop % grep -n IPHONEOS_DEPLOYMENT_TARGET Loop.xcodeproj/project.pbxproj
4996:				IPHONEOS_DEPLOYMENT_TARGET = 15.1;
5106:				IPHONEOS_DEPLOYMENT_TARGET = 15.1;
```

The lines with hard-coded values in `Loop.xcodeproj/project.pbxproj` need to be removed for the value of `IPHONEOS_DEPLOYMENT_TARGET ` in `Loop.xcconfig` to be recognized. But since that change came in with a tidepool sync in 2023 [Loop PR 1991](https://github.com/LoopKit/Loop/pull/1991) - let's find out if @ps2 wants a separate PR to remove those, or have that removal come in with this feature branch.

ps2:
It's ok to remove IPHONEOS_DEPLOYMENT_TARGET settings in this PR.

bastiaanv:
@marionbarker fixed! branch is up-to-date with dev and removed the extra `IPHONEOS_DEPLOYMENT_TARGET`

marionbarker (approved):
## Status

All good. LGTM.

## Test

Begin with LoopWorkspace dev clone.
In Loop, switch to this branch and rebuild.
Build works as expected.

When built on a phone that supports Dynamic Island, the glucose shows up in the Dynamic Island when Live Activity is enabled.

The Live Activity, on the Lock Screen, works for
* SE 3rd running iOS 18.3.2
* iPhone 8 running iOS 16.7.10

ps2:
Colors should match the Loop standard colors; glucose should not be orange, for example.  The glucose graph colors should match the normal Loop graph colors if possible. The weight of the glucose points should be adjusted for the more dense display (right now all the green dots run into each other), and the forecast should be thin long dashed lines.

We generally try to avoid jargon like "IOB" and "COB" in the app. I know it might take some redesign to fit it, but we've tried to maintain this, even on the tiny watch interface.

IOB -> Active Insulin
COB -> Active Carbohydrates, or Active Carbs

Thanks!

mns6754:
Thank you, Bastiaan and Marion, for developing and adding the Live Activity feature to the Loop. It’s a great addition to the app. 
Personally, everything is good for what I need from this feature, but I thought adding the pump status (reservoir fill status) could be useful, as graphic or in digits format next to the glucose graph on the top section, and/or to the list of “Bottom Row Configuration” options. I think for people like me who use only three items in the bottom row to get a full reading of the Current Glucose number (and to keep the bottom row tidy to quickly see what is important, and with just a glance), the preferred location would be in top section next to the glucose graph, if possible. Thanks,

mns6754:
In reference to my comment above, these pictures are for comparing the options of having three or four items in the bottom row. Personally, I find three items to be optimal in this current view design. 

IMG_7306
IMG_7331

marionbarker:
Reporting a display glitch when using 24 hour clock. Note the time stamp (for 13:00), shows a 1.

pr2191-display-glitch-24-hour

elnjensen:
Hi @ps2 - A question as I'm working on this: 

> Colors should match the Loop standard colors; glucose should not be orange, for example. The glucose graph colors should match the normal Loop graph colors if possible.

Do you object to different glucose colors being an *option*, but not the default?  I've updated the code to use the standard Loop blue for the glucose points and prediction line (which also now match the main graph style better), but Bastiaan's code exposes a user option for color-coding the glucose points based on whether they are high, low, or in-range. 

A possible middle ground would be to retain that color coding only for the display of the current BG number below the graph, but not the graph points themselves.  That could be handy for glanceability, including in the CarPlay display when driving, and on the Watch Smart Stack.  There's no graph in those views, but seeing the color of the BG number out of the corner of your eye while driving would tell you whether or not you're in range, even if you don't read the number itself. 

I'm working on the other issues you flagged, and hope to share something soon.  @bastiaanv, is it simplest for me to PR possible changes against your branch?  Then if you merge, changes would automatically feed into this PR?   Or let me know if another route is preferable.

bastiaanv:
@elnjensen A PR against my branch would be the simplest flow yes. I will resolve any merge conflicts that are currently present on this PR afterwards

bastiaanv:
Update by @elnjensen!

| Lock-screen                                                                                                              | WatchOS & Carplay                                                                                                                                    |
|--------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------|
| IMG_2678 | Main-VideoStream-2025-11-01-17-36-49 |


Changes:
- Synced Loop dev to this PR
- Processed feedback from @ps2  ([link](https://github.com/LoopKit/Loop/pull/2191#issuecomment-2968807275))
- Added custom view for WatchOS & Carplay

marionbarker:
Thanks for the updates @elnjensen and @bastiaanv.
LGTM

## Test

This feature has been used as part of a customization for more than a year.
The requests from @ps2 were addressed.
I did not test with CarPlay, but I did test on my personal phone with my watch.

**NOTE: the new iOS minimum supported is iOS 16.**

### Test details

started with LoopWorkspace `update_to_dev_3.9.3` branch to see if we can add this feature to that update.
locally merged Loop `dev` branch into the `feat/live-activities` commit 4310a3bc
Built with no issues onto these physical phones:
* SE 2nd gen phone running iOS 18.7.2
* SE 3rd gen phone running iOS 26.1
* iPhone 15 running iOS 18.7.2
* iPhone 8 running iOS 16.7.12

The graphic below shows
* left side is the version before @elnjensen made his updates
* right side shows the updated display

pr-2191-updates

The graphic below shows
* left side is the default BG coloring
* right side shows `Use BG coloring` enabled with the high limit set to 120 mg/dL
   * the color is displayed in the dynamic island and on the watch 

pr-2191-bg-color

PR #2224, opened 2024-09-10
Swedish translations

jakalm:
I noticed that there were a lot of missing Swedish translations in general for the Loop app. This PR addresses this in the Loop repository.

- Added missing Swedish translations in the respective `Localizable.strings` files.
- Added `NSLocalizedString` calls where needed, which will help with other translations as well.
- Removed outdated translations that are no longer needed.


marionbarker:
Closing because this will soon be fixed with this change to Loop: c3c15e1d followed by a cycle of uploading and downloading from lokalise.

PR #2241, opened 2024-10-09
Adjust provisioning profile path in script for Xcode 16 compatibility

bjorkert:
Summary:
	•	Reverted Common/Models/BuildDetails.swift to its state before the “xcode16_profile_support” merge.
	•	Updated Scripts/capture-build-details.sh to search for .mobileprovision files in both Xcode 15.4 and Xcode 16 locations.

Background:
Loop expects a provisioning profile expiration date even for browser builds. However, this expiration date is not available at runtime for TestFlight-deployed apps and therefore the suggested solution in "xcode16_profile_support" did not work.

To resolve this, we have reverted BuildDetails.swift and updated the .sh script to correctly locate the provisioning profile for both Xcode 15.4 and Xcode 16 without requiring changes to Loop’s runtime behavior. This avoids the need for refactoring Loop to handle missing expiration dates differently.

Testing:
	•	Verified with Mac Xcode 15.4 build
	•	Verified with Mac Xcode 16 build
	•	Verified with Browser Build/TestFlight-deployment

### DIFF
--- a/Common/Models/BuildDetails.swift
+++ b/Common/Models/BuildDetails.swift
@@ -13,18 +13,16 @@ class BuildDetails {
     static var `default` = BuildDetails()
 
     let dict: [String: Any]
-    private var cachedProfileExpirationDate: Date?
 
     init() {
         guard let url = Bundle.main.url(forResource: "BuildDetails", withExtension: ".plist"),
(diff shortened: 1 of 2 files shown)


ps2 (approved):
LGTM!

PR #2272, opened 2024-12-11
tidepool-merge: force unwrap to remove Optional in front of two markdown headers

marionbarker:
## Purpose:

Two of the headers in the Loop Report are printed as optionals, e.g.:

`Optional(## WatchDataManager`

`Optional(## StatusExtensionDataManager`

This messes up the syntax and makes it difficult to automatically parse the report.

## Approach:

Add a force unwrap to the two lines of code.

## Test:

With this change, all the markdown headers are now printed as expected, e.g.:

`## WatchDataManager`

`## StatusExtensionDataManager`


ps2 (approved):
LGTM!

Issue #2168, opened 2024-05-13
Avoid unnoticed carb entries


MaBeWi:
I am a the father of a young new looper using dexcom g7, omnipod dash and loop 3.2.3. Thanks to loop, our family finally sleeps well at night, so thank you for an awesome app! 

We had a scary situation a couple of days ago. After entering carbs and approving the insulin suggestion, the bolus injection failed because the pump was out of reach. Without us noticing, the carb entry was however saved in the app even though the insulin injection failed. 

We then entered the carbs again and because the app now assumed we were eating twice of our intended carb amount, the app suggested twice the insulin amount, which we then approved by mistake. My worry is that this can happen when at school etc. 

Potential solution: 

in cases where an insulin injection is approved by user; is it possible to save the carb entry  only if insulin injection proceeds without error?

marionbarker:
I don’t mean to minimize that this was frightening. But this is the design and it is documented. The Button you pressed says Save and Bolus. Perhaps we should modify the button to say Save Carbs and Bolus. Would that have made it more likely you would have taken a different action?

I certainly will move the warning box in this section to be higher up. And I will ask for more review from mentors (and you) on better words to get this concept across to new Loopers. 

https://loopkit.github.io/loopdocs/operation/features/carbs/#meal-entry_1

oliverstory:
Maybe 'Save carbs and attempt bolus' would be a good label for the button? Although it is a bit long.

Or 'Save carbs then bolus' - shorter but clear that the 'save carbs' happens first, so if you then get a 'bolus failed' message it is clearer that it's the bolus part rather than the 'save carbs' part that failed.

Our little looper is at daycare and I have received several calls from staff members who've encountered an issue at this step and wondered what to do next. I've seen this occur both with the 'pod not connected' error and when there's an automatic bolus in progress when you press 'save and bolus'.

motinis:
@MaBeWi - just wanted to let you know that I have something that may help with this - it will show you possible duplicate meals for the current carb entry on the meal bolus screen. So hopefully that will help people avoid this common pitfall (it's very common with caregivers who haven't working with Loop a lot). I can't say when it will be generally available but hopefully soon.

marionbarker:
A preliminary fix was added to Loop `dev` by modifying the `Save and Deliver` label to `Save Carbs & Deliver`. A longer term fix would put an in-line message in appropriate to the case where a bolus could not be delivered after that button is selected.

For that reason - I'm leaving this Issue open.

PR #2300, opened 2025-03-27
Update meal label, proposed remedy for Issue #2168

marionbarker:
## Purpose:

Provide a different label for the button when saving carbs and bolusing to make it more obvious that tapping the button saves the carbs.

## Associated Issue

See Issue #2168 for explanation.

Not understanding that carbs are saved even if bolus is not successful is a problem experienced by many new loopers and often by new caregivers for younger loopers.

### DIFF
--- a/Loop/Views/BolusEntryView.swift
+++ b/Loop/Views/BolusEntryView.swift
@@ -368,7 +368,7 @@ struct BolusEntryView: View {
                 case .saveWithoutBolusing:
                     return Text("Save without Bolusing", comment: "Button text to save carbs and/or manual glucose entry without a bolus")
                 case .saveAndDeliver:
-                    return Text("Save and Deliver", comment: "Button text to save carbs and/or manual glucose entry and deliver a bolus")
+                    return Text("Save Carbs, then Bolus", comment: "Button text to save carbs and/or manual glucose entry and deliver a bolus")
                 case .enterBolus:
                     return Text("Enter Bolus", comment: "Button text to begin entering a bolus")
                 case .deliver:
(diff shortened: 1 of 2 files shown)


marionbarker:
## Test

Apply this modification to LoopWorkspace dev, commit `3d48a5c`, Loop dev, commit 760174bd.
The screenshot below is from an iPhone 8 running iOS 16.7.10
The left side was built with unmodified dev, the right side was modified by this PR.

loop-pr-2230

marionbarker:
## Status

I suggest this button change be accepted for now with potential updates coming at a later time.
* Final wording on the button is "Save Carbs & Deliver"

## Discussion

I had a discussion with some Tidepool folks:
* They were concerned that the button change alone would not be sufficient to solve #2168
* I agreed to explore adding messages that indicated that "If carbs were added, they were saved"
   * After looking into this, my suggestion is to just make the button change for now

There are 2 places where added messages can happen
* Uncertain comms modal alert
* Bolus Issue alert

One negative of this approach is the carbs saved message would be displayed for any bolus or uncertain comms issues.
 
### Option: Modify uncertain comms modal message
 
If I implemented this, it could look like the graphic below, where the new text is underlined to make it easy to find.

updated_unacknowledged_save_carbs

### Option: Add to the Bolus Issue alerts

Default screens are seen in the [previous comment](https://github.com/LoopKit/Loop/pull/2300#issuecomment-2795382250) for some Bolus Issue alerts.

* I tried adding the sentence to the alert in the body of the message but the message truncates and important information is lost
* I then tried modifying the title of the alert, i.e., alert title changed to "Bolus Issue (Carbs Saved)" and was not convinced and even this is truncated with larger text sizes

ps2 (approved):
We'll have to back this out for a merge back to Tidepool, but I'm ok to land this for DIY for now.

PR #2303, opened 2025-03-29
Configure the version number in LoopWorkspace, not Loop

marionbarker:
## Purpose:

Simplify release process
* add a version number file to the LoopWorkspace folder and point to it from Version.xcconfig in the Loop folder
* this requires coordinated PR for Loop and LoopWorkspace

## Details

The Version.xcconfig file in the Loop repository is modified to point to the new file, LoopVersion.xcconfig, in the LoopWorkspace repository.

## Associated PR

The [LoopWorkspace PR 241](https://github.com/LoopKit/LoopWorkspace/pull/241) requires an update to the new Loop SHA after this PR is merged.

### DIFF
--- a/Version.xcconfig
+++ b/Version.xcconfig
@@ -6,10 +6,11 @@
 //  Copyright © 2020 LoopKit Authors. All rights reserved.
 //
 
-// Version [DEFAULT]
-LOOP_MARKETING_VERSION = 3.5.0
-CURRENT_PROJECT_VERSION = 57
+// Version
+//   [for DIY Loop] configure the version number in LoopWorkspace
+LOOP_MARKETING_VERSION = 1.0
+CURRENT_PROJECT_VERSION = 1
 
-// Optional override
-#include? "VersionOverride.xcconfig"
+// Optional override (enables override of version)
 #include? "../VersionOverride.xcconfig"
+#include? "VersionOverride.xcconfig"
--- a/Version.xcconfig
+++ b/Version.xcconfig
@@ -6,9 +6,6 @@
 //  Copyright © 2020 LoopKit Authors. All rights reserved.
 //
 
-// Version [DEFAULT]
-LOOP_MARKETING_VERSION = 3.5.0
-CURRENT_PROJECT_VERSION = 57
-
-// Optional override
-#include? "VersionOverride.xcconfig"
+// Loop is only built using LoopWorkspace, so read
+//   the version number from LoopWorkspace
+#include "../LoopVersion.xcconfig"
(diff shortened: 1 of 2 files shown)


Version.xcconfig
```diff
@@ -6,9 +6,6 @@
 //  Copyright © 2020 LoopKit Authors. All rights reserved.
 //
 
-// Version [DEFAULT]
-LOOP_MARKETING_VERSION = 3.5.0
-CURRENT_PROJECT_VERSION = 57
-
-// Optional override
-#include? "VersionOverride.xcconfig"
```
ps2, line 14: I believe if you add a VersionOverride.xcconfig to the workspace, you can override it there; that will remain compatible with the Tidepool build system.

ps2 (changes requested):

marionbarker:
I will test this.

Version.xcconfig
marionbarker: I added on line to enable looking for VersionOverride.xcconfig file in LoopWorkspace

Version.xcconfig
```diff
@@ -12,3 +12,4 @@ CURRENT_PROJECT_VERSION = 57
 
 // Optional override
 #include? "VersionOverride.xcconfig"
+#include? "../VersionOverride.xcconfig"
```
ps2, line 15: I think this might not be needed, as long as the VersionOverride.xcconfig file is added to the workspace. This change may break the way Tidepool Loop lays out its workspace.

marionbarker: For DIY Loop, we want to pick up VersionOverride in LoopWorkspace so we need a "../" to find it.

Do I need to reverse the order? Then if Tidepool has a VersionOverride in Loop, it would get picked up and override whatever is in the LoopWorkspace folder.

It would help if I knew how Tidepool lays out their workspace. Is there a link?

ps2: Tidepool's workspace is private as it references undisclosed partner information.  I just looked at how the resolution was happening for Tidepool Loop, and it's via a symlink that is put into the Loop directory at build configuration time. We can keep your change, and I'll update the Tidepool scripts to change where they put the override file. 

If we're moving this, then we can also remove the file from the .gitignore.  LoopOverride.xcconfig also behaves in a similar way, but we can leave this as is for now.

marionbarker:
To clarify - 
1. I leave the PR as is?

or

2. I remove the expectation of VersionOverride.xcconfig in the Loop folder?
   * remove the one line in Loop Version.xcconfig so it only points up one folder (in which case, I think removing the version lines in the Version.xcconfig makes sense)
   * remove VersionOverride.xcconfig from Loop .gitignore

ps2 (approved):
LGTM!

PR #2306, opened 2025-04-08
[QAE-456] Simple Calculator Text Field Identifiers 

Foscottl-TP:
Adds Accessibility Identifiers to the DismissibleKeyboardTextField elements within the Simple Bolus Calculator and Simple Meal Calculator screens. Needed for automating test cases associated with these elements.


Foscottl-TP:
Needs to be added to Tidepool Loop instead.

PR #2340, opened 2025-08-30
Updated translations from Lokalise on Sat Aug 30 10:22:12 PDT 2025

marionbarker:


marionbarker:
I have reviewed the changes.
LoopWorkspace builds with this modification and the other translations branches.

Note - there are a number of translations brought in because they were recently updated by translators. Some of them do not need to be included. 

Several weeks ago, I made the choice to delete some existing keys and translations on lokalise (for items like "CFBundleNames" and "NSHumanReadableCopyright"). However, for the next upload to lokalise, those items come back.  

The better choice - and what I'm doing now is to marked such items to not be shown to translators. This way they should not show up in future downloads from lokalise. This particular batch of imports has some new translations of older keys that will not show up in the future because they are marked not to be shown to translators.

If desired - I could go an hand-edit all the modified keys so they aren't changed, but it did not seem worth the manual effort.

LGTM.

ps2 (approved):
LGTM!

PR #2342, opened 2025-09-05
Convert to String Catalogs

marionbarker:
Use the automatic conversion in Xcode


marionbarker:
While working with String Catalogs
* add localization to some strings that were not previously localized for Favorite Foods
* add comments to words that were in the customization list without a comment
* disable localization for these strings: empty string, single dash, double dash and triple dash
* build after these changes to update the Localizable.xcstrings file(s)

These changes are brought in with the commit, c3c15e1d.

ps2 (approved):

PR #2343, opened 2025-09-11
Disable Liquid Glass

MikePlante1:
This just prevents Liquid Glass from being used when built to iOS 26 from Xcode 26, to keep Loop's UI consistent with prior iOS versions.

More info on this Boolean: https://developer.apple.com/documentation/BundleResources/Information-Property-List/UIDesignRequiresCompatibility

### DIFF
--- a/Loop/Info.plist
+++ b/Loop/Info.plist
@@ -84,6 +84,8 @@
 		<string>processing</string>
 		<string>remote-notification</string>
 	</array>
+    <key>UIDesignRequiresCompatibility</key>
+    <true/>
 	<key>UILaunchStoryboardName</key>
 	<string>LaunchScreen</string>
 	<key>UIMainStoryboardFile</key>


MikePlante1:
Biggest issue I see with Liquid Glass is that it's not graying out buttons that are disabled.

Screenshot 2025-09-11 at 12 42 21
Screenshot 2025-09-11 at 12 41 41
Screenshot 2025-09-11 at 12 40 24
Screenshot 2025-09-11 at 12 39 11
Screenshot 2025-09-11 at 12 36 24
Screenshot 2025-09-11 at 12 35 58
Screenshot 2025-09-11 at 12 35 42
Screenshot 2025-09-11 at 12 33 38
Screenshot 2025-09-11 at 12 32 34
Screenshot 2025-09-11 at 12 31 41

marionbarker:
@MikePlante1 Thank you for testing this on an iOS 26 beta phone.
I successfully tested this on iOS 15, 16 and 18 devices.
Behavior is nominal.

## Test

I tested on 3 different phone models:

* [✅] iPhone SE (3rd gen), iOS 18.2.6
* [✅] iPhone 8, iOS 16.16.7.11
* [✅] iPhone 7, iOS 15.8.4

Tests performed
* build to device with Xcode 16.4
* while testing observe that all button colors and actions appear nominal
* test that all plugins connect
   * CGM
   * Pump
   * Service (all work except Mixpanel is not available for iOS 15, no plan to modify this)
* finish by setting CGM to Nightscout, Pump to rPi DASH and Service to Nightscout
   * Get a closed loop (green icon)
   * Do a meal entry and manual bolus
   * Let it run several cycles
   * Deactivate pod
   * Logout of Nightscout Service

marionbarker (approved):
Code reviewed, tested and approved.
LGTM

## Details:

I reviewed the change and looked up what this boolean does.

It is provided by Apple to disable Liquid Glass, but is intended for temporary use while developers update and refine their app's UI. The option is expected to be removed in a future Xcode version, likely Xcode 27.

Given that several of the UI features in Loop do not work well with liquid glass, this gives us time to make proper updates.

itsmojo (approved):
Looks good for now. Wish Apple would have handled this differently though.

ps2:
Thanks; I think this works as a stopgap, but the longer term fix is to fix the usability issues that arose with recent iOS updates, and still retain as much of the system design idioms as we can.

marionbarker:
Agreed. I hope we take this opportunity to update the UI as you suggest.

PR #2347, opened 2025-09-16
Xcode Cloud Docs

Camji55:
This change allows building and distributing with Xcode Cloud which I have been doing on my personal branch for some time now. The intent is that existing workflows for browser builds with GitHub Actions and all Continuous Integration actions should still work as-is since Fastlane should override these code-signing settings. This will need some testing before it gets adopted but having the option to build and deploy within Xcode itself would be beneficial for users who'd like to take advantage of it especially if other options still remain available.

I'm also working on a write up on how to setup Xcode Cloud as an alternative to GitHub Actions for browser builds.

Please let me know what I can do to help verify all Fastlane and GitHub Actions remain working!

`dev` clone of #2476


marionbarker (changes requested):
In all other places in this file, where the syntax
```
DEVELOPMENT_TEAM = "$(LOOP_DEVELOPMENT_TEAM)";
```
is used, you see this syntax:
```
				CODE_SIGN_IDENTITY = "$(LOOP_CODE_SIGN_IDENTITY_DEBUG)";
				CODE_SIGN_STYLE = "$(LOOP_CODE_SIGN_STYLE)";
				DEVELOPMENT_TEAM = "$(LOOP_DEVELOPMENT_TEAM)";
--
				CODE_SIGN_IDENTITY = "$(LOOP_CODE_SIGN_IDENTITY_RELEASE)";
				CODE_SIGN_STYLE = "$(LOOP_CODE_SIGN_STYLE)";
				DEVELOPMENT_TEAM = "$(LOOP_DEVELOPMENT_TEAM)";

```

Where those environment variables are configured in `Loop.xcconfig`
```
// Code signing and provisioning [DEFAULT]
LOOP_CODE_SIGN_IDENTITY_DEBUG = Apple Development
LOOP_CODE_SIGN_IDENTITY_RELEASE = Apple Development
LOOP_CODE_SIGN_STYLE = Automatic
```

I expect @ps2 will request that you follow that model, so I am requesting changes. If he says it is not important, then I will withdraw my request for changes.

marionbarker:
## Test

The functional behavior is independent of whether "Apple Development" is entered directly for the CODE_SIGN_IDENTITY, so these tests are valid and successful.

✅ Mac-Xcode build: no issues building with this change.
✅ Browser Build: [Successful Build](https://github.com/docs-test/LoopWorkspace/actions/runs/17898979851)

marionbarker:
Having a write up for how to use Xcode Cloud will be excellent. Thank you for doing this.

Camji55:
@marionbarker thank you for testing! Yes, with this change there can be some variables removed from the xcconfig in the LoopWorkspace repo including:

LOOP_CODE_SIGN_IDENTITY_DEBUG
LOOP_CODE_SIGN_IDENTITY_RELEASE
LOOP_CODE_SIGN_STYLE

[https://github.com/tidepool-org/LoopWorkspace/pull/1521](Here’s what this looks like for Tidepool Loop). I’d imagine it’s the same here but let’s definitely let @ps2 chime in!

Camji55:
@marionbarker thank you for testing! Yes, with this change there can be some variables removed from the xcconfig in the LoopWorkspace repo including:

LOOP_CODE_SIGN_IDENTITY_DEBUG
LOOP_CODE_SIGN_IDENTITY_RELEASE
LOOP_CODE_SIGN_STYLE

Let’s definitely let @ps2 chime in!

Edit: No changes to LoopWorkspace needed

ps2 (approved):
Thanks! LGMT. Removing those other unused settings would be nice too.

marionbarker:
> @ps2 said: Removing those other unused settings would be nice too.

I want to be clear which settings should be removed.

Loop.xcconfig has these 3 lines:
```
LOOP_CODE_SIGN_IDENTITY_DEBUG = Apple Development
LOOP_CODE_SIGN_IDENTITY_RELEASE = Apple Development
LOOP_CODE_SIGN_STYLE = Automatic
```

I was suggesting that those variables be used in `Loop.xcodeproj/project.pbxproj` instead of hardcoding them as is done in this PR.

@Camji55 refers to those being in the LoopWorkspace: `LoopConfigOverride.xcconfig` file.
But they are not included in the DIY LoopWorkspace file at this time.

Camji55:
@marionbarker & @ps2 I added a doc on how to get Xcode Cloud running with help from Claude. Lmk if y'all think it's sufficient. I think it's important to note this change does not effect local builds/browser builds.

It might be worth another regression test @marionbarker, as it's been a bit since it was tested last.

Camji55:
I've got a mirroring PR targeting `next-dev` here: https://github.com/LoopKit/Loop/pull/2476

PR #2341, opened 2025-08-31
Bolus view fixes, and updates for iOS26

ps2:
Should fix https://github.com/LoopKit/Loop/issues/2335

PR #2349, opened 2025-09-19
Clear bolus recommendation on initial edit

ps2:
The onTapGesture callback is not called on a TextField. Better to use focus anyways; that works with keyboard focus as well.

### DIFF
--- a/Loop/Views/BolusEntryView.swift
+++ b/Loop/Views/BolusEntryView.swift
@@ -248,7 +248,11 @@ struct BolusEntryView: View {
                     .multilineTextAlignment(.trailing)
                     .foregroundColor(.loopAccent)
                     .focused($bolusFieldFocused)
-                    .onTapGesture { didBeginEditing() }
+                    .onChange(of: bolusFieldFocused) { focused in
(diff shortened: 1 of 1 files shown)


marionbarker:
## Test

Success. Works as expected on 4 iOS versions.

### Test Details

Use this version of Loop with v3.8.0 of LoopWorkspace.

Test with 4 different operating systems.
* iOS 26: works as designed
* iOS 18: works as designed
* iOS 16: works as designed
* iOS 15: works as designed

## Background information about the BolusEntryView screen

The designed behavior of tapping on the bolus line should automatically replace the recommended value with zero.

Previous tests showed that while this did happen for v3.6.4, it stopped working for some iOS systems after the update, PR #2341, commit a634e2f, was made to accommodate iOS 26 and to clean up old code that needed to be updated.

Loop v3.8.0 worked as designed for iOS 15 and iOS 16.
Loop v3.8.0 did not work as designed for iOS 18 and iOS 26.

marionbarker (approved):
Approved from code review and test.

PR #2353, opened 2025-09-20
Prefer opening Loop cgmManagerUI over CGM app when tapping CGM in HUD

elnjensen:
When the user taps the CGM info in the HUD, try to open the local interface for managing the CGM before trying to launch the separate CGM app.  Fixes a reported issue with Dexcom G6 no longer opening the Loop config page.

### DIFF
--- a/Loop/Extensions/DeviceDataManager+DeviceStatus.swift
+++ b/Loop/Extensions/DeviceDataManager+DeviceStatus.swift
@@ -82,12 +82,12 @@ extension DeviceDataManager {
     func didTapOnCGMStatus(_ view: BaseHUDView? = nil) -> HUDTapAction? {
         if let action = bluetoothProvider.bluetoothState.action {
             return action
+        } else if let cgmManagerUI = (cgmManager as? CGMManagerUI) {
+            return .presentViewController(cgmManagerUI.settingsViewController(bluetoothProvider: bluetoothProvider, displayGlucosePreference: displayGlucosePreference, colorPalette: .default, allowDebugFeatures: FeatureFlags.allowDebugFeatures))
         } else if let url = cgmManager?.appURL,
             UIApplication.shared.canOpenURL(url)
         {
             return .openAppURL(url)
-        } else if let cgmManagerUI = (cgmManager as? CGMManagerUI) {
-            return .presentViewController(cgmManagerUI.settingsViewController(bluetoothProvider: bluetoothProvider, displayGlucosePreference: displayGlucosePreference, colorPalette: .default, allowDebugFeatures: FeatureFlags.allowDebugFeatures))
         } else {
             return .setupNewCGM
         }


marionbarker:
Thanks for trying @elnjensen 

## Status

I don't think this is the right fix. I am testing a fix in CGMBLEKit. PR in a few minutes.

I believe the design is that tapping on glucose in the main screen should take you to the Loop screen for the G5, G6/ONE, G7/ONE+ or Libre and take you to the Nightscout URL for Nightscout as a CGM.

I do not have a Libre to test, so marked it as n/a in the table below.

I assume users of Nightscout as a CGM would be upset at losing that quick shortcut to see the Nightscout URL.

This shows the results of tapping on glucose in the main screen:

| CGM | dev Action | PR Action |
|:--|:-:|:-:|
| G5 |  Loop G5 screen | Loop G5 screen |
| G6 | Dexcom app | Loop G6 Screen |
| G7 | Loop G7 Screen | Loop G7 Screen |
| Libre | n/a | n/a |
| Nightscout | NS URL | Loop Nightscout Screen |

elnjensen:
Does Nightscout have a Loop status screen (with cgmManagerUI) though?  If not, then this will fall through to opening the URL. In general, if a cgmManagerUI is defined, there should be a way to access it.

elnjensen:
The other way to do this of course is just not to define `appURL` prior to when it's being invoked (the cgmManagerUI screen in Loop does it on the fly to open the app for both G6 and G7), but you said that Trio devs had requested the change that defines the URL separately. 

Maybe that's the simplest way (it's what G7 and Libre do) but then you just have to figure out what Trio needs and figure out an alternate way to handle that.  Not clear to me why that needs to be different for G6 vs. G7.  

(Edited to clarify)

marionbarker:
Please see: https://github.com/LoopKit/CGMBLEKit/pull/200

marionbarker:
> The other way to do this of course is just not to define `appURL` prior to when it's being invoked (the cgmManagerUI screen in Loop does it on the fly to open the app for both G6 and G7), but you said that Trio devs had requested the change that defines the URL separately.

I'm not quite sure what you mean about Trio. Trio does not use the returned appURL. I just tested it and all 3 dexcom choices open up what I was calling the Loop Gx screen but is in fact the CGMManager Gx screen.

elnjensen:
> I'm not quite sure what you mean about Trio. Trio does not use the returned appURL.

Sorry to be unclear.  When you made [this commit](https://github.com/LoopKit/CGMBLEKit/commit/905cc0a58509e417aa7bb2ef9dacfb36f009231c), which is what caused the change in G6 behavior, you [said in Zulip](https://loop.zulipchat.com/#narrow/channel/144182-development/topic/Inadvertent.20Change.20to.20G6.20Behavior/near/480740736) that you had done so “at the request of the Trio team.”  So I assumed there was some reason to keep that change - hence my reason for making the change I proposed.  I’m not wedded to that approach at all - reverting your previous commit so the G6 URL is `nil` is fine.  But if you can reconstruct the history of your comment, that could save problems or misunderstandings with Trio.

marionbarker:
Thanks for keeping me honest. Trio was completely rewritten between 0.2.x and 0.5.x (public beta first available in May 2025). So the reason for putting that in seems to no longer be an issue. I tested Trio 0.5.1.25 with and without a URL returned. It behaves the same way.

PR #2386, opened 2025-12-19

sachahartman:
Changed DefaultAbsorptionTimes


sachahartman:
I made a mistake, so sorry!

Issue #2396, opened 2026-01-16
Live Activity Glucose value unreadable due to color and truncation


Lytrix:
**Describe the bug**
The glucose value on the live activity widget is unreadable using the default settings.
This is due to the chosen color and default background of Apple. Especially for a color blind person the contrast is too low. 

Also the truncation dots are 70% of the time visible with mmol values instead of a decimal value on a iPhone 13 mini.

**To Reproduce**
1. Turn on live activity
2. Turn off/on phone to see the value in the daytime setting.

**Expected behavior**
As a color blind person I would like to have the default color be more contrast rich in the default setting so I can read the glucose value in the daytime and that the value always shows one decimal on my iPhone 13 mini.

**Screenshots**
<img width="281" height="609" alt="Image" src="https://github.com/user-attachments/assets/2958426c-4e42-4405-ad74-d237908152dc" />

**Phone**
 - Hardware: iPhone 13 mini
 - OS Version: iOS 18.7

**Loop Version**
 - Version Number: 3.9.5 (67) dev
 - Repo: LoopKit/Loop

**CGM**
 - Device: Dexcom G6
 - Manager app: Dexcom App

**Pump**
 - Manufacturer: Omnipod
 - Model: Dash
 - Firmware version: n.a.

**Additional context**
It's a great addition!

elnjensen:
I agree that we probably should consider changing the default settings here.  But to easily fix this in the short term, turn off the BG coloring option and you'll get black text.  I'd guess the coloring isn't giving you much additional info that you don't get at a glance anyway. 

You might also consider removing one of the displayed items in the bottom row (going from four to three) so they all display fully.

marionbarker:
FYI: I’m working on a PR to shorten names and expand descriptions for the bottom row. Also investigating options to automatically adjust Live Activity display when larger fonts are selected.

marionbarker:
Close because I think this was addressed.

Issue #2392, opened 2026-01-13
Live activity predictive line displays incorrectly if you are predicted to go below the glucose safety limit


StijnvanderLippe:
**Describe the bug**
There seems to be an issue with the predictive line in the live activity if you are expected to go below the glucose safety limit (before rising above the glucose safety limit). In this case, during the duration that you are expected to be below the glucose safety limit, there are two lines displayed. Note that this only happens if the "Use BG coloring" setting is enabled.

The predictive line in red is correct. The predictive line in green stays horizontal for the duration in which you are expected to be below the glucose safety limit. This part should not be there.

I've tried clearing the notification, closing Loop completely and reopening the app, but this makes no difference.

**Attach an Issue Report**

[Loop Report 2026-01-13 20_04_42+01_00.md.txt](https://github.com/user-attachments/files/24598606/Loop.Report.2026-01-13.20_04_42%2B01_00.md.txt)

**To Reproduce**
1. Enable the "Use BG coloring" setting in Live activity settings.
2. The predictive line in the Loop app must go below the glucose safety limit, before rising above the glucose safety limit.
2. Open your notification center
3. See that the predictive line is duplicated during the time in which you are expected to be below the glucose safety limit

**Expected behavior**
The green horizontal line should not be displayed and only the red line should be displayed during this time.

**Screenshots**

<img width="828" height="436" alt="Image" src="https://github.com/user-attachments/assets/eb5a6474-aae5-4acb-a4e2-163eb2e20d08" />

**Phone**
 - Hardware: iPhone 11
 - OS Version: iOS 26.1

**Loop Version**
 - Version Number: 3.10.0
 - Repo: 

**CGM**
 - Device: Dexcom G7
 - Manager app: Dexcom App

**Pump**
 - Manufacturer: Insulet
 - Model: Omnipod Dash
 - Firmware version: 4.10.0 1.40

marionbarker:
Thanks for reporting this.

I was able to reproduce it - see the last test case below.

My hardware is iPhone 15 pro running iOS 26.2.
Branch is feat/dev-dana-medtrum (commit `0b7b75f`)

I then tested with a SE 2nd gen phone running iOS 18.7.2 and reproduced it there as well. However, it only shows up when `Use BG coloring` is enabled.

## Test

My Live Activity plot did not have the flat green dashed line in my first attempts where the plot did not have an upward curve after dipping below the glucose safety limit:
[...]

Issue #2391, opened 2026-01-12
Live activity in 3.10.0 has occluded BG value with ellipses "..."


mattlevine22:
BG gets cut off (via "...") in lock-screen live activity widget.

**To Reproduce**
BG: 136, Active Insulin 4.0U, Active Carbs 41g, 1:39PM Updated.
**BG shows up as "13..."**

**Expected behavior**
Full visualization of the BG number ought to be prioritized spatially on the widget.

**Screenshots**
Screenshot attached

**Phone**
 - Hardware: iPhone 15
 - OS Version: iOS 26.1

**Loop Version**
 - Version Number: 3.10.10
 - Repo: LoopKit/Loop

**CGM**
 - Device: Dexcom G7
 - Manager app: Dexcom App

**Pump**
 - Manufacturer: Omnipod
 - Model: Dash
 - Firmware version: ?

**Additional context**
N/A. Reposted after closed in LoopKit issues.

<img width="590" height="1278" alt="Image" src="https://github.com/user-attachments/assets/88a722ea-09ad-445c-8f0d-9cc682b6a3d0" />

marionbarker:
I also use an iPhone 15, running 26.2. With standard text sizes, I do not have this issue. However, when I increase the text size I reproduce it. 

At that point I reduced the number of items in the bottom row to 3 instead of 4. This gives an immediate workaround.

I also noticed that there were times when the save button didn’t change between blue (active) and grey (inactive). 

Perhaps @elnjensen can take a look.

mattlevine22:
I also sometimes see the behavior even when all values fit in the window eg “168...". Will attach a screenshot when I can.

elnjensen:
> I also sometimes see the behavior even when all values fit in the window eg “168...". Will attach a screenshot when I can.

This is because the trend arrow is part of that display item, and it is being truncated.

As @marionbarker noted, you can easily choose to prioritize full display of the current BG (or any other fields) by choosing to have fewer fields present in the bottom row. I use large text and a small phone (iPhone 13 mini), and three fields works well.

elnjensen:
> I also noticed that there were times when the save button didn’t change between blue (active) and grey (inactive).
> 
> Perhaps [@elnjensen](https://github.com/elnjensen) can take a look.

[...]

PR #2397, opened 2026-01-16
Modify live activity

marionbarker:
## Purpose

Address Issues #2391, #2396 reported with the version of Live Activity release with Loop version 3.10.0.
Address issue #2392.

## Status (1 Feb 2026)

* improves behavior for large fonts
* adds more detail to description when choosing which widgets to include in live activity while using appropriate abbreviations to minimize truncation
* fixes the error reported in #2392

## Method

Information:
* The `name` is the string displayed in the widget
* The `description` is the string displayed when choosing what is shown in the bottom row of the widget

Action:
* Decrease the string length in some places for the `name`
* Expand the `description` to make it easier for the user to know what that selection does

Did a search for how to prevent truncation and settled on adding this
```
                    .minimumScaleFactor(0.5)
                    .lineLimit(2)
```

## Test

For both an iOS 18 and an iOS 26 test phone, set font size to use accessibility sizes and select the largest size.
* build with the version in this PR and observe strings were no longer truncated
* but then restored to `dev` branch and the strings were still no longer truncated
* I'm puzzled by this


marionbarker:
## Explanation

When testing some of the changes, the difficulty of handling large fonts made some of my initial changes impractical.

Instead, I added one more section to the Live Active screen display and added section headings to each. This enables expanded text strings that work for large and bold font.

### Suggested Changes
There are several suggestions in lokalise (from Felix) for preferred wording for the row labels.
1. Suggest "Show" or "Display" instead of "Add" for "Add predictive line"
2. Suggest "Configure bottom row" instead of "Bottom row configuration"
3. Suggest using "Size" instead of "Mode" for changing the size of the live-activity widget

### Discussion
1. I think "Display" is better.
   * The word "predictive" is only used in Live Activity, everywhere else in Loop "prediction" is used
   * The variable name associated with this option is `addPredictiveLine` - I don't propose changing the variable name
   * I modified the row label to "Display prediction in plot" 
2. Agree
3. I proposed changing "Mode" to "Display on Lock Screen" with these additions
   * I proposed "Plot and Bottom Row" instead of Large
   * I proposed "Bottom Row Only" instead of Small
   * Furthermore, item 1 above (Display prediction in plot) needs to be available regardless of what was originally called Mode, because it is used for what is shown when the user presses and holds the Dynamic Island for Loop
   * In practice - making these string longer was not practical; so instead I used section headings

### Additional Changes

* changed "Use BG coloring" to "Display colors for glucose" and added conditional to only show Upper and Lower limits when colors are enabled
* removed the conditional for displaying the Display prediction in plot toggle

## Test

When I made the proposed change from "Mode" to "Display on Lock Screen" the use of large font made it very difficult to modify the selection. 

Even restoring to "Mode" was too long to see both the label and the picker. 

Therefore, I restored the title to "Mode" and the shortened the picker selection choices to "Plot and Row" vs "Row" only.

Section headings were added to every existing section as well as for the new section.

## To Do

Things to fix -
1. Display and entry for glucose limits with Display colors for glucose
   * The display formatting for glucose levels when selecting Display colors for glucose is not limited to the nomal 0 decimal places for mg/dL or 1 decimal place for mmol/L
   * The values for the glucose limits should be entered with a picker wheel or at least checked for validity
      * The person types in the value instead of using a picker wheel
      * There is no restriction that the Lower Limit be less than the upper Limit
      * Consequence, now, if lower limit is > upper limit, all glucose is plotted and reported in red
2. The strings "Upper Limit" and "Lower Limit" are not localized

bastiaanv:
I like these changes. @ps2 has a strong opinion about the use of COB & IOB: https://github.com/LoopKit/Loop/pull/2191#issuecomment-2968807275.

A proposal for the #2392 is send via DM to Marion, Inspiration: https://stackoverflow.com/questions/75697214/swift-charts-draw-a-line-chart-where-line-color-is-determined-by-the-y-value

I do think Active Insulin & Active Carbohydrates should be consistent in the Configure Display view (if that would be the new label for it):

Simulator Screenshot - iPhone 17 Pro - 2026-01-19 at 20 47 10

polyfeux:
Hi Marion,
I am glad my suggestions are welcomed. 

In the current implementation changing the "large"/"small" selection to a more descriptive wording should be easy.
But today I thought about another solution and it depends on how flexibel the live-activity should be configurable in the future:
How about showing always the buttom row with the numbers and display the graph optionally with a switch "display graph"?

On the other hand: in the current implemention with the selection-wheel it should be quite easy to add additional configurations of the live-activity, e.g. displaying only the graph without any numbers in the bottom row.

Cheers, Felix

marionbarker:
@polyfeux I modified the UI substantially. I think adding screenshots of the current configuration will make discussions easier.

These graphics were prepared with commit f66d3320
* The first graphic is a composite of the Live Activity UI screens using a nominal text size.
* The second graphic shows just the Live Activity Mode selection screen
   * the top middle is with the largest font with bold font enabled, before the user taps on the selector row
   * the lower graphics show the selector options 
      * the left displays appearance with a large but not extreme font size
      * the right displays appearance with the largest font with bold font enabled
 
### Composite of Live Activity UI screens with nominal font size

pr-2397-commit-f66d3320-nominal

 
### Composite of Live Activity Mode selection with enlarged font size

pr-2397-commit-f66d3320-large


> Test Tip for others: for this test, rather than change font for the entire phone, I used the iPhone, Settings, Accessibility, Per App feature. This allows me to put the font size to max without making the entire phone hard to manipulate.

marionbarker:
## Comments

Refer to the screenshot composites in https://github.com/LoopKit/Loop/pull/2397#issuecomment-3770050442.

1. I think the descriptions in the configuration should have in parentheses the abbreviation used in the Live Activity display
   * For example, change description from "Active Insulin" to "Active Insulin (IOB)"
2. I think there may be a modification to the `ResizeablePicker` used for what is now called `Mode` that handles truncation and wrapping better - but I don't know how to do that. I handled the very large font by using a very short name and adding Section header strings

marionbarker:
## Update the glucose color plot

Using code revision from @bastiaanv, commit 101cde2, fixes the bug reported in Issue #2392

## Test

Configure a test phone which has flat glucose pattern
* Add 4 U non-pump insulin
* Add 45 g carbs 1 hour in the future
* Resultant lock screen plot (with glucose color enabled) has the bug reported in 2392

Build with the proposed change from @bastiaanv.

The graphic shows the lock screen plot prior to the update on the left with the modified code on the right.

loop-pr-2397-fix-glucose-color-plot

Confirm when color is disabled, the glucose plot returns to Loop blue with bottom row using black font when phone is light mode or white font when phone is in dark mode.

marionbarker:
## Update descriptions

With this commit, 6575f96, the following changes were implemented.

* The descriptions were updated to include the short label (name) shown on the display in parentheses. 
* The comments used for localization hints were included to stop them being modified frequently and to remove the `"isCommentAutoGenerated" : true` lines in the `Localizable.xcstrings` files

## Test

This was tested with nominal, larger and extremely large text sizes.  The interface continues to be usable.

## Graphics

The composite below uses this more recent commit.

loop-pr-2397-description

Lytrix:
I am testing the PR using:
- iPhone 13 mini
- iOS 18.7 (many friends told me they had a bad experience in battery/speed by updating to iOS 26 on a 13 model)
- App SKU: 123

It really helps turning the color off, really helps for readability. Only the graph is still very hard to read on the Light setting.
DayNight


I do see 2 issues:
**1. The prediction slider does not seem to update properly. I get different outcomes:**

1.  It only updates after force closing the app, but then it only removes it in the Main view Left top, not in the Live Activity.
2. If you hit save with the slider to OFF and force quit the app and start it, live activity will sometimes update properly but sometimes still show the prediction, but initially when opening Loop, the prediction arrow will be removed in the Main view off Loop. You see it is added back in a split second when you click on Live Activity to open Loop again.
3. If I save it to ON and force quit, it will not show the Predication in the Main Left top of the app, but does show it in the Live Activity view, it does revert to also show in the Main view if I click on the live Activity to open Loop.

IMG_0404

**2. You cannot Save the initial settings of the display options after saving another setting, the save button is greyed out.**
The challenge with that it gives issues when you save a new setting and then want to save the initial setting again. You expect it to be blue to save, but instead it will set it to grey remembering nothing has changed since the initial state.
Save state

marionbarker:
Update the localizations for this PR. Confirm it builds correctly.

IMO - this PR provide sufficient improvements that it should be merged as is and added to the proposed 3.11.1 branch for further testing.

* improves behavior for large fonts
* adds more detail to description when choosing which widgets to include in live activity while using appropriate abbreviations to minimize truncation
* fixes the error reported in #2392

ps2 (approved):
Looks ok to me. Not great to move in the direction of being more "jargony", but if you haven't found other ways to make it work, then ok.

StijnvanderLippe:
I believe the code introduced in https://github.com/LoopKit/Loop/commit/101cde26a8e4dc94f373fc94ac97b6da1e6e8cd8 does not work fully as intended. This can lead to issues such as the lowest portion of the predictive line always being coloured red, even though the user is within the ranges defined in the Live Activity "Display control options".

See examples here: https://github.com/LoopKit/Loop/issues/2392#issuecomment-3983883839

I tried looking at the changes in the code and observed the following. I am a complete amateur at Swift however so I may be interpreting the code wrong. So everything below is just my assumption of how the code works by inspecting it. Please correct me if this is wrong.


In `getGradient`, `lowerStop` and `upperStop` are normalised based on the highest value in the predicted line.
From `Loop Widget Extension/Live Activity/ChartView.swift`:
```swift
private static func getGradient(useLimits: Bool, lowerLimit: Double, upperLimit: Double, highestValue: Double) -> LinearGradient {
    var stops: [Gradient.Stop] = [Gradient.Stop(color: Color("glucose"), location: 0)]
    if useLimits {
        let lowerStop = lowerLimit / highestValue // Normalisation of values
        let upperStop = upperLimit / highestValue // Normalisation of values
        stops = [
            Gradient.Stop(color: .red, location: 0),
            Gradient.Stop(color: .red, location: lowerStop - 0.01),
            Gradient.Stop(color: .green, location: lowerStop),
            Gradient.Stop(color: .green, location: upperStop),
            Gradient.Stop(color: .orange, location: upperStop + 0.01),
            Gradient.Stop(color: .orange, location: 600), // Just use the mg/dl limit for the most upper value
        ]
   }
   return LinearGradient(
        gradient: Gradient(stops: stops),
        startPoint: .bottom,
        endPoint: .top
   )
}
```
So the Gradient.Stop points are assigned based on normalised glucose levels.

However, when applying the gradient in the following piece of code, the values passed are _not_ normalised. This results in the gradient being applied in the wrong way.

From `Loop Widget Extension/Live Activity/ChartView.swift`:
```swift
ForEach(predicatedData) { item in
    LineMark (x: .value("Date", item.x),
                     y: .value("Glucose level", item.y) // NOT normalised
    )
    .lineStyle(StrokeStyle(lineWidth: 2, dash: [6, 5]))
    .foregroundStyle(colorGradient)
}
```

By mixing normalised and not normalised values, the line is coloured in the wrong way.

Again, this is just my hypothesis based on reading the code so I may be understanding it wrong.

elnjensen:
@StijnvanderLippe thanks for this comment.  The fix is more complicated than just having one dataset normalized and the other not, but looking at normalizations put me on the right track.  I just submitted a PR (#2410) to fix this.  Try it out when you have a chance.

PR #2399, opened 2026-02-04
Include submodule SHA in LoopWorkspace build details

marionbarker:
## Purpose

Report the names and SHA for all the submodules included in the LoopWorkspace build for Loop installed on a particular device.

This is especially important for feature branches under test.

## Method

Borrow the additions found in this commit from Trio:
*  https://github.com/nightscout/Trio/commit/b79d0904b

Then add the submodule list to the Loop Report.

## Test

The code in this PR was tested by building and examining the Loop Report as shown in the graphic below with the added lines highlighted by the blue rectangle.

> Note - this is not the final format. See https://github.com/LoopKit/Loop/pull/2399#issuecomment-3850300191

add-submodule-sha

## Nice to have

These are additional changes to consider:
* The current report lines for gitRevision and gitBranch could be removed.
* It would be nice if the submodules were reported in alphabetical order
* If might also be nice to include this feature from Trio in which the user can tap on the version information on the Settings screen to see the submodule information on their phone without sharing a log file.
   *  https://github.com/nightscout/Trio/commit/69b6d10f2


ps2 (approved):
LGTM

marionbarker:
## Update

1. I got the alphabetized submodule "nice-to-have" feature added, thanks to @bjorkert for the assist
2. I modified the format so my python parser, which reports Build Details as part of the report, continued to work
3. I revised the workspace output
4. I removed the redundant lines as a separate commit (3b868170) to make it easy to restore them if they are needed for some reason

## New Screen Shot

This is an example of the Build Detail from the final version of the updated code:

Screenshot 2026-02-04 at 3 39 03 PM

PR #2401, opened 2026-02-05
Add submodule commit timestamps to build details

jwoglom:
Extends the submodule metadata from PR #2399 to also capture and display each submodule's commit timestamp. The timestamp is stored as a unix epoch string ("commit_timestamp") in the plist, and shown as a YYYY-MM-DD UTC date in parentheses after the SHA in the build details report.

### DIFF
--- a/Common/Models/BuildDetails.swift
+++ b/Common/Models/BuildDetails.swift
@@ -67,16 +67,17 @@ class BuildDetails {
     }
 
     /// Returns a dictionary of submodule details.
-    /// The keys are the submodule names, and the values are tuples (branch, commitSHA).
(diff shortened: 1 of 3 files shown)


marionbarker:
Thank you for your contribution @jwoglom.

I understand you are traveling and do not have access to a computer.

I've prepared the graphics that show the change introduced by this PR.

This example uses the feat/eversense branch of LoopWorkspace, so includes the DanaKit, EversenseKit and MedtrumKit repositories.

Notice the Loop submodule information points to the use of this branch.

loop-pr2401-add-date-to-submodule-build-details

marionbarker (approved):
Approved by code review and test.

PR #2410, opened 2026-03-06
Fix Live activity plot: colors in prediction, override target, time-range and labels

elnjensen:
This PR addresses two issues in the Live Activity plot: 

1. Correctly color-code the predicted line, addressing issue #2392; 
2. Make override and target range bars visible when min and max are the same.

### DIFF
--- a/Loop Widget Extension/Live Activity/ChartView.swift
+++ b/Loop Widget Extension/Live Activity/ChartView.swift
@@ -141,9 +141,9 @@ struct ChartView: View {
                 }
             }
             .chartForegroundStyleScale([
-                "Good": colorInRange,
-                "High": colorAboveRange,
-                "Low": colorBelowRange,
+                "Good": Self.colorInRange,
+                "High": Self.colorAboveRange,
+                "Low": Self.colorBelowRange,
                 "Default": Color("glucose")
             ])
             .chartPlotStyle { plotContent in
--- a/Loop Widget Extension/Live Activity/ChartView.swift
+++ b/Loop Widget Extension/Live Activity/ChartView.swift
@@ -18,7 +18,11 @@ struct ChartView: View {
     private let preset: Preset?
     private let yAxisMarks: [Double]
     private let colorGradient: LinearGradient
-    
+
+    private static let colorInRange = Color.green
+    private static let colorBelowRange = Color.red
+    private static let colorAboveRange = Color.orange
+
     init(glucoseSamples: [GlucoseSampleAttributes], predicatedGlucose: [Double], predicatedStartDate: Date?, predicatedInterval: TimeInterval?, useLimits: Bool, lowerLimit: Double, upperLimit: Double, glucoseRanges: [GlucoseRangeValue], preset: Preset?, yAxisMarks: [Double]) {
         self.glucoseSampleData = ChartValues.convert(data: glucoseSamples, useLimits: useLimits, lowerLimit: lowerLimit, upperLimit: upperLimit)
         self.predicatedData = ChartValues.convert(
@@ -29,7 +33,7 @@ struct ChartView: View {
             lowerLimit: lowerLimit,
             upperLimit: upperLimit
         )
-        self.colorGradient = ChartView.getGradient(useLimits: useLimits, lowerLimit: lowerLimit, upperLimit: upperLimit, highestValue: yAxisMarks.max() ?? 1)
+        self.colorGradient = ChartView.getGradient(useLimits: useLimits, lowerLimit: lowerLimit, upperLimit: upperLimit, lowestValue: predicatedGlucose.min() ?? 1, highestValue: predicatedGlucose.max() ?? 1)
         self.preset = preset
         self.glucoseRanges = glucoseRanges
         self.yAxisMarks = yAxisMarks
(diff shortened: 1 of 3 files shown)


StijnvanderLippe:
The line colouring work as desired. Override bars do not display correctly.

Here are a couple of tests. I had an override active from 11:38 to 12:38, with 80% insulin needs as the only override.

| Lower limit (mmol/L) | Upper limit (mmol/L) | Screenshot | Comment |
| - | - | - | - |
| 7 | 13 | image | Line is red below 7 mmol/L as desired. There is no orange colouring since I'm never above the upper limit. The override bar is displayed inside the "Eventual BG" text instead of inside the graph. |
| 6 | 13 | image | Same as above. The red colouring now starts below 6 mmol/L as desired. The override bar again displays incorrectly. |
| 4 | 7 | image | Now the line is never red, since I'm never below the lower limit, as desired. The line is coloured orange above the upper limit as expected. The BG level is also coloured according to the user's limits. The override bar again displays incorrectly. | 
|4 | 6 | image | Similar test as previous case, line colouring starts where expected and the BG level is coloured correctly. The override bar again displays incorrectly. |
| 5 | 5.5 | image| Line is red below 5 mmol/L, and orange above 5.5 mmol/L, as desired. The BG level is coloured correctly. The override bar again displays incorrectly, this time between the current glucose level and the "Eventual BG" text since some time has passed compared to previous tests. | 

In all tests the line colouring is how it's supposed to be, so that's the good news. The override bar does not seem to display correctly if the only override is the insulin needs. Below some additional tests to display this behaviour:

| Override target range (mmol/L) | Override insulin needs | Override duration | Screenshot | Comment |
| - | - | - | - | - | 
| 8.5 - 9.0 | Unchanged | 3.5 hours | image| Displays correctly |
| 4.4 - 4.5 | 140% | Indefinite | image | Displays correctly | 
| Unchanged | 130% | Indefinite | image| Displays incorrectly. The override bar seems to be placed at where 0 mmol/L would be, but this is just a guess. |

From these tests it seems that if correction range is changed, the override bar displays correctly, regardless of insulin needs override. If _only_ the insulin needs are changed, then the override bar displays incorrectly.

elnjensen:
> If only the insulin needs are changed, then the override bar displays incorrectly.

Is this behavior new with this PR?  Are you able to test it with the released code? I wonder if it's a separate bug. 

Edit to add that this happens for live activity mg/dl graph as well. 

I didn't even realize it was possible for the range field to be blank in an override! (Maybe @bastiaanv didn't either, though he's probably ahead of me. 😉) I'm traveling and can't edit code right now, but I think this is a separate issue, and that the LA code needs to check to see if the range fields in an override are blank, and then fall back to displaying the override bar with the main range bounds if so.

StijnvanderLippe:
> Is this behavior new with this PR?  Are you able to test it with the released code? I wonder if it's a separate bug. 

I went back to 3.12, the issue is also present but in a slightly different way. For the duration of the override the override bars are not visible at all (at least, I cannot find them in the live activity view).

Example below for an override of 1.5 hours with only insulin needs set at 70%:

image

So it seems the issue was already present, but the changes in this PR make it so the issue shows up in a slightly different way.

StijnvanderLippe:
Hi @elnjensen, I was just wondering what should be done about the the override display issue? Where the live activity displays the target range incorrectly if the override has a target range without any entries.
Should I make a separate issue for it? Since it also displays incorrectly in version 3.12 and thus was not introduced by this PR.

Because other than that, I've been running the code in this PR for weeks now and the colouring works exactly as expected.

elnjensen:
Honestly, I had forgotten about this! I'll try to have a look when I have a chance (hopefully soon) and update the pull request. But in terms of when it might get merged, I don't have any control over that.

elnjensen:
@StijnvanderLippe I've added a fix for the case you flagged; overrides without a glucose range set should now display correctly in the Live Activity graph.  Try it out when you have a chance.

marionbarker:
## Test

This fixes the color glitch and the empty target for overrides.

marionbarker (approved):
Approve by code review and test.

marionbarker:
## Test

✅ successful test, but . . .

In the past - the left most tick of the x-axis truncated the displayed value.  I know you fixed it in the past by adjusting the time-range slightly.

I suspect claude did not realize that was important and dropped that fix. 

If you have time to fix this, that would be great. Otherwise, I will accept this as is because it's a big improvement.

### Test Case

Turn on an override that lasts for one hour.
Wait until that override is in the past.
Notice the left side of the x-axis is too far in the past as time goes by.

Build new version with latest commit, 273f875a

Plot x-axis range is now correct.

### Screenshot

* ❌ Left side of graphic  is screenshot of Live Activity plot using Loop dev branch, commit 367a9878
* ✅ Right side of graphic is screenshot of Live Activity plot after building with live_activity_plot branch, commit 273f875a
   * ❓ The initial hour marking is messed up - this is something that was fixed before and needs to be fixed again 

loop-pr2410

marionbarker:
## Test

✅ successful fix of truncated labels

### Screenshots

* Left side with prior commit 273f875a
* Right side with latest commit 25c41514

loop-pr2410-label-fix

LGTM

PR #2466, opened 2026-07-20
Enable Liquid Glass + Fix Separated Toolbar + Liquid Glass Fixes

Camji55:


ps2 (approved):
LGTM!

ps2 (commented):
LGTM!

marionbarker (approved):
The code review looked nominal but I am not an expert.

✅ I did a successful test.

### Test Details

I tested using an SE 3rd gen phone running iOS 26.5.2.

The current test phone is connected to a DASH pod, so I did not delete and add back the pump. However, I did test every other button for the toolbar and the HUD. I added carbs and accepted the bolus, I did a manual bolus, I added a premeal target so I could enable and disable that feature, I added a preset so I could select and dismiss the preset. I did delete and add a new CGM so I saw the selection of CGM with the new liquid glass layout both from the HUD and the settings screen, screen shots below. I assume the add pump will look the same on that phone.

I also built to an iPhone 17 simulator and added a pump there.

Screenshots:
* add CGM from HUD

Image

* add CGM from Settings

Image

PR #2467, opened 2026-07-20
Presets Training CTA Fix

Camji55:
Fixes https://loop.zulipchat.com/#narrow/channel/144182-development/topic/Loop.20next-dev/with/611640192

### DIFF
--- a/Loop/Views/Presets/Training/PresetsTrainingView.swift
+++ b/Loop/Views/Presets/Training/PresetsTrainingView.swift
@@ -152,16 +152,14 @@ public struct PresetsTrainingView: View {
                         .padding(.bottom, 24)
                         .padding(.horizontal, 16)
                 }
+
+                if let cta = step.cta {
+                    ctaButtons(for: cta)
+                        .padding(.top, 8)
+                        .padding(.horizontal, 16)
+                }
             }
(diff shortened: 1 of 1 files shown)


marionbarker:
## Test

✅ successful test (but the prior version worked for this phone)
❓ The behavior has changed - the continue button is also visible, you do not need to scroll down

### Configuration

start with next-dev, commit 14b70bd
point Loop submodule to this PR and build on test phone with no loop installed.
SE 3rd gen running iOS 26.5.2

* press and hold the green loop to skip onboarding therapy settings
* do not bother to add pump or cgm
* tap on presets and see the 4 initial presets
* tap on + Start Required Training
* all screens were accessible and I did not have to scroll to reach the continue button
   * the not needing to scroll is different behavior
   * not sure if requiring the user to scroll to the bottom of each screen is desired. 

Note - with this phone I was able to complete the training without going to landscape mode and all of the continue rows were accessible and worked without this change. However, I understand this did not work with other phone models.

Camji55:
@marionbarker okay, latest changes should restore the behavior while still applying a fix. Are you able to reproduce the bug in an iOS 26.5 iPhone 15 Pro simulator?

PR #2458, opened 2026-07-16
Xcode 27 Build Support

Camji55:
- SceneDelegate migration
- Bump watch deployment target to 8
- `@State` macro migration fix
- `capture-build-details.sh` script update to not be blocked by stale git links

PR #2492, opened 2026-08-12
Add iOS 27 scene lifecycle support

htenlik:
## Problem

When Loop is built with the iOS 27 SDK, the app builds successfully but fails at launch because UIKit now requires adoption of the scene-based lifecycle:

`Application failed to launch: UIScene life cycle is required for apps built with this SDK.`

Xcode 27 also exposes an initialization-order issue in `FavoriteFoodDetailView`.

## Changes

- Add `SceneDelegate` and register it through `UIApplicationSceneManifest`.
- Move `LoopAppManager` launch and active-state handling from `AppDelegate` to the scene lifecycle.
- Preserve the existing storyboard and root navigation architecture.
- Fix `FavoriteFoodDetailView` stored-property initialization order for Xcode 27.

## Testing

Tested with:

- macOS 26.5.2
- Xcode 27 beta 5
- iOS 27
- iPhone 16 Plus

The patched build installs and launches successfully. I exercised normal application navigation and usage without observing launch crashes or regressions.


ps2:
Was fixed with https://github.com/LoopKit/Loop/pull/2458, which just landed. Thanks!

htenlik:
My pleasure!

PR #2498, opened 2026-08-20
Xcode 27 Fixes

Camji55:


marionbarker (approved):
Approve by test.

Builds to iOS 26 and iOS 18 phones. I assume this will work for iOS 27.

PR #2504, opened 2026-09-07
Increase spacing and size of main screen Liquid Glass toolbar icons

elnjensen:
On iOS 26 the five toolbar items share one Liquid Glass background with no spacing between them. Any space item inserted between them splits that shared background into separate pills, so the gaps here are transparent padding baked into each rendered icon instead, which widens the spacing while leaving the single button group intact.

The button sizing is scaled on each icon's visible artwork rather than its canvas. It turns out that the button graphics pad themselves by very different amounts -- Bolus has essentially none, Carbs has ~3.7pt top and bottom, and Settings is a 25x25 gear centered on a 25x40 canvas -- so matching canvas sizes leaves the glyphs visibly uneven and gives Settings a noticeably taller pill. The ink bounds were measured by rasterizing the assets at 600dpi and taking the alpha bounding box; artwork is centered in every one to within 0.6pt, so scaling alone aligns them.

This PR also detects and records the Pre-meal item's index at setup instead of hardcoding it per OS version, since the two layouts place it differently.


ps2:
Pushed two follow-ups to this branch (43b4f5f) — happy to drop either if you disagree.

**Runtime ink measurement instead of the constants table.** Replaced `ToolbarLayout.Ink` with an alpha-channel scan, so `toolbarIcon()` drops its `ink:` argument and the five constants and their call-site plumbing go away. Your measurements were good — I re-rasterized at 600dpi and they sit a consistent ~0.4pt above the α≥128 bounding box on every icon, which lands rendered ink within 2% across the set. The only reason to change it is the staleness caveat your own comment calls out. It also gives the ink's origin, so each icon centers on its artwork rather than its canvas; that removes the dependency on the "centered to within 0.6pt" property rather than relying on it. Cost is scanning five ~90×120px images once at button construction.

**`iconVerticalMargin` 4 → 7.** Widths were already past 44pt from the horizontal padding, but height was `inkHeight + 8` = 38pt, under the 44pt minimum — and tap target is part of what is being reported alongside the visual size. At 7 every item is 44pt tall (carbs/pre-meal 55.5×44, workout 54.8×44, settings 50×44, bolus 46.2×44).

One thing I specifically checked before pushing, since Loop allows all four orientations and a compact-height toolbar is ~32pt: a 44pt item could have been a squeeze there. Verified in both portrait and landscape on an iPhone 16 Pro running iOS 26 and it looks right. Worth knowing the settings asset already renders 40pt tall on `dev` today, so this is 4pt onto an existing situation rather than something new.

Nice find on the Liquid Glass spacers splitting the shared background — I had gone down the path of padding the canvas to 44pt and got a wider pill with unchanged icons, exactly the wrong half of the problem. Scaling on ink is the right call.

ps2:
Pushed a third commit (6f03460) that takes a different approach to the sizing, after finding a problem that affects both the constants table and the runtime scan I had replaced it with. Details, since it supersedes the earlier commit:

**`Pre-Meal Selected` and `workout-selected` are inverted badges, not line art** — a filled plate covering the whole canvas with the glyph knocked out. Their alpha bounds are therefore the plate (29×29), not the glyph inside it (25.2×21.1 and 25.6×23.2). Both `preMealImage(selected:)` variants go through the same `toolbarIcon` call, so:

- With the `Ink` constants, the selected pre-meal image is scaled by the *unselected* glyph's 21.57, so the plate renders 40.3pt tall on a 38pt item and gets clipped ~1.2pt top and bottom, flattening its rounded corners. Workout lands at 36.4pt and fits, but the item still jumps ~26% larger on toggle.
- With the runtime scan, the plate itself becomes 30pt, so the knocked-out glyph shrinks 27% (pre-meal) / 20% (workout) when the button is toggled on.

Today neither shows because nothing scales and both states share a canvas.

**Fix: normalize the assets instead of compensating in code.** All seven imagesets are now on a common 40×40pt canvas with the artwork 30pt tall. The PDFs stay vector — each page is re-placed with a scale and translate through CoreGraphics, not rasterized. The two badges use their unselected partner's transform, so toggling swaps in place with the plate at its natural size (38pt / 35pt). The settings gear is the one PNG and is upscaled 1.2× from its @3x source (75→90px); its `height-class: compact` variant is dropped so landscape loads the same asset as portrait — that variant was 20×31 with a 20pt gear, so the single `settings` constant would have rendered it 20% small in landscape.

With the assets normalized, `toolbarIcon()` only centers the image on the item canvas (`ToolbarLayout.itemSize`, 60×44 — the 44pt tap target plus your Liquid Glass spacing). `Ink`, `inkBounds()`, `inkHeight` and the padding constants are all gone: net −90 lines. Your `preMealItemIndex` fix is untouched.

Verified on an iPhone 16 Pro / iOS 26 in portrait. One follow-up worth doing separately: the gear is now the only raster asset in the set and would benefit from a vector redraw.

ps2:
16c8b8d73494ee63114c5c6bb2d4b70b04da2321

elnjensen:
This looks good to me - I like the approach of normalizing the assets in the first place, much cleaner.

marionbarker:
## Test

✅ successful test

### Configuration

added this version of Loop to 2 different phones:
* SE 2nd gen running iOS 18
* SE 3rd gen running iOS 26

### Narrative

The iOS 18 phone does not display liquid glass.
* ✅ The tool bar icons on the main screen are slightly larger with this PR
* ✅ All buttons worked nominally

The iOS 26 phone does display liquid glass
* ✅ The tool bar once again takes the full width of the main screen with this PR
* ✅ Tapping anywhere near each icon worked as expected

### Screenshots

✅ The screenshots are nominal

#### iOS 18

The composite below shows before (left) and after (right) this PR for an iOS 18 phone

loop_pr2504_18


#### iOS 26

The composite below shows before (left) and after (right) this PR for an iOS 26 phone

loop_pr2504_26

marionbarker (approved):
Approve by test following code review by others.

Issue #815, opened 2018-09-21
Prediction calculation is missing bounds checks


dennisgove:
I received the following prediction showing a predicted value less than 0.

![prediction](https://user-images.githubusercontent.com/654519/45853712-c9f3e500-bd14-11e8-87d5-a9bb1b1593ce.jpg)

dennisgove:
I wasn't able to find the code calculating the prediction. If this belongs in [LoopKit](https://github.com/LoopKit/Loopkit) I'm happy to move it.

PR #2511, opened 2026-09-14
Show bolus progress for automatic boluses, and offer an issue report when one is stopped

ps2:
Two commits: show the bolus progress row for automatic boluses, then use that row to offer an issue report when one is stopped.

## Show the progress row for automatic boluses

`bolusState`'s `didSet` created the progress reporter only when `dose.automatic != true`, so an automatic bolus displayed nothing on the status row. Those guards came in with #815, when limiting the dosing strategy to automatic bolus.

The row is also the stop target — `didSelectRowAt` on `.bolusing` is what cancels a bolus — so with nothing displayed there was no way to stop an automatic bolus from the status screen at all. That is the dose a user is most likely to want to interrupt.

Both guards are removed. A bolus replacing an in-progress one now also gets a fresh reporter, which covers a manual bolus superseding an automatic one.

No size or duration threshold: every automatic bolus shows the row, matching manual behaviour. Worth knowing that the status row is inserted and removed rather than swapped in place (`shouldShowStatus` follows `statusRowMode.hasRow`, and `updateBannerAndHUDandStatusRows` does a real `insertRows`/`deleteRows`), so under automatic-bolus dosing the table will shift on most loop cycles, including for 0.05 U boluses that deliver in a couple of seconds.

## Offer an issue report when one is stopped

Stopping an automatic bolus is the moment a user has decided the dose was wrong, and the last moment the data explaining it is cheap to collect. The canceled row already reports what was delivered; it now carries an info button that explains why Loop bolused and offers to open the issue report screen.

- The button appears **only when the canceled dose was automatic**. Someone who stops their own manual bolus has nothing to report.
- The canceled row now stays up **30 seconds instead of 10** for automatic cancels. Ten is not long enough to notice an affordance, read it and act on it; adding a button to a row that disappears first would achieve nothing. Manual cancels are unchanged at 10.
- Issue report presentation is factored out of `didTapIssueReport`, which reached it through a `dismiss` that only makes sense when called from the settings sheet.

## What this does not cover

An automatic bolus cut short by something other than the row tap still just disappears — no "Bolus Canceled" state. `canceledDose` is set only on the tap-to-cancel path, and `DeviceDataManager.enactBolus` cancels a running automatic bolus directly when a manual one arrives.

Left alone deliberately: `determineStatusRowMode` ranks `canceledDose` above `.bolusing`, so surfacing that interruption would mask the replacement bolus's own progress for the display duration. Reporting it properly wants somewhere other than this row.

## Implementation note

The info button is appended to the cell's stack view programmatically rather than added to `BolusProgressTableViewCell.xib`, via `progressLabel.superview as? UIStackView`. The `progressLabel` outlet points at the stack's first arranged subview, so the cast holds, but it is worth knowing it would fail silently rather than loudly if the xib were rearranged.

## Testing

Built and installed on an iPhone 16 Pro. Confirmed the progress row appears for automatic boluses and can be tapped to stop one, that the canceled row then shows the info button, and that the alert opens the issue report screen.


elnjensen:
Thanks for bringing this back - I think adding the report button is a nice addition.

Camji55 (approved):
lgtm!

ps2:
Follow-up: this PR exposed a latent bug, fixed in #2519.

Giving automatic boluses a progress reporter and a place in the status row is what makes a cancel able to be followed by another bolus through the same row. `bolusState.didSet` skipped creating a reporter when the previous state was `.canceling`, so the next bolus inherited the cancelled bolus's reporter and the row reported one bolus's delivered units against another's programmed units — seen in the field as "Bolus Canceled: Delivered 1.80 of 0.20 U".

That hole, and the cell reseed that `main` does on bolus start, were both dropped in #872; they were simply unreachable until this change.

PR #2519, opened 2026-09-16
Don't carry a bolus progress reporter across a cancel

ps2:
Reported in the field as a status row reading **"Bolus Canceled: Delivered 1.80 of 0.20 U"** — delivered larger than programmed, and neither figure matching the recorded dose.

## What happened

The two numbers come from two different boluses. From the reporter's Loop Report:

```
23:50:45 → 23:51:31   prog 1.9   deliv 1.15   manual      ← stopped from the status row
23:51:44 → 23:51:52   prog 0.2   deliv 0.2    automatic   ← 13 s later
```

`0.20` is the automatic bolus's `programmedUnits`, taken from the status row's dose. `1.80` came from the progress reporter still tracking the 1.9 U manual bolus that had just been cancelled — 1.80 was that reporter's time-based estimate at the moment of cancel, which is why it matches neither the 1.9 programmed nor the 1.15 the pump actually delivered.

## Cause

`bolusState.didSet` skipped creating a reporter when the previous state was `.canceling`:

```swift
case .inProgress(let doseNew):
    switch oldValue {
    case .inProgress(let doseOld):
        guard doseNew.syncIdentifier != doseOld.syncIdentifier else { break }
        bolusProgressReporter = ...
    case .canceling:
        break                  // ← stale reporter survives into the next bolus
    default:
        bolusProgressReporter = ...
    }
```

`main` does not have this hole — any transition into `.inProgress` from a state other than `.inProgress` builds a fresh reporter, and it also reseeds the visible cell:

```swift
// If there is an existing bolus progressCell, update its dose values now in case the app is currently in the
// background as otherwise these values won't get initialized and can contain stale data from some earlier bolus.
progressCell.totalUnits = dose.programmedUnits
progressCell.deliveredUnits = 0
```

Both behaviours were dropped here in #872 when that switch was restructured. They stayed dormant until #2511, which gave automatic boluses a progress reporter and a place in the status row — that is what allows a cancel to be followed by another bolus through the same row.

## Fix

Recreate the reporter on any transition into `.inProgress` that is not the same dose, and seed a visible cell with the new dose's programmed units and zero delivered.

The seed matters independently of the reporter: a bolus that starts while the app is backgrounded produces no progress callback, so without it the cell shows the previous bolus's numbers until one arrives.

## Testing

Builds clean. Not reproduced on hardware — it needs a bolus cancelled from the status row followed closely by another bolus, which is the sequence in the report above.

## Third commit: don't loop on CGM data that wasn't stored

Found from Marion's logs on this PR. Both times an automatic bolus could not be cancelled, it was one that started within seconds of the user cancelling the previous one — and both were triggered like this:

```
Skipping adding glucose sample due to existing cached sync identifier: 6aaa…
Triggering loop from new CGM data at 04:10:11
bolusState: initiating
```

The Nightscout CGM plugin re-delivers the newest reading Loop already holds on every poll. The glucose store rejects it as a duplicate, but `cgmManager(_:hasNew:)` ran the loop on any `.newData` more than 4.2 minutes after the last CGM-triggered loop, and discarded the one thing that said nothing was stored — `addGlucoseSamples`' return value. So a poll with no new glucose ran a full dosing cycle against the just-reduced IOB and recommended the shortfall back.

`processCGMReadingResult` now returns whether a sample was stored, and the loop trigger requires it. The plugin-side half (stop re-delivering the cutoff reading) is LoopKit/NightscoutRemoteCGM#36.

Not addressed: on the next *legitimate* reading, the loop will still re-recommend the shortfall of a bolus the user just cancelled. Nothing on the dosing side remembers a cancel; that's a policy question, not a bug fix.

## Fourth commit: also gate the heartbeat/refresh loop on new CGM

The third commit gated the CGM *push* path (`cgmManager(_:hasNew:)`). Marion had that commit and still saw two autoboluses ~19 s apart, because the *pull* path — `refreshCGM()`, run on the pump BLE heartbeat — looped on `lastLoopCompleted > 4.2 min` and ignored whether a new reading had arrived. It dosed 1.5 U on the previous reading when the heartbeat fired, then the fresh reading dosed 1.35 U 19 s later.

`refreshCGM()` now loops only when the CGM check stored a new value (the `storedNewGlucose` it previously discarded). With both the push and pull paths gated on a stored-new sample, the glucose store's syncID dedup is the single arbiter — whichever stores a given reading first loops, the other no-ops — so heartbeat and push can no longer double-fire on the same reading. A heartbeat with no new CGM now just wakes and fetches, per the intended design (loop only on fresh CGM).

Note: this also applies to a manual pull-to-refresh (via `refreshDeviceData()`), which now likewise loops only on a new reading.


marionbarker:
## Test

✅ Successful fix
See section about single failure to cancel AB - it was immediately after a rebuild

### Reproduce the error

Build next-dev, commit f0c0f55
This has the Loop modification that adds autobolus back to the bolus progress on the main screen but is missing the code in this PR.
Use an SE 3rd gen phone running iOS 27.0 that is using Nightscout as a CGM and is attached to an Omnipod 5 pod.

The glucose pattern on the test phone is flat.
* enter 45 g carbs and reduce the actual bolus to 1 U
* as soon as bolus progress shows up, lock the phone (0.1 U of 1 U)
* at next CGM entry, should hear pod delivering
* ❌ unlock phone and note mismatch in expected total, it incorrectly indicates 1 U instead of the actual amount
* cancel the bolus as soon as it exceeds 1 U (1.1 of 1.0)

### Demonstrate the fix

Open the Loop while rebuilding.
Point the Loop submodule to fix/stale-bolus-progress and rebuild.

❌ As soon as the build was finished, I closed the loop and the app immediately start an autobolus of 0,65 U. I was not able to cancel the autobolus before it completed.

* add more carbs and issue a 0.5 U manual bolus, ensure I can cancel this bolus and I can
* issue an 0.4 U manual bolus and lock the phone as soon as the bolus progress display starts
* wait for next CGM update when I hear the pod start to click and unlock to confirm the display updated properly
   * ✅ it successfully shows progress for xx of 1.3 U
   * and I was successful at canceling the autobolus

### Test ability to cancel autobolus

Disable closed loop again but do not do a rebuild.
* wait for next cgm reading to come in
* enable closed loop and see if an autobolus starts
* ✅ autobolus started and was able to cancel it

### Single failure to cancel AB

Very disconcerting that I could not cancel an autobolus when I first enabled closed loop following a rebuild.

I think it was because I had just rebuilt and sometimes the phone is sluggish to respond to taps after a rebuild - takes a few minutes to be responsive.

I continued testing by letting autobolus run for a few cycles and was ✅ always successful canceling later AB.

The Xcode debug log from building with this PR is attached.

[20260915_2054_Xcode_CouldNotCancelFirstAB_after_rebuild,txt.zip](https://github.com/user-attachments/files/32270504/20260915_2054_Xcode_CouldNotCancelFirstAB_after_rebuild.txt.zip)

marionbarker (approved):
Approve by test

marionbarker:
This may be a new issue. I unlocked the phone when I hear clicks, cancelled an autobolus and a new one started and I could not cancel that one.

I will do more testing tomorrow.

marionbarker:
This is the last few minutes of the Xcode debug log.
* [20260915_2115_Xcode_cannot_cancel_AB.txt](https://github.com/user-attachments/files/32270829/20260915_2115_Xcode_cannot_cancel_AB.txt)

ps2:
Thanks Marion — both "could not cancel" cases are the same bug, and it's mine from LoopKit/Loop#2511, not the rebuild. Your logs show it exactly:

```
03:45:05  AB 0.65 U → cancelled           04:09:52  AB 0.55 U → cancelled
03:45:12  AB 0.60 U starts (+7 s)  ✗       04:10:12  AB 0.40 U starts (+20 s)  ✗
```

After a cancel, the "Bolus Canceled: Delivered …" row stays up for 30 seconds for an automatic bolus, and `determineStatusRowMode` ranks that row above an in-progress bolus. So a bolus that starts inside that window has no "Tap to Stop" row at all until the timer expires — and a 0.4 U bolus is long finished by then. Before #2511 automatic boluses never used this row, so the ordering was harmless.

Fixed in the second commit here: a new bolus now clears the canceled row and takes it over, so it can be stopped. Your first two steps (lock during progress, unlock at next reading) still exercise the original stale-values fix; the re-test for this one is simply "cancel an autobolus, then cancel the one that starts right after".

Separately, in both logs Loop starts a **new automatic bolus within seconds of the user cancelling one**, and at 03:50 it even tried to enact another while one was running (`enactBolus failed: unfinalizedBolus`). Cancelling an automatic bolus is effectively futile if the next cycle re-issues it immediately.

ps2:
Added the fix for the immediate re-dose, since it's the same failure you saw: the loop that issued the second autobolus was triggered by a Nightscout poll re-delivering a reading Loop already had (`Skipping adding glucose sample… → Triggering loop from new CGM data` in both your logs). Loop now only triggers from CGM data that was actually stored. There's a companion fix on the Nightscout plugin side so it stops re-delivering that reading in the first place.

marionbarker:
## Test

✅ Successful test

### Configuration

Update the SE 3rd gen iOS 27 phone connected to Nightscout as CGM and Omnipod 5 pod using next-dev for all submodules (except Loop) and use fix/stale-bolus-progress for Loop.

List of submodules at end of this comment.

### Demonstrate the solution

* add carbs and change bolus to 0.5 U, lock phone as soon as bolus progress shows
* at next cgm reading, hear pod begin to deliver and unlock
* ✅ confirm that the bolus progress is correct

### Repeat the cancel for multiple autobolus

* add more carbs with no manual bolus
* cancel each autobolus and ensure another one does not start right away
   * ✅ CGM reading 1 - cancel was successful, left phone unlocked
   * ✅ CGM reading 2 - cancel was successful, this time I locked the phone after the cancel
   * ✅ CGM reading 3 - cancel was successful
      * ❓ I tapped the (i) for more information and while reading about the issue report option, another autobolus started again, but I was able to ✅ cancel it successfully
   * ✅ CGM reading 4 - cancel was successful
      * repeat the cancel and tap the (i), read it, tried the two options (issue report and not now)
      * this time there was not another autobolus, lock phone
   * ✅ CGM reading 5 - cancel was successful
 
### List of submodules:
```
* buildDateString: Wed Sep 16 11:13:21 PDT 2026
* xcodeVersion: 17F113
* Workspace branch: next-dev
* Workspace SHA: f0c0f55
* Submodule name: branch, SHA
*   AmplitudeService: next-dev, d73d20d
*   CGMBLEKit: next-dev, 48bbfad
*   EversenseKit: detached, e7b1ad7
*   G7SensorKit: next-dev, fb87aa9
*   LibreCRKit: detached, c562170
*   LibreLoop: detached, 415b799
*   LibreTransmitter: next-dev, afba845
*   LogglyService: next-dev, 44f97c9
*   Loop: fix/stale-bolus-progress, 56fb6ad
*   LoopAlgorithm: detached, 7d016fb
*   LoopKit: next-dev, 6083d81
*   LoopOnboarding: next-dev, da95d69
*   LoopSupport: next-dev, b93873f
*   MedtrumKit: detached, 3b3af0c
*   MinimedKit: next-dev, 50b942b
*   Minizip: detached, 66483b4
*   MixpanelService: detached, 4a4b433
*   NightscoutRemoteCGM: next-dev, 526328e
*   NightscoutService: next-dev, 09dc797
*   OmnipodKit: next-dev, d9b5966
*   RileyLinkKit: next-dev, 8ba261c
*   TidepoolService: next-dev, 765bcb9
*   TrueTime.swift: master, 5dc1285
*   dexcom-share-client-swift: next-dev, e4c2796
```

marionbarker:
## Configuration Update:

I checked the submodule listing in the previous comment and noticed this:
* NightscoutRemoteCGM: next-dev, 526328e

This is behind actual next-dev. I did not update properly. I will repeat this test.

For this one submodule, these are the commits I was missing:
```
NightscoutRemoteCGM % gl -10
1a73e9e (HEAD -> next-dev, origin/next-dev) Merge pull request #36 from LoopKit/fix/exclude-cutoff-sample
6b0aeb3 Don't report the newest known reading as new on every poll
fc78fec Merge commit '6ca838aa6872a55ffabf30c6594a32f018b74817' into integrate-next-dev
443ff32 Localize optional API Secret helper text and placeholder
6ca838a Update xcstrings
526328e Sort glucose before filterDateRange; fix latestBackfill selection
```

marionbarker:
## Test again

### Updated configuration

I was using an old clone - so I pulled a new one so all my remotes point to the correct place.
I used a modified define_common so the update_submodules_ref got the correct branches.

Starting with LoopWorkspace, next-dev `f0c0f55`, these are the updates to submodules:
LoopWorkspace % git submodule status Loop NightscoutRemoteCGM 
+bf02eb3a7d45c48e6a9bdb76e1ec2da733b3d4f7 Loop (v3.4.2-636-gbf02eb3a)
+1a73e9e7c491111273c46b3b7d9ee57339c1b9e5 NightscoutRemoteCGM (heads/next-dev)

### Planned test

Build to the same phone and configure so each CGM reading will trigger an autobolus.
Cancel each autobolus and observe the behavior, making sure no back to back autoboluses happen during the test.

marionbarker:
## Test Duplicate Autobolus

✅ Successful test

The last test was successful in demonstrating Autobolus could always be canceled.
This test is to see if we get back to back autobolus as happened in one time out of five in this test: https://github.com/LoopKit/Loop/pull/2519#issuecomment-5702912807

### Configuration

The configuration is detailed here: https://github.com/LoopKit/Loop/pull/2519#issuecomment-5706279496

### Test Details

Upload a flat glucose pattern of about 110 mg/dL to Nightscout at 5 minute intervals.
* Enter carbs and save without bolusing
* Lock the phone
* For each new CGM reading, you can hear the pod start to click, unlock phone and cancel the autobolus and then lock the phone again

### Test Result

For the table below, the ✅ indicates I was able to cancel the autobolus and another did not happen.

| time | cancel autobolus? | 
|:--|:-:|
| 17:16 | Enter 100 g carbs |
| 17:19 | ✅ |
| 17:24 | ✅ |
| 17:29 | ✅ |
| 17:34 | ✅ |
| 17:39 | ✅ |

### Screenshot of cancelled autoboluses

IMG_7911

marionbarker (approved):
approve by test

PR #2521, opened 2026-09-16
Don't re-evaluate glucose alerts from a backfill older than the last reading

ps2:
CGMs deliver the live reading and then backfill the gap behind it, each as its own `.newData` batch, and `DeviceDataManager` hands every batch to `GlucoseAlertManager.evaluate(samples:)`. The backfill carries a sample no newer than the one just evaluated, and `evaluate` re-decided the alert from it.

Seen with a G7 on `next-dev`, from the AlertStore and DeviceLog:

```
22:58:59  didRead        glucose 86  timestamp 89595   → evaluate → urgent low issued
22:58:59  didReadBackfill glucose 86  timestamp 89591   → evaluate → urgent low retracted 22:59:00
```

Retraction goes through `unscheduleAlertWithSchedulers`, which also calls `criticalAlertAudioPlayer.stop()` — so on an un-entitled build the in-process alarm audio was cut off one second after it started, and the notification was pulled with it. The user saw a banner slide in and out and heard nothing.

## Fix

Skip any batch whose newest sample is not newer than the last evaluated reading. Backfill still reaches the glucose store; it just cannot override an alert decision that was made on a newer reading.

## Testing

Builds clean. The sequence above is reproducible with the "next reading as urgent low" test on a G7, which is how it was found; with the guard the backfill batch is logged and ignored.


ps2:
Folding this into #2520 — same fix set, one PR.

PR #2515, opened 2026-09-15
Use real bar button items for the status toolbar

ps2:
Fixes #2513.

## What it does

Makes all five status-screen toolbar items plain `UIBarButtonItem`s with an image, a title and an action, instead of four `UIBarButtonItem(customView:)` wrapping padded `UIButton`s plus one image item. UIKit now provides each item's tap target, spacing, and — when the bar does not fit — a working entry in the overflow menu. The 60×44pt icon padding (`toolbarIcon()`, `ToolbarLayout`), the custom buttons, and the index bookkeeping for swapping the pre-meal item in and out are removed; the pre-meal and workout items are updated in place. 35 insertions, 125 deletions, one file.

## Why

On iOS 27 the five items no longer fit narrower phones in portrait and UIKit collapses the tail into a `···` overflow menu. A custom-view item has no title, image or action for that menu to show, so the menu was empty and Settings — last in the array — was unreachable without rotating. The bar was also wider than it needed to be: #2504 (`c1c170a6`, in both `dev` and `main`/v3.14.7) padded every icon to a 60×44pt canvas, 300pt of icon canvas before toolbar insets.

## Verified

iPhone 12 mini simulator (360pt):

- iOS 27.0, v3.14.7 code: the pill overruns both screen edges, gear clipped.
- iOS 27.0, this change: all five items fit.
- iOS 26.5, this change: all five items fit, as before.

Not exercised: an actual `···` tap (no tap automation), and landscape, which is unchanged in structure.

The branch is based on `2b4cd9a2`, the commit both `dev` and `main` pin, so it cherry-picks clean to `main` for a 3.14.8.

Issue #2513, opened 2026-09-15
UI is buggy on smaller screens, e.g. iPhone 12 mini


sergiu-fintineru:
<img width="563" height="1218" alt="Image" src="https://github.com/user-attachments/assets/e49bc802-30a6-4732-9f3f-2786099f772b" />
<img width="1218" height="563" alt="Image" src="https://github.com/user-attachments/assets/a1ad9f25-f7f8-4594-9c04-89f8f21be01a" />

**Describe the bug**
I cannot access the settings section in portrait mode. The 3 dot button is not functional.

[Loop Report 2026-09-15 20:30:02+03:00.md](https://github.com/user-attachments/files/32252926/Loop.Report.2026-09-15.20.30.02%2B03.00.md)

sergiu-fintineru:
I have to rotate the screen in order to acces the settings cogwheel.
I am running iOS 27.
Dexcom ONE+ and Omnipod Dash pumps.

sascha-egerer:
I’ve got the Same Problem. If I remember correctly, the Interface was different before I’ve updated to iOS27

sergiu-fintineru:
I did both updates in the same time, meaning I updated my iPhone to iOS 27 and Loop to 3.14.7, I wasn't sure where the problem stemmed from. It's probably an issue with iOS 27 and the new interface rules.

bastienmoulia:
On bigger screen (iPhone 17 Pro)  it's better but still not centered and a bit small:
![image](https://github.com/user-attachments/assets/0044d821-b5bf-4e08-82b9-ada06adeaf34)

sascha-egerer:
I mean the Main Problem is Not that the dots are there, but the Problem is that the Button does just nothing :) Rotating the phone is my current workaround.

And this is a iPhone 16 Pro by the way

<img width="1206" height="547" alt="Image" src="https://github.com/user-attachments/assets/809c078d-af61-4249-9bf2-ea8f0a8bed0e" />

<img width="2579" height="485" alt="Image" src="https://github.com/user-attachments/assets/ca4b2634-3499-41c5-9d43-7a9e013a31cf" />

sergiu-fintineru:
The developers did warn us not to update yet to iOS 27, but we decided to take one for the team.
Hopefully it will be resolved soon, the workaround is indeed to use the phone in landscape mode.

ps2:
Thanks for the report — the screenshots and the Loop Report pinned this down.

[...]

PR #2524, opened 2026-09-17
Fit five toolbar items at 375pt when built with Xcode 27

ps2:
Follow-up to #2515 for LoopKit/Loop#2513.

## What it does

An app linked against the iOS 27 SDK gets wider padding around each bar button item on iOS 27. At 375pt (iPhone SE 3rd gen, iPhone 12 mini) only three of the five toolbar items fit and the rest go to the overflow menu. The same commit built with Xcode 26.5 fits all five, which is why the browser builds (pinned to Xcode 26.5) were unaffected and local Xcode 27 builds were not.

Sets `isPaddingRemoved` on the five items, which brings them back within 375pt, and gives Settings `visibilityPriority = .high` so it is the last item to overflow rather than the first. Both properties are new in SDK 27, so the block compiles only with Swift 6.4 (Xcode 27); SDK 26.5 builds do not need it and are unchanged. One file, 8 lines.

## Verified

Xcode 27 build, iOS 27.0: iPhone SE (3rd generation) and iPhone 12 mini both show all five items (three + `···` before). iPhone 12 mini, iOS 26.5: unchanged.


marionbarker:
## Test

✅ successful test

### Configuration with the problem

Code configuration: branch update_dev_to_3.14.8, commit `3502c50`, (includes #2515)

Build onto an SE 3rd gen phone running iOS 27.0 using both Xcode 26.6 and Xcode 27 and see that 5 icons do not fit. The "..." icon must be tapped to reach the Settings and Workout Targets toolbar items.

IMG_7912

### Test the Fix

The desired configuration is for all 5 toolbar icons to show in portrait mode

point the Loop Submodule to fix/toolbar-sdk27-padding

Build with both Xcode 26.6 and Xcode 27.

The Xcode 26 screenshot (left) and Xcode 27 screenshot (right) have slightly different format, but both show the 5 toolbar items. I tested tapping near the icons and all buttons were properly selected without deadspace between icons.

loop_pr2524

### Simulator result

I also built to an iPhone 18 Pro Max simulator and the 5 icons are presented as shown in the screenshot below.

Screenshot 2026-09-17 at 4 12 29 PM

marionbarker (approved):
Approve by code review and test.

PR #2525, opened 2026-09-19
Show the development branch warning at most once a week

ps2:
Tapping "I'm a tester" records the date; the warning is skipped for seven days after that. "Switch to main" does not record it, so it shows again on the next launch.

Stored in `UserDefaults.standard`, like the other once-only alert flags in `AlertManager`. Same change is going to `next-dev`; the file is identical on both branches.


marionbarker:
## Test

✅ Successful at suppressing the dev branch I am a tester modal question.

One issue is sometimes Loop quits and restarts, and because the modal blocks the app from continuing, this does not necessarily first that issue. Although it does fix the annoyance of having to tap it quite so often.

### Configuration

Use dev as base branch
* point Loop to fix/dev-branch-warning-weekly
* build onto an SE 2nd gen iOS 18 phone

### Test

First build of this version at 15:32, I had to tap, I am a tester
Second build of this version at 15:33, the app opened without a modal reminder that I was building the dev branch.

marionbarker (approved):
Approve by test.

Is the date updated every time the app restarts?  In that case, it would be likely the app would be restarted at least once in a week.

If not, is there a way to show the reminder without blocking Loop from continuing in case it's been a week since the modal was acknowledged and Loop quits/restarts in the middle of the night or while the user is not paying attention.

